PR #16, opened 2021-07-22
Update README.md

gegoune:
Insipired by/copied from https://github.com/NTBBloodbath/color-converter.nvim/edit/main/README.md.


ibhagwan:
This is great tysm!

I started writing the FAQ which will be placed under the Wiki section, I might be able to post it later if I make enough progress, you’re welcome to beautify and contribute to that as well once it’s published :)

PR #25, opened 2021-07-26
open the correct man page section

zeertzjq:
If there are multiple man pages of the same name in different sections, `getmanpage(selected)` removes the section number, and the wrong man page is opened. In man pages, references to other man pages have no space between name and section number either.

### DIFF
--- a/lua/fzf-lua/providers/manpages.lua
+++ b/lua/fzf-lua/providers/manpages.lua
@@ -13,8 +13,17 @@ local actions = require "fzf-lua.actions"
 local M = {}
 
 local function getmanpage(line)
-  -- match until comma or space
-  return string.match(line, "[^, ]+")
+  -- extract section from the last pair of parentheses
+  local name, section = line:match("^(.*)%((.-)%)[^()]-$")
+  if name:sub(-1) == " " then
+    -- man-db
+    name = name:sub(1, -2)
+  else
+    -- mandoc
+    name = name:match("^[^, ]+")
+    section = section:match("^[^, ]+")
+  end
+  return name .. "(" .. section .. ")"
 end
 
 M.manpages = function(opts)


ibhagwan:
Tysm! I’ll do some more testing tomorrow and merge this if I find no issues.

ibhagwan:
I'm getting a nil concat error on some entries, for example:
```lua
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
> print(string.match("zzz, ZZZ(8)", "^(.*) (%([^()]*%)) *$"))
nil
```

zeertzjq:
Does `man -k` print man pages in this format? It seems to always add a space before the section number.

ibhagwan:
You can see the command at [config.lua#L208](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/config.lua#L208), `man -k .` and that’s what gets printed on my machine.

zeertzjq:
That weird. I looked into the source code of man-db and found this:
```c
	if (!long_output && strlen (page_name) > (size_t) (line_len / 2))
		string = xasprintf ("%.*s...", line_len / 2 - 3, page_name);
	else
		string = xstrdup (page_name);
	string = appendstr (string, " (", newinfo->ext, ")", (void *) 0);
	if (!STREQ (newinfo->pointer, "-") && !STREQ (newinfo->pointer, page))
		string = appendstr (string, " [", newinfo->pointer, "]",
				    (void *) 0);

	if (strlen (string) < (size_t) 20) {
		int i;
		string = xrealloc (string, 21);
		for (i = strlen (string); i < 20; ++i)
			string[i] = ' ';
		string[i] = '\0';
	}
	string = appendstr (string, " - ", (void *) 0);
```

ibhagwan:
I can't argue with what my eyes are seeing, few examples below:

screenshot-1627352769

screenshot-1627352850

zeertzjq:
I'm trying to find a way to produce that kind of output.

BTW, I just found another problem: `man -k .` trims page names if they are too long (e.g. `__gnu_pbds::detail::left_child_next_sibling_heap_node_point_const_iterator_` can be trimmed if the terminal is not wide enough). Using `apropos -l .` instead of `man -k .` can solve this problem.

ibhagwan:
> I'm trying to find a way to produce that kind of output.
> 
> BTW, I just found another problem: `man -k .` trims page names if they are too long (e.g. `__gnu_pbds::detail::left_child_next_sibling_heap_node_point_const_iterator_` can be trimmed if the terminal is not wide enough). Using `apropos -l .` instead of `man -k .` can solve this problem.

That's some hard work you're doing for a man page :-)

I knew my string parsing there wasn't optimal but I didn't know to which extent, even Telescope has issues with some man pages, I just decided to deal with it "another day" as I didn't believe it was that critical.

Definitely appreciate the help!

zeertzjq:
Can you run `/usr/bin/man -k .` instead?

ibhagwan:
> Does `/usr/bin/man -k .` also produce this output?

yep

zeertzjq:
I just found out that there are two `man` implementations: `man-db` and `mandoc`.

zeertzjq:
I've moved the processing to `getmanpage()`. `getmanpage()` determines whether `man-db` or `mandoc` is used by checking whether there is a space before the parentheses surrounding the section number.

zeertzjq:
I also found out that Neovim `:Man` command uses `-S`, not `-s`, so it only supports specifying section when using `man-db`, not when using `mandoc`. I guess changing it to `-s` can make it support both `man-db` and `mandoc`.

ibhagwan:
Thank you for this most intensive man research!

I’ll test it tomorrow and hopefully merge.

ibhagwan:
Works great, just merged this, tysm for your hard work on this!

PR #27, opened 2021-07-30
Add open in background file buffer

kdheepak:
Hi, 

First, thanks for making this plugin! It is such a nice lua version of fzf.vim :) 

I'm interested in opening selections in "the background", i.e. just adding them to the buffer list but not switching to them. What are your thoughts about adding it to such an action? I didn't add them any default keybindings since if someone wants to use this they can create keybinding in their user config.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -69,6 +69,11 @@ M.file_tabedit = function(selected)
   M.vimcmd_file(vimcmd, selected)
 end
 
+M.file_open_in_background = function(selected)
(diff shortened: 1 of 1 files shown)


ibhagwan:
I think it’s great, I surely wouldn’t mind adding this, can we make the function name a bit more self explanatory though? Perhaps `file_open_in_ background` or something similar?

kdheepak:
Yes, of course! I've changed it to `file_open_in_background`.

ibhagwan:
Merged, ty!

kdheepak:
Thank you!

PR #36, opened 2021-08-09
add lsp.workspace_diag_only_cwd option to filter out diagnostics entries from files outside of the current working directory

greenhat:
### Motivation
`rust-analyzer` LSP server loads a new LSP server instance and pushes diagnostics for dependencies (external crates) when you open their files (via "jump to definition", etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ require'fzf-lua'.setup {
   lsp = {
     prompt            = '❯ ',
     -- cwd               = vim.loop.cwd(),
+    cwd_only          = false, -- show workspace diagnostics only for the files in cwd
     file_icons        = true,
     git_icons         = false,
     lsp_icons         = true,
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -302,6 +302,7 @@ M.globals.lsp = {
           ["Information"] = { icon = "", color = "blue" },      -- info
           ["Hint"]        = { icon = "", color = "magenta" },   -- hint
       },
+      workspace_diag_only_cwd = false,
   }
 M.globals.builtin = {
       prompt              = 'Builtin> ',
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -302,7 +302,7 @@ M.globals.lsp = {
           ["Information"] = { icon = "", color = "blue" },      -- info
           ["Hint"]        = { icon = "", color = "magenta" },   -- hint
       },
-      workspace_diag_only_cwd = false,
+      cwd_only = false,
   }
 M.globals.builtin = {
       prompt              = 'Builtin> ',
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -315,6 +315,10 @@ M.diagnostics = function(opts)
   local lsp_type_diagnostic = vim.lsp.protocol.DiagnosticSeverity
   local current_buf = vim.api.nvim_get_current_buf()
 
+  if opts.cwd_only then
+    opts.cwd = vim.loop.cwd()
+  end
+
   -- save this so handler can get the lsp icon
   opts.cfg = config.globals.lsp
 
@@ -384,14 +388,12 @@ M.diagnostics = function(opts)
         end
       end
       for bufnr, diags in pairs(buffer_diags) do
-        if not opts.workspace_diag_only_cwd or (opts.workspace_diag_only_cwd and string.find(vim.api.nvim_buf_get_name(bufnr), vim.loop.cwd(), 1, true) ) then
-          for _, diag in ipairs(diags) do
-            -- workspace diagnostics may include empty tables for unused bufnr
-            if not vim.tbl_isempty(diag) then
-              if filter_diag_severity(opts, diag.severity) then
-                diagnostics_handler(opts, cb, co,
-                  preprocess_diag(diag, bufnr))
-              end
+        for _, diag in ipairs(diags) do
+          -- workspace diagnostics may include empty tables for unused bufnr
+          if not vim.tbl_isempty(diag) then
+            if filter_diag_severity(opts, diag.severity) then
+              diagnostics_handler(opts, cb, co,
+                preprocess_diag(diag, bufnr))
             end
           end
         end
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -384,12 +384,14 @@ M.diagnostics = function(opts)
         end
       end
       for bufnr, diags in pairs(buffer_diags) do
-        for _, diag in ipairs(diags) do
-          -- workspace diagnostics may include empty tables for unused bufnr
-          if not vim.tbl_isempty(diag) then
-            if filter_diag_severity(opts, diag.severity) then
-              diagnostics_handler(opts, cb, co,
-                preprocess_diag(diag, bufnr))
+        if not opts.workspace_diag_only_cwd or (opts.workspace_diag_only_cwd and string.find(vim.api.nvim_buf_get_name(bufnr), vim.loop.cwd(), 1, true) ) then
+          for _, diag in ipairs(diags) do
+            -- workspace diagnostics may include empty tables for unused bufnr
+            if not vim.tbl_isempty(diag) then
+              if filter_diag_severity(opts, diag.severity) then
+                diagnostics_handler(opts, cb, co,
+                  preprocess_diag(diag, bufnr))
+              end
             end
           end
         end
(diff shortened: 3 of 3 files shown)


greenhat:
@ibhagwan I added `workspace_diag_only_cwd` to `config.lua`. I was thinking it should make this option more discoverable.  Should it be mentioned in readme as well?

ibhagwan:
Good idea, I’m thinking 2 minor changes, perhaps it’s better to keep the variable name consistent as with other functions that have this functionality,  for example this is named `cwd_only` in `oldfiles`:

```lua
if opts.cwd_only then
    opts.cwd = vim.loop.cwd()
    local cwd = opts.cwd
    cwd = cwd:gsub([[\]],[[\\]])
    results = vim.tbl_filter(function(file)
      return vim.fn.matchstrpos(file, cwd)[2] ~= -1
    end, results)
  end
```

We can then use `opts.cwd` to filter the entries, this gives finer control to anyone who wishes to send any other folder as `cwd = ` when calling the function.

ibhagwan:
I took another look at the code this, the only thing that’s required is setting `opts.cwd = vim.loop.cwd()`, the `string .find()` isn’t required, it’s already taken care of inside [`make_entry_file`](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/core.lua#L96).

These are the only lines required and the existing code should take care of the rest:
```lua
if opts.cwd_only then
    opts.cwd = vim.loop.cwd()
end
```

greenhat:
> I took another look at the code, the only thing that’s required is setting `opts.cwd = vim.loop.cwd()`, the `string .find()` isn’t required, it’s already taken care of inside [`make_entry_file`](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/core.lua#L96).
> 
> These are the only lines required and the existing code should take care of the rest:
> 
> ```lua
> if opts.cwd_only then
>     opts.cwd = vim.loop.cwd()
> end
> ```
> 
> perhaps it’s even better if I add this globally to `make_entry_file`, this will enable this option for all functions that use `core.fzf_files`.

Thanks! I switched the code to `cwd_only` version you suggested and also rebased it on the latest main branch. It does not work now. I assume it's ok since you want to filter non-cwd originated diagnostic messages in soon-to-be-updated `make_enty_file`.

ibhagwan:
> > I took another look at the code, the only thing that’s required is setting `opts.cwd = vim.loop.cwd()`, the `string .find()` isn’t required, it’s already taken care of inside [`make_entry_file`](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/core.lua#L96).
> > These are the only lines required and the existing code should take care of the rest:
> > ```lua
> > if opts.cwd_only then
> >     opts.cwd = vim.loop.cwd()
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > perhaps it’s even better if I add this globally to `make_entry_file`, this will enable this option for all functions that use `core.fzf_files`.
> 
> Thanks! I switched the code to `cwd_only` version you suggested and also rebased it on the latest main branch. It does not work now. I assume it's ok since you want to filter non-cwd originated diagnostic messages in soon-to-be-updated `make_enty_file`.

You're right it's not working now, I see I made an inconsistent mess between how I handle it in `buffers` vs `oldfiles` I guess we can just leave the original PR as is with the option name as `cwd_only` and I'll merge it. I can refactor the mess later into `make_entry_file`.

Sorry for the back and forth, I would have done this already myself but I don't want to rob you of the contribution :-)

greenhat:
> > > I took another look at the code, the only thing that’s required is setting `opts.cwd = vim.loop.cwd()`, the `string .find()` isn’t required, it’s already taken care of inside [`make_entry_file`](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/core.lua#L96).
> > > These are the only lines required and the existing code should take care of the rest:
> > > ```lua
> > > if opts.cwd_only then
> > >     opts.cwd = vim.loop.cwd()
> > > end
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > perhaps it’s even better if I add this globally to `make_entry_file`, this will enable this option for all functions that use `core.fzf_files`.
> > 
> > 
> > Thanks! I switched the code to `cwd_only` version you suggested and also rebased it on the latest main branch. It does not work now. I assume it's ok since you want to filter non-cwd originated diagnostic messages in soon-to-be-updated `make_enty_file`.
> 
> You're right it's not working now, I see I made an inconsistent mess between how I handle it in `buffers` vs `oldfiles` I guess we can just leave the original PR as is with the option name as `cwd_only` and I'll merge it. I can refactor the mess later into `make_entry_file`.
> 
> Sorry for the back and forth, I would have done this already myself but I don't want to rob you of the contribution :-)

No problem! I see you're running this project clean and tidy! It takes effort to get things done the right way. :) I restored the original filtering.

ibhagwan:
I realized I can squash it myself with "squash and merge", ty for the contribution!

ibhagwan:
FYI, I moved the `cwd_only` filtering into `make_entry_file`, it's now possible to filter any files provider to display results only from the current working directory, that includes all lsp providers too (diagnostics, symbols, references, etc).

greenhat:
> FYI, I moved the `cwd_only` filtering into `make_entry_file`, it's now possible to filter any files provider to display results only from the current working directory, that includes all lsp providers too (diagnostics, symbols, references, etc).

Thank you for the update! I find "only cwd" filters for references and symbols doing more harm in Rust project. Let me explain with a sample workflow. When I jump to definition into some external dependency (crate) new lsp client is spawned for this crate's directory and "polluting" workspace diagnostics list. However, references and symbols are working as expected while I'm in that crate's file, meaning that workspace symbols only show that crate's symbols, and references behave similarly. And when I jump back to my project(cwd) references and symbols from external dependencies don't "pollute" references and symbols lists. I mean that symbols and references lists did not have this problem of "polluting" local project views with external data as diagnostics had. Now, with "only cwd" enabled it shows empty references list and symbols while I'm in that external crate. I understand that lsp settings should not be Rust-centric but maybe having one global "only cwd" filtering for lsp is not ideal and better to let it be configured for each diagnoctic/references/symbols command separately.

ibhagwan:
> > FYI, I moved the `cwd_only` filtering into `make_entry_file`, it's now possible to filter any files provider to display results only from the current working directory, that includes all lsp providers too (diagnostics, symbols, references, etc).
> 
> Thank you for the update! I find "only cwd" filters for references and symbols doing more harm in Rust project. Let me explain with a sample workflow. When I jump to definition into some external dependency (crate) new lsp client is spawned for this crate's directory and "polluting" workspace diagnostics list. However, references and symbols are working as expected while I'm in that crate's file, meaning that workspace symbols only show that crate's symbols, and references behave similarly. And when I jump back to my project(cwd) references and symbols from external dependencies don't "pollute" references and symbols lists. I mean that symbols and references lists did not have this problem of "polluting" local project views with external data as diagnostics had. Now, with "only cwd" enabled it shows empty references list and symbols while I'm in that external crate. I understand that lsp settings should not be Rust-centric but maybe having one global "only cwd" filtering for lsp is not ideal and better to let it be configured for each diagnoctic/references/symbols command separately.

That’s why it’s disabled globally by default, it’s just an option if you wish, I also wouldn’t enable it globally to all LSP functions, just use it per function and map your keybind to the function with the parameter:

```lua
:lua require”fzf-lua”.lsp_workspace_diagnostics({ cwd_only = true })
— or
:FzfLua lsp_workspace_diagnostics cwd_only=true
```

greenhat:
> > > FYI, I moved the `cwd_only` filtering into `make_entry_file`, it's now possible to filter any files provider to display results only from the current working directory, that includes all lsp providers too (diagnostics, symbols, references, etc).
> > 
> > 
> > Thank you for the update! I find "only cwd" filters for references and symbols doing more harm in Rust project. Let me explain with a sample workflow. When I jump to definition into some external dependency (crate) new lsp client is spawned for this crate's directory and "polluting" workspace diagnostics list. However, references and symbols are working as expected while I'm in that crate's file, meaning that workspace symbols only show that crate's symbols, and references behave similarly. And when I jump back to my project(cwd) references and symbols from external dependencies don't "pollute" references and symbols lists. I mean that symbols and references lists did not have this problem of "polluting" local project views with external data as diagnostics had. Now, with "only cwd" enabled it shows empty references list and symbols while I'm in that external crate. I understand that lsp settings should not be Rust-centric but maybe having one global "only cwd" filtering for lsp is not ideal and better to let it be configured for each diagnoctic/references/symbols command separately.
> 
> That’s why it’s disabled globally by default, it’s just an option if you wish, I also wouldn’t enable it globally to all LSP functions, just use it per function and map your keybind to the function with the parameter:
> 
> ```lua
> :lua require”fzf-lua”.lsp_workspace_diagnostics({ cwd_only = true })
> — or
> :FzfLua lsp_workspace_diagnostics cwd_only=true
> ```

Yeah, my bad, I should’ve read it more carefully. This looks great!

PR #38, opened 2021-08-10
Optimize git performance

jdrouhard:
1) Faster method to determine if in git repo
2) Bypass entirely if git icons disabled

These two changes make fzf-lua perform _much_ faster when run in large codebases. `git status` takes between half a second a second to run on our large codebase, but this change makes fzf-lua nearly as snappy as fzf.vim.


ibhagwan:
I see, I didn’t realize the git status can be an issue in large codebases, ty!

PR #40, opened 2021-08-11
remove `vim.lsp.buf_get_clients` check in diagnostics;

greenhat:
It returns an empty list after re-sourcing nvim's config (BTW, why?). I found a similar check in `check_capabilities` which should suffice.

### DIFF
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -305,12 +305,6 @@ end
 M.diagnostics = function(opts)
   opts = normalize_lsp_opts(opts, config.globals.lsp)
 
-  local lsp_clients = vim.lsp.buf_get_clients(0)
-  if #lsp_clients == 0 then
-    utils.info("LSP: no client attached")
-    return
-  end
-
   opts.winid = vim.api.nvim_get_current_win()
   local lsp_type_diagnostic = vim.lsp.protocol.DiagnosticSeverity
   local current_buf = vim.api.nvim_get_current_buf()


ibhagwan:
I’m not sure if I want to remove this, running lsp functions with no clients attached can cause hangs and other weird stuff, instead of removing it I’d rather focus on why this happens, especially when the issue is only after resource’ing the config.

As for why it happens idk, it happens to me too and is not isolated to fzf-lua, I also lose the LSP indicators in my status line (feline.nvim), for that specific reason [I have a hook in my config reload function which restarts the LSP after a reload](https://github.com/ibhagwan/nvim-lua/blob/main/lua/plugin/nvim-reload/init.lua#L34).

greenhat:
My understanding was that this check for LSP clients attached is made in `wrap_module_fncs` including `diagnostics_handler`. If it's not the case then, yes it makes sense to leave the check there.

Hey! Your config is very well organized and neat! I'm going to steal it and build mine the same way!

ibhagwan:
> My understanding was that this check for LSP clients attached is made in `wrap_module_fncs` including `diagnostics_handler`. If it's not the case then, yes it makes sense to leave the check there.

The check doesn’t happen for both `diagnostics` and `workspace_diagnostics` as both functions have [`capability` set to nil](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/providers/lsp.lua#L528), which then [skips the check at line 550](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/providers/lsp.lua#L550), this is done because the LSP doesn’t have diagnostics in `client.resolved_capabilities`.

I probably could have made `check_capabilities` work regardless and check for an attached client for diagnostics but I wrote diagnostics lasts and decided this minimal code duplication wasn’t so terrible and left it this way.

> Hey! Your config is very well organized and neat! I'm going to steal it and build mine the same way!

That’s why I posted it, good luck :)

greenhat:
Hey! I did not succeed with reloading hook, but I mostly need config reloads
for keymap updates, so I extracted them in a separate module and reload only
this keymaps.
This PR can be closed.

ibhagwan:
Ty for the update!

Issue #51, opened 2021-08-24
Actions for git commits/b_commits 


jpe90:
Currently, the git_commits and  git_bcommits will let you inspect the diffs of each commit, but I didn't see any actions available with the results. Potential actions include:

- opening previous commits in a read only buffer
- checking out previous commits
- resetting to previous commits

Would it be feasible to add any of these?

ibhagwan:
Most certainly, this is actually quite easy to add yourself by creating a custom action function.

Some of these make sense to be shipped as default actions, if you wish to provide the specific git command line to perform some of these actions I can easily add these actions and have them as an option.

jpe90:
Since the UI seems to display abbreviated hash ids, I think the commands would be:
- git show [abbreviated commit hash]:[path to file] 
(it seems you need the ":[path]" or it will show the diff rather than the full file)
- git checkout [abbreviated commit hash]
- git reset --soft [abbreviated commit hash]
- git reset --soft [abbreviated commit hash]

ibhagwan:
Ty, I’ll add these as soon as I’m done with my current task (builtin previewer).

jpe90:
I'm happy to give it a try this weekend if you would like, I haven't worked with lua or neovim plugin development before but I wouldn't mind giving it a shot. No worries if you would rather implement it yourself though.

jpe90:
I made an attempt at checkout functionality and opening previous commits and added them as defaults. I figured it made sense to have checkout functionality associated with project-wide commits and viewing an old version of a file to be associated with buffer commits. Let me know what you think.

ibhagwan:
Your PR was just tested and merged, I believe we can close this issue.

PR #55, opened 2021-08-28
Add actions to checkout and inspect previous git commits

jpe90:
Addresses #51. Implements opening previous commits in a read only buffer/split/tab and corresponding actions for b_commits. Also implements checkout action for git commits.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -216,14 +216,16 @@ end
 
 M.git_checkout = function(selected)
   local commit_hash = selected[1]:match("[^ ]+")
-  local current_commit = vim.fn.systemlist("git rev-parse --short HEAD")
-  if(commit_hash == current_commit) then return end
-  local output = vim.fn.systemlist("git checkout " .. commit_hash)
-  if utils.shell_error() then
-    utils.err(unpack(output))
-  else
-    utils.info(unpack(output))
-    vim.cmd("edit!")
+  if vim.fn.input("Checkout commit " .. commit_hash .. "? ") == "y" then
+    local current_commit = vim.fn.systemlist("git rev-parse --short HEAD")
+    if(commit_hash == current_commit) then return end
+    local output = vim.fn.systemlist("git checkout " .. commit_hash)
+    if utils.shell_error() then
+      utils.err(unpack(output))
+    else
+      utils.info(unpack(output))
+      vim.cmd("edit!")
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


ibhagwan:
Tysm this looks great!

I’m almost done with the [builtin previewer](https://github.com/ibhagwan/fzf-lua/tree/builtin_preview), when I’m done with it I’ll test this and add it.

I do have one concern though, do you think it would be beneficial to add a confirmation prompt for the the checkout when it’s the default action?

jpe90:
Yeah you're right, a confirmation prompt is a good idea. I'll try to figure out how to add one in and let you know if I get stuck.

ibhagwan:
This is how I do it for `grep`:

```lua
  if not opts.search or #opts.search == 0 then
    opts.search = vim.fn.input(opts.input_prompt)
  end

  if not opts.search or #opts.search == 0 then
    utils.info("Please provide a valid search string")
    return
  end
```

jpe90:
While adding that prompt I noticed that an issue with the git operation will only flash in the message area for like half a second (mine failed because my local changes would be overridden), so I need to address that as well.

ibhagwan:
> While adding that prompt I noticed that an issue with the git operation will only flash in the message area for like half a second (mine failed because my local changes would be overridden), so I need to address that as well.

Looks great!

ibhagwan:
Tysm for this wonderful contribution, works great!

One issue I noticed (and fixed) was that the temp bcommit buffers weren't wiping after being hidden, I added this to make sure they disappear when they are no longer in use `vim.api.nvim_buf_set_option(buf, 'bufhidden', 'wipe')`

jpe90:
No problem! Thanks for the guidance and for catching that issue.

ibhagwan:
I must confess I’m a bit of a git noob, I didn’t even know it was possible to checkout a specific commit with such ease, I already used this to test an earlier commit, such a great addition, ty!

ibhagwan:
FYI,  https://github.com/ibhagwan/fzf-lua/commit/7b9fe4495e5777ea4f506778fab39dd6b304a0d4 added more changes to the git actions to be able to support running the git commands in a different working directory, for example, `lua require('fzf-lua').git_status({ cwd = "~/Sources/nvim/fzf-lua" })`, this works now with all commands (files, status, commits, branches).

While doing so I found out there's a bug in the `checkout` action, subsequent checkouts don't work, let's say you checkout the 4th previous commit and then want to checkout the 2nd one (or master) it doesn't work, I'll debug it later when I have time but just in case  you want to take a stab at this :-)

jpe90:
Sure, I'll make an issue  and address it in a PR to help keep track of everything.

PR #60, opened 2021-08-29
Update README.md

saltsucker:
Added example for lua keymaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ which can be easily mapped to:
 ```vim
 nnoremap <c-P> <cmd>lua require('fzf-lua').files()<CR>
 ```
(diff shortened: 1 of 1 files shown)


ibhagwan:
Great addition, perhaps just change `opts` to `{ noremap = true, silent = true } `?

ibhagwan (changes requested):
Replace opts with `{ noremap = true, silent = true }`

ibhagwan:
I apologize for not merging this, I requested the change but there was reply on your end for a while and I felt this was important.

https://github.com/ibhagwan/fzf-lua/commit/d8b5981e9240ef4684500d5e64988e60ef28b922

PR #61, opened 2021-08-29
Set cursorlineopt=both in the builtin previewer

fsouza:
This way, if the user has defined cursorlineopt=number elsewhere, we
override it to behave as intended.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -96,6 +96,7 @@ function Previewer:gen_winopts()
     number          = true,
     relativenumber  = false,
     cursorline      = true,
+    cursorlineopt   = 'both',
     cursorcolumn    = false,
     signcolumn      = 'no',
     foldenable      = false,


ibhagwan:
Tysm for this I can already see myself struggling for maybe a couple of hours struggling with an issue why someone can’t see the cursor line, you just saved me those hours :)

ibhagwan:
Ok, I just tested I can't merge this as is, this option doesn't exist in the 0.5 release, I can only find it in the nightly version, I have to find a way to first check if the option exists before trying to set it will spawn the error:

```lua
Error executing vim.schedule lua callback:
...ack/packer/opt/fzf-lua/lua/fzf-lua/previewer/builtin.lua:112: Invalid option name: 'cursorlineopt'
```

ibhagwan:
Merged, solved this by using `vim.fn.exists(&option)` to verify the option exists before setting it.

fsouza:
@ibhagwan ohhh my bad, I didn't realize this was a 0.6+ thing. Thanks for fixing it!

PR #69, opened 2021-08-30
Add support for fzf --color option

krady21:
`fzf.vim` provides the global option `fzf_colors`, which allows the user to match some of the fzf highlight groups with vim highlight groups. Without `--color`, the fzf colorscheme doesn't always acceptably match the vim one (for example when using light themes). Setting the enviroment option `$FZF_DEFAULT_OPS` can help, but only if you are not changing vim colorschemes often. This PR should help in this regard. We should also probably discuss which vim highlight groups should match to which fzf `--color` options.

Great plugin, btw! I've decided to switch to it completely today, after using `fzf.vim` for a while now and it's been nothing but smooth sailing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,7 +217,8 @@ require'fzf-lua'.setup {
     hl_border        = 'FloatBorder',   -- window border color
   },
   -- fzf_bin             = 'sk',        -- use skim instead of fzf?
-  fzf_layout          = 'reverse',      -- fzf '--layout='
+  fzf_colors          = nil,            -- fzf '--color=' options
+  fzf_layout          = 'reverse',      -- fzf '--layout=' options
   fzf_args            = '',             -- adv: fzf extra args, empty unless adv
   fzf_binds           = {               -- fzf '--bind=' options
       'f2:toggle-preview',
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -35,6 +35,7 @@ M.globals = {
   },
   default_prompt      = '> ',
   fzf_bin             = nil,
+  fzf_colors          = nil,
   fzf_layout          = 'reverse',
   fzf_binds = {
     -- <F2>        toggle preview
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -29,6 +29,7 @@ M.fzf = function(opts, contents, previewer)
       end
     end
   end
+
   fzf_win:attach_previewer(previewer)
   fzf_win:create()
   local selected = fzf.raw_fzf(contents, M.build_fzf_cli(opts),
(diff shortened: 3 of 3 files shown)


ibhagwan:
Tysm this is great, and probably one of the only parts of customizations I left out.

Regarding `vim.env`: at first I used the `$FZF_DEFAULT_OPTS` for some options but I decided against it as it can behave unpredictably with different user setups and shell terminals.

Manual setup as you added is my preferred way, if we are to go through this route we should probably move the hard-coded highlights into a section of the config so users can modify them.

In addition, since most users don't mind the defaults and probably won't mess with this I would avoid sending the `--color` option unless the user specifically configured highlights in the `setup` function.

krady21:
One more question. Would you want to keep some sort of compatibility to how `g:fzf_colors` worked in fzf.vim?
Would you prefer setting the option like this:
```
require'fzf-lua'.setup {
  fzf_colors = {
    ["fg"] = { "fg", "Normal" },
    ["bg"] = { "bg", "Normal" },
    ["hl"] = { "fg", "Comment" },
    ["fg+"] = { "fg", "Normal" },
    ["bg+"] = { "bg", "CursorLine" },
    ["hl+"] = { "fg", "Statement" },
    ["info"] = { "fg", "PreProc" },
    ["prompt"] = { "fg", "Conditional" },
    ["pointer"] = { "fg", "Exception" },
    ["marker"] = { "fg", "Keyword" },
    ["spinner"] = { "fg", "Label" },
    ["header"] = { "fg", "Comment" },
    ["gutter"] = { "bg", "Normal" },
  }
}
```

or like this:
```
require'fzf-lua'.setup {
  fzf_colors_fg = {"fg", "Normal"},
  fzf_colors_bg = {"bg", "Normal"}
  fzf_colors_hl = {"fg", "Comment"}
...etc
}
```
I tend to prefer the first version, but would also like your feedback.

ibhagwan:
> I tend to prefer the first version, but would also like your feedback.

I’m with you on that, I also prefer the first.

krady21:
You can take a look at the PR now. Here's how the changes look like with the settings described as above.

Without fzf_colors:
Screenshot from 2021-09-07 12-55-53
With fzf_colors:
Screenshot from 2021-09-07 12-56-28

ibhagwan:
The only changes I would do is make it optional to send different colors for each command and not be bound by just the globals, this way I'm also able to send separate `fzf_colors` array to a customized function:

```diff
diff --git a/lua/fzf-lua/core.lua b/lua/fzf-lua/core.lua
index cd44b8c..552a13d 100644
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -103,7 +103,6 @@ M.build_fzf_cli = function(opts, debug_print)
     -- support the '--info=' flag
     opts.fzf_info = utils._if(opts._is_skim, "--inline-info", "--info=inline")
   end
-  opts.fzf_colors = M.create_fzf_colors(config.globals.fzf_colors)

   local cli = string.format(
     [[ %s %s --layout=%s --bind=%s --prompt=%s]] ..
@@ -111,7 +110,7 @@ M.build_fzf_cli = function(opts, debug_print)
     [[ --height=100%%]] ..
     [[ %s %s %s %s %s %s]],
     opts.fzf_args or config.globals.fzf_args or '',
-    opts.fzf_colors,
+    M.create_fzf_colors(opts.fzf_colors or config.globals.fzf_colors),
     opts.fzf_layout or config.globals.fzf_layout,
     utils._if(opts.fzf_binds, opts.fzf_binds,
       vim.fn.shellescape(table.concat(config.globals.fzf_binds, ','))),
```

As a side note, the original version was also bugging out on me when defining colors and and raising an issue inside the `preview_window` function (???), very puzzling but I decided against debugging the issue and just making this change instead.

ibhagwan:
> As a side note, the original version was also bugging out on me when defining colors and and raising an issue inside the preview_window function (???), very puzzling but I decided against debugging the issue and just making this change instead.

I figured it out, there's a table merge inside the `preview_window` function with the `keep` option specified, which will eventually also try to merge `opts.fzf_colors` with `globals.fzf_colors`, after your change, they will be different types and raise an error (trying to merge the global table into a string), so either use my diff or define the new fzf color args as `opts._fzf_colors` (or any other name you want).

ibhagwan:
Nevermind, already merged and fixed, tysm for this fantastic contribution!

krady21:
> As a side note, the original version was also bugging out on me when defining colors and and raising an issue inside the `preview_window` function (???), very puzzling but I decided against debugging the issue and just making this change instead.

This was also bugging me for some time now and was one of the reasons the PR took so long.

> Nevermind, already merged and fixed, tysm for this fantastic contribution!

Glad I could be of help!

PR #72, opened 2021-08-30
Fix empty line in the builtin previewer

fsouza:
Whenever the file ends with newline, the builtin previewer was adding an
empty line at the end of the file. That's a side effect splitting and
rejoining the lines with \n. Basically, when we split a\nb\nc\n we get
{'a', 'b', 'c', ''}, then nvim_buf_set_lines treats each item in the
table as a line, and inserts an empty line at the end.

We do need to account for files that don't end with a new line though,
hence the check.

I don't have a way to test the \r\n logic locally, no Windows hosts
around, but it should work too.

Before:

image

(line 23 doesn't really exist, file only has 22 lines: https://github.com/fsouza/dotfiles/blob/main/.github/dependabot.yml)

After:

image

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -314,7 +314,16 @@ function Previewer:display_entry(entry)
       if not vim.api.nvim_buf_is_valid(bufnr) then
         return
       end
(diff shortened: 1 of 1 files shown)


ibhagwan:
It was annoying me too and I was going to fix it, you just saved me some time, ty! :-)

Btw, no need to worry about windows, `nvim-fzf` doesn't support windows and I have no intentions of installing this privacy abusing OS to create an equivalent windows RPC interface to fzf.

PR #83, opened 2021-09-01
Implement support for columns in vimcmd_file

fsouza:
Something I noticed while using `rg` locally with `--column`.

Let me know what you think.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -42,12 +42,23 @@ end
 M.vimcmd_file = function(vimcmd, selected)
   if not selected or #selected < 2 then return end
   for i = 2, #selected do
-    -- check if the file contains line
-    local file, line = selected[i]:match("^([^:]+):(%d+)")
-    if file and line then
-      vim.cmd(string.format("%s +%s %s", vimcmd, line, vim.fn.fnameescape(file)))
+    -- check if the file contains line and col or just line.
+    local file, line, col = selected[i]:match("^([^:]+):(%d+):(%d+):")
+    if not file then
+      file, line = selected[i]:match("^([^:]+):(%d+):")
+      if file then
+        col = 1
+      else
+        file = selected[i]
+      end
+    end
+
+    file = vim.fn.fnameescape(file)
+    print(file, line, col)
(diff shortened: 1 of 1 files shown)


lua/fzf-lua/actions.lua
```diff
@@ -42,12 +42,23 @@ end
 M.vimcmd_file = function(vimcmd, selected)
   if not selected or #selected < 2 then return end
   for i = 2, #selected do
-    -- check if the file contains line
-    local file, line = selected[i]:match("^([^:]+):(%d+)")
-    if file and line then
-      vim.cmd(string.format("%s +%s %s", vimcmd, line, vim.fn.fnameescape(file)))
+    -- check if the file contains line and col or just line.
+    local file, line, col = selected[i]:match("^([^:]+):(%d+):(%d+):")
```
fsouza, line 46: Unfortunately Lua doesn't support optional group captures in patterns 😭

ibhagwan:
> Unfortunately Lua doesn't support optional group captures in patterns sob

Yes :/ it feels so ugly to have to do so many `:match`es, I was going to refactor the actions anyways and send `entry { file = .., bufnr = ... }` etc so that it's easier and clearer how to create new actions.

lua/fzf-lua/actions.lua
```diff
@@ -42,12 +42,23 @@ end
...
+      file, line = selected[i]:match("^([^:]+):(%d+):")
+      if file then
+        col = 1
+      else
+        file = selected[i]
+      end
+    end
+
+    file = vim.fn.fnameescape(file)
+    print(file, line, col)
```
fsouza, line 57: Ops this was me debugging. Will fix

ibhagwan:
I implemented this more simply using already existing code in `path.entry_to_file`, if the results now contain a column it will set the cursor to the exact spot, feels right :-)

Sorry I couldn't merge this but we got the solution nonetheless, take a look here https://github.com/ibhagwan/fzf-lua/commit/2ae75cba9691df07cb4a0f1981bd21ac2402c814

fsouza:
Not a problem, if it works it works! :)

Issue #86, opened 2021-09-02
Unable to check out commits newer than where HEAD is pointing


jpe90:
If you check out an old commit, and then want to check out a newer one, you are unable to do so. See [here](https://github.com/ibhagwan/fzf-lua/pull/55#issuecomment-910856166) for more info.

jpe90:
Closing now that PR was merged

PR #87, opened 2021-09-02
Fix loss of git records after checking out older commits

jpe90:
Addresses #86 by adding --reflog flag to the "git log" command which populates the list of commits. See [reflog docs](https://git-scm.com/docs/git-reflog) for more info.


ibhagwan:
Such a neat fix tysm!

jpe90:
Gah! Just noticed I missed the README again and pushed the commit a minute too late :^)

PR #90, opened 2021-09-02
Improve git commands so they always run in the git root directory

jdrouhard:
This matches the behavior of both fzf.vim and telescope.nvim.

It also fixes a bug with the "git_buf_edit" action if the cwd is not in
the git root.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -252,20 +252,22 @@ end
 
 M.git_buf_edit = function(selected, opts)
   local cmd = path.git_cwd("git show ", opts.cwd)
+  local git_root = path.git_root(opts.cwd, true)
   -- there's an empty string in position 1 for some reason?
   table.remove(selected,1)
   local win = vim.api.nvim_get_current_win()
   local buffer_filetype = vim.bo.filetype
-  local file = path.relative(vim.fn.expand("%"), vim.loop.cwd())
+  local file = path.relative(vim.fn.expand("%:p"), git_root)
   local commit_hash = selected[1]:match("[^ ]+")
   local git_file_contents = vim.fn.systemlist(cmd .. commit_hash .. ":" .. file)
-  local buf = vim.api.nvim_create_buf(true, false)
-  local file_name = string.gsub(file,"%.","[" .. commit_hash .. "]%.")
+  local buf = vim.api.nvim_create_buf(true, true)
+  local file_name = string.gsub(file,"$","[" .. commit_hash .. "]")
   vim.api.nvim_buf_set_lines(buf,0,0,true,git_file_contents)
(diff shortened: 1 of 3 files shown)


ibhagwan:
Have you seen this commit from yesterday?

https://github.com/ibhagwan/fzf-lua/commit/7b9fe4495e5777ea4f506778fab39dd6b304a0d4

jdrouhard:
Yep, this builds directly on top of that. That commit didn't make it work if the cwd is deeper in a git repo than the root, and there were still a couple bugs.

ibhagwan:
Oh I see this change is reflected after the commit, I’m not sure I understand what this does, when I tested the `git_buf_edit` it worked even with supplying `cwd` as an argument.

> This matches the behavior of both fzf.vim and telescope.nvim.

Can you expand a bit in what way the current behavior fzf.vim and Telescope.

ibhagwan:
> Yep, this builds directly on top of that. That commit didn't make it work if the cwd is deeper in a git repo than the root, and there were still a couple bugs.

Oh I see, it didn’t work with nested repos?

jdrouhard:
Not nested repos, but if your current working directory is in _subdirectory_ of the git repo, then using `git_files()` (for instance) would only list files in your cwd and below. This is also what was broken with the `git_buf_edit` action (when cwd is deeper than the gir root). Telescope.nvim and fzf.vim both show you all files in the current repo, regardless of where the cwd is (or the target cwd is if you specify it manually).

This commit changes it so that the fzf itself runs in the git toplevel dir, which lets `git ls-files` (the command for `git_files()` list all files in the repo. It also updates the `git_cmd` helper to just use `-C <cwd>` which applies the git lookup logic to find the git toplevel dir for the given dir.

ibhagwan:
> Not nested repos, but if your current working directory is in _subdirectory_ of the git repo, then using `git_files()` (for instance) would only list files in your cwd and below. This is also what was broken with the `git_buf_edit` action (when cwd is deeper than the git root). Telescope.nvim and fzf.vim both show you all files in the current repo, regardless of where the cwd is (or the target cwd is if you specify it manually).
> 
> This commit changes it so that the fzf itself runs in the git toplevel dir, which lets `git ls-files` (the command for `git_files()` list all files in the repo. It also updates the `git_cmd` helper to just use `-C <cwd>` which applies the git lookup logic to find the git toplevel dir for the given dir.


Oh I see, did you test it with supplying `cwd` as an argument as well? Does the `git_icons` still work when it’s a sub directory?

jdrouhard:
> Oh I see, did you test it with supplying cwd as an argument as well? Does the git_icons still work when it’s a sub directory?

Yes and yes :)

ibhagwan:
I love the small optimizations you did with the code, tysm for your amazing contributions!

PR #94, opened 2021-09-03
Builtin previewer: improve mechanism for backing up/restoring winopts

jdrouhard:
See :help local-options for more details, but essentially, we have to be
very careful with how we set window-local options on popup windows,
because buffers might take some window-local options from the last
window they were displayed in with them to the new window in certain
circumstances.

This commit changes the backup/restore for the window options so that
the possible window-local options that might get clobbered are saved
from the original active window and reapplied to the preview window just
before it's closed. This ensures any buffers loaded from the picker will
have the same window options applied that were in the active window when
fzf was opened.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -72,11 +72,7 @@ function Previewer:new(o, opts, fzf_win)
 end
 
 function Previewer:close()
-  -- restore winopts backup for those that weren't restored
-  -- (usually the last previewed loaded buffer)
-  for bufnr, _ in pairs(self.backups) do
-    self:restore_winopts(bufnr, self.win.preview_winid)
-  end
+  self:restore_winopts(self.win.preview_winid)
   self:clear_preview_buf()
   self.backups = {}
   _self = nil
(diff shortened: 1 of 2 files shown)


ibhagwan:
Very interesting, based on my inspection of the `nvim-bqf` code I thought the options needed to be saved per buffer,win combo but this makes more sense, so basically all we need is to save the sending window option and reset the options before closing of the preview?

One comment, I saw you removed the `if utils.nvim_has_option(opt)` statement from `restore_ winopts`, I assume you did that because if the option was backed up, the restore is pretty much guaranteed to succeed just wanted to make sure this is on purpose and there would be no side effects, for example with the new 0.6 option `cursorlineopt   = 'both'` (trying to set this option on a 0.5 instance will cause an error).

jdrouhard:
> One comment, I saw you removed the if utils.nvim_has_option(opt) statement from restore_ winopts, I assume you did that because if the option was backed up, the restore is pretty much guaranteed to succeed just wanted to make sure this is on purpose and there would be no side effects, for example with the new 0.6 option cursorlineopt = 'both' (trying to set this option on a 0.5 instance will cause an error).

Yeah, the option was already checked before it was saved in the `backups` dict, so only options that exist are there.

I was playing around with the buffer picker and kept noticing that occasionally a window would lose a bunch of window options, it looked like it was applying the default "popup window" options to the active window when the previewer was closed occasionally. This PR basically makes the previewer work the way telescope.nvim's builtin previewer works.

ibhagwan:
Tysm for **yet another** core contribution!

jdrouhard:
> Tysm for **yet another** core contribution!

Thank you for writing an awesome Lua-based fzf plugin! It's nearly perfect, there are just little things I've noticed as I'm using it in my daily workflow so I fix them as I come across them! I probably have a couple more I'll have in the next few days, but these last two were bigger ones.

ibhagwan:
Thank you for you kind words, I also feel it's getting close to fulfilling all my neovim needs, can't wait for your next contributions :-)

PR #100, opened 2021-09-04
Improve performance of finding extension from an entry

jdrouhard:
Recent changes to support filenames with spaces caused a performance
regression. This should get things back up to speed.

### DIFF
--- a/lua/fzf-lua/path.lua
+++ b/lua/fzf-lua/path.lua
@@ -44,12 +44,8 @@ function M.basename(path)
 end
 
 function M.extension(path)
-  -- path = M.basename(path)
-  -- return path:match(".+%.(.*)")
-  -- 1. match anything before the first ':'
-  -- 2. greedy match anything after the last dot
-  -- 3. remove coloring escape sequence
-  return utils.strip_ansi_coloring(path:match("[^:]+"):match("[^.]*$"))
+  _, _, ext = string.find(path, ".*%.([^.:]*):?")
+  return utils.strip_ansi_coloring(ext)
 end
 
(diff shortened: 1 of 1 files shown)


jdrouhard:
Sorry, accidentally deleted the other branch.

ibhagwan:
No worries, I'll test it with all the edge cases just in case, took me a while to get this one right :-)

ibhagwan:
Unfortunately, this has an issue when parsing entries returned by `rg|grep`, the entries are returned with text which if contains '.' could return the wrong result:

```lua
string.match("file.ext:100:55 matching.text -- foo", ".*%.([^.:]*):?")
-- results in "text -- foo"
```

Another current behavior (which I can accommodate in the code) is the function never returns nil, if the file doesn't contain a '.' the result will be the file name, not sure how much I like this behavior but if this is to be changed I'm gonna have to make sure the code knows how to handle `nil` as a result.

jdrouhard:
I'll tweak it a bit more. I'm sure I can make the one regex return the right result and nil if not found.

ibhagwan:
Note that I had to update the regex to not croak on empty string input so the current branch conflicts, should probably rebase before working on the new regex.

As a note, I also wanted to have one regex “to rule them all” but lua pattern matching is somewhat limited so I ended up with two I believe rather simple ones that do the work:

- If string contains `:`, remove everything past the `:`, otherwise return the string itself (this removes all the `grep` match text)
- Greedy match anything past the last dot (to make sure it’s working with filenames with multiple dots)

jdrouhard:
Closing in favor of #108

PR #104, opened 2021-09-05
Add file_icon_padding option

### LINKED ISSUE #98: Add an option to add spaces for double width file icons
Hi, I would like to see this option implemented, I use kitty terminal which can render double width icons when they have extra space.

Current behavior
image

Expected behavior 
image

EdwarKenedy:
Fixes #98 
This option helps kitty terminal users to render double-width icons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,7 @@ require'fzf-lua'.setup {
   file_icon_colors = {
     ["lua"]   = "blue",
   },
+  file_icon_padding = '',
 }
 ```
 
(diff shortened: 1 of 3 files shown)


ibhagwan:
Merged, ty!

Note that I modified this a tiny bit, replacing all spaces in the padding with an invisible unicode character from `utils.nbsp`.

Probably wouldn't have been an issue but since some of the commands use fzf fields (which are sometimes delimited by spaces) it's better to treat the icon+padding+filename as one non-spaced object.

PR #107, opened 2021-09-05
Very minor README change

hqkhan:
Git command update in README: `git_branch` -> `git_branches`. The command in `builtin` shows the correct name already which is `git_branches`.


ibhagwan:
Ty! :)

PR #108, opened 2021-09-06
File icons: use colors from nvim-web-devicons if available

jdrouhard:
This improves the colors for file icons if nvim-web-devicons is available. If it is, fzf-lua now uses colors defined by the plugin for the file icons.

This should also improve the performance of `make_entry_file()` by simplifying the regex(es) applied to each line from the callback. It avoids pattern matching with the end of string character (`$`) which in my testing is slower than looping through matches with `gmatch`. It also fixes a bug where ansi escape codes could affect cwd/relative path logic.

This is an alternate attempt to fix the same issue as #100 but also improves the colors.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -442,8 +442,24 @@ M.globals.nvim = {
       },
     },
   }
+
 M.globals.file_icon_padding = ''
-M.globals.file_icon_colors = {
+
+if M._has_devicons then
+  M.globals.file_icon_colors = {}
+
+  local function hex(hex)
+    local r,g,b = hex:match('.(..)(..)(..)')
+    r, g, b = tonumber(r, 16), tonumber(g, 16), tonumber(b, 16)
+    return r, g, b
+  end
+
+  for ext, info in pairs(M._devicons.get_icons()) do
+    local r, g, b = hex(info.color)
+    utils.add_ansi_code('DevIcon' .. info.name, string.format('[38;2;%s;%s;%sm', r, g, b))
+  end
+else
+  M.globals.file_icon_colors = {
     ["lua"]       = "blue",
     ["rockspec"]  = "magenta",
     ["vim"]       = "green",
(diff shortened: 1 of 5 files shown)


ibhagwan:
Tysm, I really like i, I’ll have to do some serious testing though as it’s such a core change in the logic.

I also thought about using `%w` in the extension regex but I thought it would create issues on any filenames in different languages with Unicode chars - do you know if that would work for examples in files with Chinese glyphs?

jdrouhard:
> Tysm, I really like i, I’ll have to do some serious testing though as it’s such a core change in the logic.

Sure, put it through the paces. I think it'll work :)

> I also thought about using `%w` in the extension regex but I thought it would create issues on any filenames in different languages with Unicode chars - do you know if that would work for examples in files with Chinese glyphs?

I'm not 100% sure. I know that lua's pattern matching works on bytes, and since individual bytes in multi-byte glyphs aren't themselves alphanumeric, I'm guessing no. _However_, the extensions are only used for looking up icons with nvim-web-devicons, and the only extensions and file names in its database have alphanumeric extensions (or +, _, and $ characters).

ibhagwan:
>  _However_, the extensions are only used for looking up icons with nvim-web-devicons, and the only extensions and file names in its database have alphanumeric extensions (or +, _, and $ characters).

Good point, I’m trying to think if there are other uses for the extension detection and I think you’re right.

jdrouhard:
> >  _However_, the extensions are only used for looking up icons with nvim-web-devicons, and the only extensions and file names in its database have alphanumeric extensions (or +, _, and $ characters).
> 
> 
> 
> Good point, I’m trying to think if there are other uses for the extension detection and I think you’re right.

I just pushed a change that will make it match anything after the last '.' again and is quite a bit faster than the old method as well!

ibhagwan:
Great, I was just getting to testing this :-)

ibhagwan:
Absolutely love it :-)

Merged.

ibhagwan:
Btw, I had no idea it was possible to get all these colors in the term like this, tysm!

https://github.com/ibhagwan/fzf-lua/blob/d04dc95a9455d1157da2c18ff386da0c47403018/lua/fzf-lua/config.lua#L469

PR #117, opened 2021-09-07
More performance improvements 

jdrouhard:
Couple improvements here:

- `path.tail()` and `path.extension()` can just look backward from the end for the first separator and take a substring. No pattern matching is done and they only have to look at the number of characters that is the size of the result instead of the size of the input sub the size of the result. This is even faster than the current method.
- improve get_devicon() to be case-insensitive

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -41,7 +41,7 @@ end
 M.get_devicon = function(file, ext)
   local icon, hl
   if config._has_devicons and config._devicons then
-    icon, hl  = config._devicons.get_icon(file, ext, {default = true})
+    icon, hl  = config._devicons.get_icon(file, ext:lower(), {default = true})
   else
     icon, hl = '', 'dark_grey'
   end
--- a/lua/fzf-lua/path.lua
+++ b/lua/fzf-lua/path.lua
@@ -1,4 +1,5 @@
 local utils = require "fzf-lua.utils"
+local string_byte = string.byte
 
 local M = {}
 
@@ -11,22 +12,23 @@ M.starts_with_separator = function(path)
 end
 
 function M.tail(path)
-  local os_sep = M.separator()
-  local match_string = '[^' .. os_sep .. ']+'
+  local os_sep = string_byte(M.separator())
 
-  local tail = path
-  for s in string.gmatch(path, match_string) do
-      tail = s
+  for i=#path,1,-1 do
+    if string_byte(path, i) == os_sep then
+      return path:sub(i+1)
+    end
   end
-  return tail
+  return path
 end
 
 function M.extension(path)
-  local ext = path
-  for s in string.gmatch(path, '[^.]+') do
-      ext = s
+  for i=#path,1,-1 do
+    if string_byte(path, i) == 46 then
+      return path:sub(i+1)
+    end
   end
-  return ext
+  return path
 end
 
 function M.to_matching_str(path)
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -21,10 +21,11 @@ M.files = function(opts)
   if not opts then return end
   opts.cwd = path.git_root(opts.cwd)
   if not opts.cwd then return end
+  local make_entry_file = core.make_entry_file
   opts.fzf_fn = fzf_helpers.cmd_line_transformer(
     {cmd = opts.cmd, cwd = opts.cwd},
     function(x)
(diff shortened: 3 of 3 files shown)


ibhagwan:
I really love your optimizations, brings me back to how I would implement these functions in pure C :)

I’m wondering regarding the changes in “git.lua” - does saving  a local reference to `make_entry_file` increased the performance? Doesn’t lua have module caching? I would assume this shouldn’t make much difference as it’s simply a call to a metatable method or at the very worst a `core.__index` call?

jdrouhard:
> I really love your optimizations, brings me back to how I would implement these functions in pure C :)

Can't take all the credit, I saw this method in nvim-fzf for finding the last newline in a string. :) (I'm still a lua newbie so didn't know how to loop backward until now, heh).

> I’m wondering regarding the changes in “git.lua” - does saving  a local reference to `make_entry_file` increased the performance? Doesn’t lua have module caching? I would assume this shouldn’t make much difference as it’s simply a call to a metatable method or at the very worst a `core.__index` call?

Might be new to lua but I've been reading a lot about performance tweaking as I'm really sensitive to even slight latencies in my workflow. So, saving a local reference to the function makes using it in a loop much faster because lua has fast access to locals, but must do a table lookup on a module to get the function of used like that. This particular case probably didn't matter much, but that function is called per line of the program's output. Anyway, See https://www.lua.org/gems/sample.pdf for some great stuff.

ibhagwan:
Hi @jdrouhard, not sure if you're still using the plugin, thought you might be interested in the latest commit

https://github.com/ibhagwan/fzf-lua/commit/2d8a4e9afc7b32cdd7552f1ccd641c8bd6c2e85c

```help
major performance improvement: process entries externally, READ BELOW:
Since LUA is single threaded I reached a limit to performance
optimization, both 'git_icons' and 'file_icons' require string
matching and manipulations which eventually hurt performance
when running on large amount of files.
In order to solve that this commit introduces the option to spawn
commands and process the entries in a separate neovim process which
prints to stdio as if it was a regular shell command. This speeds up
things significantly and also makes the UI super responsive as if fzf
was run in the shell. This required a few lua hacks to be able to load
nvim-web-devicons in a '--headless --clean' instance and sharing the
user configuration through the RPC interface from the running instance.
This is enabled by default for 'files' and 'grep' providers and can also
be enabled for 'git.files' if required, control using the 'multiprocess'
option.
```

jdrouhard:
I do still use this plugin! (Constantly, I couldn't live without it). I did have to disable git_icons for my project, but after this commit, wow! Your commitment to this plugin and your users is just fantastic and very much so appreciated.

It still takes a slight blip for the results to start coming in, but the fact that the UI doesn't freeze anymore is absolutely amazing. I can start typing the file I know I'm looking for and it's responsive as the results start coming in. This was definitely the change I needed to be able to start using all the fancy features again!

One notable thing, the nvim-web-devicons hack doesn't seem to use the config overrides I have for a couple icons, any longer. See kyazdani42/nvim-web-devicons#80 for the workaround I'm using that doesn't currently work here anymore. Maybe it's time for me to make that PR to fix the upstream project so config overrides wouldn't be required.

ibhagwan:
Tysm for the kind words @jdrouhard!

>It still takes a slight blip for the results to start coming in

That's because it was still retreieving the git diff files in the "main" process, the commit below fixes that, the only remaining task the main process has is testing for git repo and disabling `git_icons` if it isn't a git repo, I could move that as well to the external process but I feel this is worthwhile to leave as is so I can disable git icons for non git repos increasing performance where possible.

If the `git rev-parse` test also creates a noticaeable delay (shouldn't IMO), I can add an option to offload it to the external process.

https://github.com/ibhagwan/fzf-lua/commit/828aceb3a3653c81622be4633c683708f49a105c

Try this commit and lmk?

>One notable thing, the nvim-web-devicons hack doesn't seem to use the config overrides I have for a couple icons, any longer. See kyazdani42/nvim-web-devicons#80 for the workaround I'm using that doesn't currently work here anymore. Maybe it's time for me to make that PR to fix the upstream project so config overrides wouldn't be required.

Yap, the devicons loading is a bif of a "hack" sending the devicons plugin path to the external process and then loading it with `loadfile`, that doesn't take into considration any use customization or setup functions, could be fixed but not sure if it's worth the extra complexity:

https://github.com/ibhagwan/fzf-lua/blob/828aceb3a3653c81622be4633c683708f49a105c/lua/fzf-lua/make_entry.lua#L70-L73

Issue #124, opened 2021-09-09
Improvement: possibility to define the action to use to jump to a single result for LSP providers


camilledejoye:
Hi,

First of all, thank for all your work!

I'm starting to play around with `fzf-lua` and one thing that I miss is the possibility to control how a single result will be opened.

Would you be open to make it configurable ?
Maybe add another option like `jump_to_single_result_action` to which we can associate any of the defined actions.

```lua
require('fzf-lua').lsp_definitions({
  async = false,
  jump_to_single_result = true,
  jump_to_single_result_action = require('fzf-lua.actions').file_vsplit,
)
```

It would require only little work in the LSP provider to:
* check if an action was provider
* if so, call the action with the formatted result (to be an entry)
* if not, keep using `vim.lsp.util.jump_to_location`

ibhagwan:
Was one of the first requests implemented, look at #13 

I really need to stop adding code and get to the documentation :)

ibhagwan:
My bad read this totally wrong.

camilledejoye:
Yeah it took me a while to find the `jump_to_single_result` ;)

ibhagwan:
> Yeah it took me a while to find the `jump_to_single_result` ;)

Ty for being so diligent, I'll implement this today :-)

ibhagwan:
Closed by PR #125, ty @camilledejoye!

PR #125, opened 2021-09-09
improvement: lsp provider uses actions

camilledejoye:
Implentation for https://github.com/ibhagwan/fzf-lua/issues/124

Usage example:
```lua
local fzf_lua = require('fzf-lua')
local actions = require('fzf-lua.actions')

fzf_lua.lsp_definitions({
  jump_to_single_result = true,
  jump_to_single_result_action = actions.file_vsplit
})
```


ibhagwan:
Simple and neat, I like it, tysm for the contribution!

PR #127, opened 2021-09-10
Change fzf_binds configuration style

krady21:
With the former configuration approach, when trying to add new bindings through `--bind`, one could end up with something like this :
```lua
opts.fzf_binds = {
    "f2:toggle_preview",  -- default binding
    "f2:abort", -- binding set by the user in the setup function
}
```
This way, the following string will be passed to the fzf cli: `--bind=f2:toggle_preview,f2:abort`.
With the new configuration approach the following string will be passed: `--bind=f2:abort`. 
This will avoid sending the default binding if the user overwrote it and also has the added benefit that the configuration style looks similar to the one found in other `fzf-lua` options. (i.e.: actions)


ibhagwan:
Ty! This change was long overdue, this option is one of the original ones that weren’t updated to the new style.

ibhagwan:
Merged, also added backward compatibility for users who already have it configured the old way.

Issue #126, opened 2021-09-10
If git_icons = true, really slow.


uga-rosa:
This setting alone will slow down the speed to open `FzfLua files` by more than a second.
Do you have any idea why?

NVIM v0.5.0 (stable)
OS: Ubuntu on WSL2
git version 2.25.1

ibhagwan:
In order to determine the icons for files a couple of commands are run in the background, it’s possible one of them is slow, can you run the below and let me know if they take a significant time?

```lua
    git_diff_cmd        = "git diff --name-status --relative HEAD",
    git_untracked_cmd   = "git ls-files --exclude-standard --others",
```

uga-rosa:
When run in a terminal, these commands returned values fast...

ibhagwan:
> When run in a terminal, these commands returned values fast...

Can you try running `files` with `git_icons = true` in a new repo with just a couple of files or on different repos and see if this happens on all git repos or just specific ones? I’m trying to think what else could be slowing things down other than the commands above.

uga-rosa:
There are three levels that I can feel. The fastest is when I set git_icons = false; even outside the git repo, setting git_icons = true slows it down a bit. The slowest was when I was inside the git repo and set git_icons = true. This happens even in a newly created repo with no files in it.

ibhagwan:
Perhaps the command that checks if you’re in a git repo slows it down, can you run this in the shell in the same locations you ran fzf-lua and let me know if this seems like the culprit?

```sh
git rev-parse --show-toplevel
```

ibhagwan:
There are the only 3 extra commands running when you specify `git_icons`, maybe it has something to do with being run inside nvim?

Can you try running the commands inside neovim like this:

```lua
:lua print(vim.fn.system('<command>'))
```

uga-rosa:
I feel a slight delay, but I am not sure if it is due to `print`.
I've also looked at the source, and I can't think of any reason other than the slow call to vim.fn.systemlist :(

ibhagwan:
> I feel a slight delay, but I am not sure if it is due to `print`.
> I've also looked at the source, and I can't think of any reason other than the slow call to vim.fn.systemlist :(

Yes, this is awfully strange, there really isn’t much code added for the git icons.

if you suspect `systemlist` how about we try with either the [official neovim 0.5 appimage](https://github.com/neovim/neovim/releases/download/v0.5.0/nvim.appimage) or the nightly?

[...]

PR #130, opened 2021-09-11
Improved the execution speed of external commands.

uga-rosa:
#126
I also fixed some problems that I noticed. For example, `vim.fn.systemlist()` returns an empty string when it fails, so there is no point in displaying it as an error message. So, I made it so that it displays the command that was executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,9 @@ require'fzf-lua'.setup {
   file_icon_colors = {
     ["lua"]   = "blue",
   },
+  -- If true, use io (lua standard library) for runnig shell command (ex. git)
+  -- If git-based commands are slow, it might improve things.
+  lua_io = false,
 }
 ```
 
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -1,5 +1,6 @@
 local utils = require "fzf-lua.utils"
 local path = require "fzf-lua.path"
+local stdio = require "fzf-lua.stdio"
 
 local M = {}
 
@@ -241,9 +242,9 @@ M.git_switch = function(selected, opts)
   local args = ""
   local is_remote = branch:find("^remotes/") ~= nil
   if is_remote then args = "--detach " end
-  local output = vim.fn.systemlist(cmd .. args .. branch)
-  if utils.shell_error() then
-    utils.err(unpack(output))
+  local output, err = stdio.get_stdout(cmd .. args .. branch)
+  if err then
+    utils.err(cmd .. args .. branch)
   else
     utils.info(unpack(output))
     vim.cmd("edit!")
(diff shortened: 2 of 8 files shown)


ibhagwan:
This is not yet ready to be merged, first the `get_stdout` function dose not recognize shell errors properly, I added `lua_io` as a parameter for testing and it returns different results than `systemlist`:

```lua
:lua _G.dump(require'fzf-lua.stdio'.get_stdout("echo 1", false))   -- { "1" } false  <-- OK
:lua _G.dump(require'fzf-lua.stdio'.get_stdout("echo 1", true))    -- { "1" } false  <-- OK
:lua _G.dump(require'fzf-lua.stdio'.get_stdout("badcmd", false))   -- "" true        <-- OK
:lua _G.dump(require'fzf-lua.stdio'.get_stdout("badcmd", true))    -- {} false       <-- NOT OK
```

Second, this messes up the error message coming out of `is_git_repo()`, due to the function unable to detect the shell error code it always returns `false` and the users won't see the error message `fatal: not a git repository (or any parent up to mount point /)`.

Also, this might need a rebase soon, I've decided to drop the two git commands to detect changes and just run `git status -s` instead.

uga-rosa:
Oh, I was mistaken. It's true that we can't recognize stderr this way.
I can use `vim.loop` to get stderr, but it is asynchronous, so I need to be creative.

I understand about the `git status -s`. I'll wait.

ibhagwan:
Just committed the change, test it out, even with `systemlist` it should be faster since I removed the untracked command and now using only `git status -s` to get all git indicators.

uga-rosa:
Unfortunately, the speed was not fast enough for regular use, although it was better :_(
I'll try to find out more about the cause.

ibhagwan:
> Unfortunately, the speed was not fast enough for regular use, although it was better :_(
> I'll try to find out more about the cause.

I'll also work on your `io.popen` idea, if I can make it exactly equal to `systemlist` then why not.

uga-rosa:
If use is limited, 2>&1 may be a good solution.

I'll close this PR and fork it again.
I'm a beginner, so I don't know much about git operations yet. Haha.

ibhagwan:
Just FYI, if you want to update your PR next time:

```sh
git reset --mixed HEAD^2    # undo your two commits (doesn't change files on disk)
git reset --hard HEAD       # resets the working dir to the current HEAD
                            # (will have discarded your on disk changes)

git remote add upstream https://github.com/ibhagwan/fzf-lua.git
git fetch upstream
git rebase upstream/main
...
<work on your new changes>
...
git push --force            # overwrite the PR branch effectively updating this PR
```

uga-rosa:
Thx!

PR #131, opened 2021-09-12
add filetype

uga-rosa:
If filetype exists, I can do something about it with my own settings.


ibhagwan:
I believe `nvim-fzf` will override this when opening the terminal:

https://github.com/vijaymarupudi/nvim-fzf/blob/74b62b00fb9959b288b5d1c6370c3c2d67866f8b/lua/fzf.lua#L137

uga-rosa:
I see. I'll use that one.
Thx!

PR #134, opened 2021-09-15
fix: wrong live_grep export

sandangel:


ibhagwan:
This is on purpose, A couple weeks back I decided to make the default live grep continue the searches for the users, if you wish to have the default not continue your search you can send `live_grep({ continue_last_ search = false })` (I have two live grep mappings, one with resume enabled and one without.

sandangel:
@ibhagwan that's is confusing. If the user want to have that behavior, they can use `live_grep_resume` right?

ibhagwan:
> @ibhagwan I think that is confusing. If the user want to have that behavior, they can use `live_grep_resume` right? I think the command should do what it says.

That’s true but my thinking was twofold, with the (yet) poor documentation users might not find the option and know resume search even exists, also, this way the behavior is a bit closer to a modern editor, where your search term is saved and is visible next rime you press `ctrl-f` (or any other search bind).

The downside of this isn’t really major as users can always disable it with a single parameter.

PR #142, opened 2021-09-17
Remove early return that blocks setting cwd

winston0410:
The early return is blocking the setting of `cwd`.

I guess you are trying not modify the command, if the user provide a customized one? I think this might break other behaviors.

### DIFF
--- a/lua/fzf-lua/providers/files.lua
+++ b/lua/fzf-lua/providers/files.lua
@@ -13,9 +13,6 @@ local get_files_cmd = function(opts)
   if opts.raw_cmd and #opts.raw_cmd>0 then
     return opts.raw_cmd
   end
-  if opts.cmd and #opts.cmd>0 then
-    return opts.cmd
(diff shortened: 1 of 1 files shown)


ibhagwan:
That’s because I can’t guarantee where to put the `cwd` within the command, not all of them are going to be `fd` or `find`.

I was going to move the `cwd` to the shell spawning like I do with grep, it’s a better solution than doesn’t require modifying the command, we can close this PR as I’m going to do this modification soon.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/f3f373beafe82de7bdbc26b724a31047f805a05e this is much better and was on my plate anyways, you just helped this get pushed into the code faster :-)

I apologize of robbing your PR opportunity.

winston0410:
Sure it will be fine, now with the dynamic cwd I can open files all days 🚀

PR #145, opened 2021-09-21
fix: no need to exit visual mode

sandangel:
no need to exit visual mode for `grep_visual`.
I normally use `grep` like this:

```vim
xnoremap <silent><expr> <leader>d ":\<C-U>lua require('fzf-lua').live_grep({ continue_last_search = false })\<cr>" . luaeval('require("fzf-lua.utils").get_visual_selection()')
```
so I can continue editing the grep words, no need to set to fixed value for search.

but the line for exiting visual mode will make the float window disappear. This PR will fix that

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -213,10 +213,6 @@ function M.get_visual_selection()
         -- visual line doesn't provide columns
         cscol, cecol = 0, 999
       end
-      -- exit visual mode
-      vim.api.nvim_feedkeys(
-        vim.api.nvim_replace_termcodes("<Esc>",
-          true, false, true), 'n', true)
     else
       -- otherwise, use the last known visual position
       _, csrow, cscol, _ = unpack(vim.fn.getpos("'<"))


ibhagwan:
2 issues here:
- without this the visual selection isn't always saved properly to be able to reselect with `gv`
- `grep_visual` opens in normal mode instead of insert mode

If you're using `get_visual_selection` directly and that's the only function you need why don't you just copy this function into your own version of `utils` and remove the `feedkey` sequence there?

sandangel:
@ibhagwan `live_grep({ search = '' })` will not allow you to edit the search keywords. How about adding an option to this function for exiting visual mode or not?

sandangel:
Oh, I found that `{ search = '' }` will do exactly what I want. thanks for your response. ^^

ibhagwan:
> @ibhagwan I thought `live_grep({ search = '' })` will not allow you to edit the search keywords. How about adding an option to this function for exiting visual mode or not?

Only `grep` won't allow you to edit the search keywords as it feeds the results from running `rg` once and then `fzf` takes over the search for a fuzzy search.

`live_grep` is always interactive mode, `fzf` is used only as a selector UI, each keystroke runs the `rg` command again with the new typed keyword (that's why live grep isn't "fuzzy").

As a side note, check out `skim`, it has a really cool feature where you can switch between interactive mode and fuzzy mode, so once you're happy with the rg command result you can press `ctrl-q` and fuzzy the result set (basically having both `grep` and `live_grep` in the same interface.

PR #182, opened 2021-10-11
Add pager option and pipe cmd to pager if exists

bartoszmaka:
Hi, so firstly I'd like to thank you for creating this amazing plugin. It is something that I was looking for for a while 😄

I've been messing around and found a way to pass a pager to the command, so for example it is possible now to set options like this
``` lua
local actions = require "fzf-lua.actions"
require'fzf-lua'.setup {
    git_diff = {
      cmd             = 'git diff',
      args            = "--color",
      pager           = "delta"
    },
 -- rest of the config
}
```
to get this result in the preview window.

image

What do you think about merging this into the main repo?

### DIFF
--- a/lua/fzf-lua/previewer/fzf.lua
+++ b/lua/fzf-lua/previewer/fzf.lua
@@ -178,7 +178,7 @@ function Previewer.git_diff:cmdline(o)
     local cmd = self.cmd
     local args = self.args
     local pager = nil
-    if self.pager and vim.fn.executable(self.pager) then
+    if self.pager and vim.fn.executable(self.pager) == 1 then
       pager = '| ' .. self.pager
     end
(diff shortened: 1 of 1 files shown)


bartoszmaka:
I've added an additional check whether pager is executable.

Also someone reading this might not know which `delta` i mean in the config, so it's this one 😄  https://github.com/dandavison/delta

ibhagwan:
FYI, it’s now possible to configure pager in n easier way (I.e. you don’t need to duplicate the preview command or use `previewers.git_diff.pager`), all you need to do is define `preview_pager = 'delta -w $FZF_PREVIEW_COLUMS'` and place it under `git.status`, `git.commits` or `git.bcommits`.

PR #183, opened 2021-10-11
fix(): update default preview config to `'border'` from `'float'`.

megalithic:
Was throwing the following:
`[Fzf-lua] fzf error 2: invalid preview window option: float`

Now matches the docs.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -43,7 +43,7 @@ M.globals = {
     },
     preview = {
       default             = "builtin",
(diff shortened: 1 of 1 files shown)


megalithic:
> Ty for fixing it! I’m not near the comp and couldn’t fix it rn :)
> 
> On another note, did you try the new 'float' scrollbar? If you did, which one do you prefer?

to be honest, i never used the scrollbar before in the preview window. the addition of the float option slowed down preview rendering big time for me, soooo, i turned it off.

Issue #204, opened 2021-11-04
Cannot preview sometimes when using `live_grep()` and `files()`


nyngwang:
As title. The result can be shown correctly in the input-area, but it cannot be previewed.

## My settings

```
      require'fzf-lua'.setup {
        winopts = { -- Only valid when using a float window
          border = { '╭', '─', '╮', '│', '╯', '─', '╰', '│' },
          height = 0.65,
          width = 0.7,
          row = 0.9,
          col = 200,
          preview = {
            default = 'cat',
            title = false,
            wrap = 'nowrap',
            vertical = 'up:80%',
            layout = 'vertical'
          }
        },
        files = {
          cmd = 'fd -H',
          prompt = ' File $ '
        },
        grep = {
          prompt = ' Word $ ',
          cmd = "rg --vimgrep",
          color_icons = false,
          file_icons = false,
        }
```

## Result

<img src="https://user-images.githubusercontent.com/24765272/140395404-31efa43e-5dc1-4671-9757-fd38a1bc73e8.png" width="600">

<img src="https://user-images.githubusercontent.com/24765272/140395589-2483415c-8354-46c5-b226-1ef4dae3ef3f.png" width="600">

ibhagwan:
I responded to this on #203, it's probably due to differences between POSIX `cat` and GNU `cat`, you can define the arguments by adding this section to your `setup()`:

```lua
require'fzf-lua'.setup {
  previewers = {
    cat = {
      cmd = "cat",
      args = "",
    },
  }
}
```

nyngwang:
> > > Can you try with `winopts.preview.default = 'bat_native'`?
> > 
> > 
> > This one works! Wow!
> 
> Great! I'd still like to solve it for `bat`/`cat`, does this work at all for different contents? can you tell what file contents consistently fails? Maybe we should discuss this under #204?

^Context

nyngwang:
> [...] does this work at all for different contents? [...]

Yes, only that line with URL will fail.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/454b0ba48c89f05ca16023cb34b1ae21c30ff8af fixed now.

I introduced this bug by mistake when fixing #195, my regex was wrong and matched URLs even if they weren't at the start of the line :-)

nyngwang:
Now the same happens when `head` as previewer:

[...]

PR #205, opened 2021-11-05
fix: too broad uri regex match - restrict to jdt / file

kraem:
found this bug independently of #204 / https://github.com/ibhagwan/fzf-lua/commit/454b0ba48c89f05ca16023cb34b1ae21c30ff8af and wrote this fix.

unfortunately it still occured with https://github.com/ibhagwan/fzf-lua/commit/454b0ba48c89f05ca16023cb34b1ae21c30ff8af  (like grepping and finding a result `[src](https://domain.tld)`).

although it may not be as elegant i figured i'd create a pr with my initial fix.

thanks for the well documented code btw :)

### DIFF
--- a/lua/fzf-lua/path.lua
+++ b/lua/fzf-lua/path.lua
@@ -146,7 +146,7 @@ function M.entry_to_file(entry, cwd)
   -- entries from 'buffers' contain '[<bufnr>]'
   -- buffer placeholder always comes before the nbsp
   local bufnr = idx>1 and entry:sub(1, idx):match("%[(%d+)") or nil
-  if not bufnr and stripped:match("^[^ ]+://") then
+  if not bufnr and stripped:match("^%a+://") then
     -- Issue #195, when using nvim-jdtls
     -- https://github.com/mfussenegger/nvim-jdtls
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for submitting this :)

However, I wished for the solution to be more generic and not necessarily limited to `jdt://` or `file://` (in case some other LSP plugin uses a similar method).

> unfortunately it still occured with 454b0ba (like grepping and finding a result [src](https://domain.tld)).

How do you get grep to produce output such as `[src](https://domain.tld)`? Even if this is contained in the line there should be a space before the contents (after the line/col) so the regex should still work properly.

If we still opt to narrow the regex result I would do this instead:
```lua
-- '%a’ matches 'a-z,A-Z'
stripped:match("^%a+://")
```

kraem:
thanks, that's a better solution :)

i don't get a space after line/col:
```
test:1:1:[src](https://domain.tld)
```

ibhagwan:
> thanks, that's a better solution :)
> 
> i don't get a space after line/col:
> 
> ```
> test:1:1:[src](https://domain.tld)
> ```

Merged, ty :)

Issue #248, opened 2021-12-12
Testers needed for `multiprocess = true`


ibhagwan:
Background:

https://github.com/ibhagwan/fzf-lua/commit/2d8a4e9afc7b32cdd7552f1ccd641c8bd6c2e85c  - This commit adds a new option named `multiprocess`, best described from the commit message:
```help
major performance improvement: process entries externally, READ BELOW:
Since LUA is single threaded I reached a limit to performance
optimization, both 'git_icons' and 'file_icons' require string
matching and manipulations which eventually hurt performance
when running on large amount of files.
In order to solve that this commit introduces the option to spawn
commands and process the entries in a separate neovim process which
prints to stdio as if it was a regular shell command. This speeds up
things significantly and also makes the UI super responsive as if fzf
was run in the shell. This required a few lua hacks to be able to load
nvim-web-devicons in a '--headless --clean' instance and sharing the
user configuration through the RPC interface from the running instance.
This is enabled by default for 'files' and 'grep' providers and can also
be enabled for 'git.files' if required, control using the 'multiprocess'
option.
```

Turns out I was a bit ballsy and there are still issues with it (e.g. #247).

Before I re-enable it again as default I'd like to have some testers use this (perferably on large codebases) and let me know the results.

It's only possible to enable it for `files`, `grep` and `git_files`, either through setup:
```lua
require'fzf-lua'.setup {
  files = {
    multiprocess      = true,
    debug             = false,
  },
  grep = {
    multiprocess      = true,
    debug             = false,
  },
  git = {
    files             = {
      multiprocess    = true,
      debug           = false,
    },
  }
}
```

Or by specifying it directly:
```lua
:lua require'fzf-lua'.files({ multiprocess = true })
-- or
:FzfLua files multiprocess=true
```

If `debug=true` is specified the neovim shell command will be printed into the `:messages` stream so you can run the command directly in the shell and see if everything seems right, the command wraps a shell command and prints the output to the shell after additonal processing, for example:
```sh
[...]

PR #252, opened 2021-12-13
Fix multiprocess mode pipes to use io.write() 

### LINKED ISSUE #251: new default setting for `multiprocess=true` breaks files and gitfiles commands
with no changes and updating the plugin this morning (did not set multiprocessing), the files and git files command stopped working properly.

When you trigger them, the split opens and you see options but the display is all messed up and its glitched out in the center of the new buffer. adding `multiprocess=false` to `files` config fixes the issue.

Here is a screenshot of the glitch:
Screen Shot 2021-12-12 at 12 14 22

This issue exists on both the default macOS terminal, and kitty terminal and also both on Neovim HEAD and Stable.

jdrouhard:
Instead of opening a new file descriptor on /dev/stdout and /dev/stderr, simply use builtin io module.

MacOS in particular doesn't seem to like processes opening /dev/stdout and /dev/stderr. Or at the very least there are some hidden nuances there that aren't actually necessary since Lua's builtin io module can already take care of writing to the pre-opened stdout and stderr file descriptors.

Fixes #251 and possibly #248

### DIFF
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -332,12 +332,12 @@ M.spawn_stdio = function(opts, fn_transform, fn_preprocess)
       end)
   end
 
-  stderr = pipe_open(opts.stderr or "/dev/stderr")
-
-  -- /dev/stdout seems to create issues on MacOS (#248, #251)
-  -- do not use a pipe unless use specifically requests it
-  -- instead we use 'io.write'
-  stdout = pipe_open(opts.stdout)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tysm @jdrouhard for once again being amazing!

Minor changes: I removed the `cb(nil)` part from the error callback as it wasn't needed and also changed the error message in `exit` to use `io.stderr:write` instead of `io.write`.

I think this should be good now!

Issue #263, opened 2021-12-16
`live_grep_glob`: lua/fzf-lua/utils.lua:83: E5560: nvim_exec must not be called in a lua loop callback


nyngwang:
As title. The error message:

```
Error executing luv callback:
...vim/site/pack/packer/start/fzf-lua/lua/fzf-lua/utils.lua:83: E5560: nvim_exec must not be called in a lua loop callback
stack traceback:
        [C]: in function 'cmd'
        ...vim/site/pack/packer/start/fzf-lua/lua/fzf-lua/utils.lua:83: in function 'warn'
        ...vim/site/pack/packer/start/fzf-lua/lua/fzf-lua/utils.lua:158: in function <...vim/site/pack/packer/start/fzf-lua/lua/fzf-lua/utils.lua:156>
Press ENTER or type command to continue
```

My settings:

```
rg_opts =
  "--column "..
  "--line-number "..
  "--no-heading "..
  "--case-sensitive "..
  "--hidden "..
  -- Disable: `-u` for .gitignore, `-uu` for hidden files/Dirs, and `-uuu` for binary files.
  "--unrestricted "..
  " "..
  "--max-columns 200 "..
  "--max-filesize 200K "..
  "--max-depth 10 "..
  " "..
  "--glob '!.npm' "..
  "--glob '!.nvm' "..
  "--glob '!.tmux' "..
  "--glob '!.cache' "..
  "--glob '!.vscode' "..
  "--glob '!.local/share' "..
  "--glob '!.config/nvim/plugin' "..
  "--glob '!.Trash' "..
  "--glob '!Library' "..
  "--glob '!Documents' "..
  "--glob '!Pictures' "..
  " "..
  "--glob '!.git' "..
  "--glob '!node_modules' "..
  "--glob '!package-lock.json' "..
  "--glob '!yarn.lock' "..
  "--glob '!tags' "..
  " ",
```

It's strange that I did add the option `--glob '!.Trash'` while the error seems to say that it's trying to access something in the `.Trash/`. As img shown:

<img src="https://user-images.githubusercontent.com/24765272/146364231-dcdf3715-ae43-443a-92a2-e172ab712292.png" width="500">

What I think relevant:

```
-g, --glob <GLOB>...
        Include or exclude files and directories for searching that match the given
        glob. This always overrides any other ignore logic. Multiple glob flags may be
        used. Globbing rules match .gitignore globs. Precede a glob with a ! to exclude
[...]

PR #265, opened 2021-12-16
fix: glob precedence in #263.

nyngwang:


nyngwang:
This doesn't work. The problem is that `*` should not be inserted by the user itself.

Issue #287, opened 2021-12-31
The results of `FzfLua files` overflow on Ubuntu 20.04 with Fzf version `0.29.0` and `0.24.3`


nyngwang:
## Fzf Version

```
✔ parallels ~ % fzf --version
0.29.0 (dc975e8)
```

## The problem:

As shown in the below demo. I remember that I have seen this bug before, and did search for the issue while I didn't find it.

The command I run is

```
lua require("fzf-lua").files()
```

https://user-images.githubusercontent.com/24765272/147813592-947aa98f-bbb7-4d75-a48b-2b2d3c5937d2.mov

nyngwang:
OK from the last second, i.e. 0:08, it seems like it's because there is a path containing white spaces:

```
.../Parallels Shared Folders/Home/...
```

ibhagwan:
>OK from the last second, i.e. 0:08, it seems like it's because there is a path containing white spaces

Just tested and I don't think this is related to spaces, you're running it on your home directory and it seems that `fd` is failing to access certain MacOS special folders and what you're seeing here is the error message returned by `fd`.

I really can't tell why `fd` fails but I can try to supress the error message as a workaround, regardless, it should not overflow into the `fzf` search field, we have to see what the maintainer of `fzf` says about this issue https://github.com/junegunn/fzf/issues/2697, which will potentially also solve #275.

ibhagwan:
I think I know what causes this, as I suspected when certain folders fail access `mutiprocess = true` will print the error to `stderr` which is then displayed by `fzf`, in your case the inaccessible folders generate an error message which then causes an overflow with `fzf`.

The correct behavior should be to display the problematic folders and their corresponding error message at the top but this seems to bug fzf in your case (which may or may not be related to https://github.com/junegunn/fzf/issues/2697):

![screenshot-1641064546](https://user-images.githubusercontent.com/59988195/147858376-7867a18d-4550-4206-b6d1-961208b8da1a.png)

Can you test this with `skim`? If this works correctly with `sk` we will confirm this as an `fzf` bug.

ibhagwan:
Solved it in #316, MacOS doesn't seem to like `stderr`, https://github.com/ibhagwan/fzf-lua/commit/e89d5bc009be8a631d5708701722fc30232879ee solves it.

nyngwang:
Thank you!!!

PR #307, opened 2022-01-16
fix devicon size in kitty using file_icon_padding

JSteitz:
This fix removes the nbsp. Kitty needs a normal space to extend the icon size.

### DIFF
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -147,7 +147,7 @@ M.get_devicon = function(file, ext)
 
   if config.globals.file_icon_padding and
     #config.globals.file_icon_padding>0 then
-    icon = icon .. config.globals.file_icon_padding:gsub(" ", utils.nbsp)
+    icon = icon .. config.globals.file_icon_padding
   end
 
   return icon, hl


ibhagwan:
> This fix removes the nbsp. Kitty needs a normal space to extend the icon size.

I see, but we need a different solution here, the default fzf separator for entires is space, I can’t recall right now what might be affected but there is a good chance some entry string manipulations rely on the fact that there is no space in the fzf-lua prefixes (icons, buffer number, etc) and stripping the string up-to the last `nbsp`.

JSteitz:
Don't know if this has something to do with the problem you are speaking and if it also affects other terminals, but the Icons have some issues if the multiprocess is enabled.
* different sizes
* different paddings
* overriden icons, do not show up

ibhagwan:
Multiprocess is a relatively new feature meant to decouple the entry processing from neovim in a different process, if you can reproduce the error consistently we can identify and fix the issue.

If you send `debug = true` as an option you’ll be able to see the underlying command that is then run by the shell which you can run in an external terminal which will “iconify” the results.

Since you mention 3k+ files this might be related to inaccessible folders and the error messages overflowing into the display? See #287, in any event it’s better to have a separate issue to deal with the issue than discussing it in the context of the PR.

JSteitz:
> > This fix removes the nbsp. Kitty needs a normal space to extend the icon size.
> 
> I see, but we need a different solution here, the default fzf separator for entries is space, I can’t recall right now what might be affected but there is a good chance some entry string manipulations rely on the fact that there is no space in the fzf-lua prefixes (icons, buffer number, etc) and stripping the string up-to the last `nbsp`.
```lua
from utils.lua
-- nbsp, U+00a0: the original separator, fails with files that contain nbsp
-- nbsp + zero-width space (U+200b): works only with `sk` (`fzf` shows <200b>)
-- word joiner (U+2060): display works fine, messes up fuzzy search highlights
-- line separator (U+2028), paragraph separator (U+2029): created extra space
-- EN space (U+2002): seems to work well
```

Anyway I have good news for this PR.
We can drop it, it seems kitty can also handle nbsp for icons.
With some testing, I found out, that it is the `multiprocess` that makes the problems.

I'll create a new ticket for that.

JSteitz:
My bad, I had it differently working. With a potential bug.

With this code, I make it work.

```lua
require('nvim-web-devicons').setup({ default = true })

local user_icons = {}

for ext, value in pairs(require('nvim-web-devicons').get_icons()) do
  user_icons[ext] = vim.tbl_extend('force', value, {
    icon = value.icon .. ' '
  })
end

require('nvim-web-devicons').set_icon(user_icons)
require('nvim-web-devicons').set_default_icon(' ', '#6d8086', '66');
```

JSteitz:
swappy-20220116_144737
Just tested it with files that contain spaces and I didn't have any issues.
Is it still relevant?

ibhagwan:
>Just tested it with files that contain spaces and I didn't have any issues.
Is it still relevant?

Most of the string manipulations rely on `:` as a separator rather than space, I guess we can have this merged and if there are issues we deal with it later.

PR #318, opened 2022-01-20
Set default prompt for ui_select to match the official doc

leiserfg:


leiserfg:
The doc says:

> prompt (string|nil) Text of the prompt.
>                                     Defaults to `Select one of:`

ibhagwan:
Ty @leiserfg, looks great!

Do you know where I can test this? I tried with spell fix (`z=`) but it seems to not yet work with `vim.ui.select` - couldn’t find anything else other than code actions that uses the new menu.

leiserfg:
I saw that https://github.com/mickael-menu/zk-nvim is using it.

leiserfg:
https://github.com/Shatur/neovim-session-manager is also using it, maybe having a look at https://github.com/stevearc/dressing.nvim is worth it.

ibhagwan:
Merged, yysm @leiserfg for your conribution :-)

PR #320, opened 2022-01-20
1546c82 doesn't work in fish

oncomouse:
When using fish shell, the fzf previewer throws an error because it uses bash syntax to set variables. Replacing with a call to "env" should fix the issue.


ibhagwan:
Actually since it's so quick I just implemented it, seems to work fine, let me know if you're still having issues and we can reopen this.

PR #358, opened 2022-03-08
Correct vim-plug install instructions

callumacrae:
Previously defaulted to master branch, this tells it to use main instead


ibhagwan:
Ty! :-)

PR #394, opened 2022-04-23
Erroneous `return`

toppair:
I think this method should not return here? my `on_create` function doesn't run after switch from grep to live_grep via action if it does.

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -712,7 +712,6 @@ function FzfWin:create()
     -- fzf will not use all the avialable width until 'redraw' is
     -- called resulting in misaligned native and builtin previews
     vim.cmd("redraw")
-    return
   end
 
   if not self.winopts.split and self.previewer_is_builtin then


ibhagwan:
I’m not sure what this would lead to, `on_create` doesn’t get called here because the window doesn’t get created here, I believe this code path is taken only when reusing the current window.

Can you expand on what isn’t working for you due to `on_create` not being called? Maybe there is a better solution?

toppair:
I set this mapping inside `on_create` function to close fzf window:

`vim.api.nvim_buf_set_keymap(0, 't', 'jj', '<C-c>', {})`

And since buffer is replaced inside this check (if I'm not mistaken) and `on_create` doesn't run, the mapping no longer works and I can't set a new one.

toppair:
Just assumed `return` was put there by mistake, haven't tested it too much. Works for my setup. Maybe should have opened an issue instead.

ibhagwan:
> Just assumed `return` was put there by mistake, haven't tested it too much. Works for my setup. Maybe should have opened an issue instead.

It might indeed be harmless but I'm not sure since it overrides the calling buffer/win (`src_bufnr` and `src_winid`), also has unintended consequences when `fzf-lua` is used in a split (as opposed to a float win).

Fzf-lua win redraw re-uses the existing window but has to generate a new term buffer, since keymaps are attached to the buffer those go away on window reuse, the below commit makes sure to recall `on_create` right after remapping the default keybinds, verified working with your custom keymap.

https://github.com/ibhagwan/fzf-lua/commit/19f79c26bda281173d04e6e5a171793d56a52d44

PR #401, opened 2022-04-30
ref(lsp): add server_capabilities support on 0.8

arsham:
In Neovim `v0.8` the `resolved_capabilities` is deprecated and is replaced by the `server_capabilities`. The `capability` property of the handlers table is split into `resolved_capability` and `server_capability`, and in the `wrap_module_fncs` is set back based on the neovim's version.

Please let me know if this is not the right way to approach this problem.


ibhagwan:
Wonderful job @arsham! <3

PR #430, opened 2022-06-05
fix(jumps): fix preview not showing correctly

Shivix:
Fixes two issues:
* Preview wasn't showing the jump location when the file was outside of
  the pwd.
* Preview wasn't showing the jump location when the jump was on a blank
  line in the current buffer.


ibhagwan:
Ty @Shivix!

PR #460, opened 2022-06-24
fix(quickfix): prevent indexing opts when nil

Shivix:
`opts` was being used before it was normalized causing the quickfix
command to fail when used without any options.


ibhagwan:
Makes sense ty @Shivix!

PR #492, opened 2022-08-03
feat: do not override user-specified cwd for git files

gpanders:
Use the `cwd` specified by the user if it exists; otherwise, fallback to the git repo root.

### DIFF
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -9,7 +9,9 @@ local make_entry = require "fzf-lua.make_entry"
 local M = {}
 
 local function set_git_cwd_args(opts)
-  opts.cwd = path.git_root(opts)
+  if not opts.cwd then
+    opts.cwd = path.git_root(opts)
+  end
   if opts.git_dir or opts.git_worktree then
     opts.cmd = path.git_cwd(opts.cmd, opts)
   end


ibhagwan:
Ty @gpanders for the pull request!

This should already possible using the `git_dir` call argument, you can even specify a `git_worktree` override  if you wish, may I ask what are you trying to achieve?

ibhagwan:
I understand what you’re after now, doesn’t `:FzfLua git_files` do somewhat the same thing?

If you’re still after traversing the git root specifically with `fd` (instead of `git ls-files`), what if instead of adding a new option (there are already way too many) we use this:

```lua
:lua require'fzf-lua'.files({ cwd = require'fzf-lua'.path.git_root({ cwd = vim.loop.cwd() }) })
```

gpanders:
I want to be able to set `use_git_root` to false, meaning if I am in a subdirectory of a git repo I want `:FzfLua git_files` to only list files beneath my current directory (i.e. what `git ls-files` shows). At present, fzf-lua hardcodes the git repo root as the cwd, so `require("fzf-lua").git_files({cwd = vim.loop.cwd()})` does not work.

ibhagwan:
Then why don’t we make this `require("fzf-lua").git_files({cwd = vim.loop.cwd()})` work or use the already existing `cwd_only` option (used in a few providers, lsp_symbols, diags, etc) option work with `git_files` instead of a new option?

gpanders:
Good idea!

ibhagwan:
Your last commit changes the default behavior for most users which lists the entire git repo regardless of your cwd inside the repo, I have a different idea how to implement this, hang on :-)

ibhagwan:
Actually I got it wrong, it didn't change the default behavior but it had one issue of showing an empty interface when using `cwd` which isn't a git repo instead of failing with an error, I manually rebase-merged and added said change in https://github.com/ibhagwan/fzf-lua/commit/830b6857965e8f98e4d491ee31f750f67905625e, you can now use:
```lua
:lua require'fzf-lua'.git_files({ cwd = vim.loop.cwd() })
```

Ty @gpanders!

PR #502, opened 2022-08-18
put diagnostic.data on get_line_diagnostics

tsukimizake:
Hi. I found code_actions of elm-language-server doesn't work with `:FzfLua lsp_code_actions`, while `:lua vim.lsp.buf.code_action()` works.

I investigated the code to find `data` argument was not passed on `get_line_diagnostics` function, and the problem goes away by passing it.

According to [lsp specification about Diagnostic](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic), seems like it should be passed.

Thanks.

### DIFF
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -590,7 +590,9 @@ local function get_line_diagnostics(_)
         line = diag[1].end_lnum,
         character = diag[1].end_col,
       }
-    }
+    },
+    data = diag[1].user_data and diag[1].user_data.lsp and
+      diag[1].user_data.lsp.data
   }} or nil
 end
 


ibhagwan:
Tysm @tsukimizake for the PR!

Technically the LSP query is not required, the only reason it's there is to verify there are code actions before running `vim.lsp.buf.code_action` in
https://github.com/ibhagwan/fzf-lua/blob/054c312b1cc327be68045ffc4926133a60ff9b82/lua/fzf-lua/providers/lsp.lua#L651

The reason I have to verify there are code actions is also esoteric, since the introduction of `vim.ui.select`, the way code actions work is by changing the `vim.ui.select` interface using `:FzfLua register_ui_select` and then calling `vim.lsp.buf.code_action`, after the window is closed `:FzfLua deregister_ui_select` is called  via the `post_action_cb`:
https://github.com/ibhagwan/fzf-lua/blob/054c312b1cc327be68045ffc4926133a60ff9b82/lua/fzf-lua/providers/lsp.lua#L647-L649

When no code actions exist it will register fzf-lua for `vim.ui.select` but the deregistration will not get called as the fzf-lua won't open.

The code is also irrelevant if the user ran `register_ui_select` prior (i.e. during setup) since we don't need to call deregister.

Perhaps I need to explore again how to deregister the UI select when no code actions exist so I can retire this part of the code.

PR #504, opened 2022-08-21
nicer UI for keymaps provider inspired by telescope

sp4ke:
A less cluttered and more minimalist UI for the *keymaps* provider inspired by telescope.

2022-08-21-235151_3412x1358_scrot

### DIFF
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -268,21 +268,27 @@ M.keymaps = function(opts)
   opts = config.normalize_opts(opts, config.globals.nvim.keymaps)
   if not opts then return end
 
-  local modes = { "n", "i", "c" }
+  local modes = {
+    n = "blue",
+    i = "red",
+    c = "yellow"
+  }
   local keymaps = {}
 
   local add_keymap = function(keymap)
     -- hijack fields
-    keymap.str = string.format("[%s:%s:%s]",
-      utils.ansi_codes.yellow(tostring(keymap.buffer)),
-      utils.ansi_codes.green(keymap.mode),
-      utils.ansi_codes.magenta(keymap.lhs:gsub("%s", "<Space>")))
+    local keymap_desc = keymap.desc == nil and keymap.rhs or keymap.desc
+    keymap.str = string.format("%s │ %-40s │ %s",
+      utils.ansi_codes[modes[keymap.mode] or "blue"](keymap.mode),
+      keymap.lhs:gsub("%s", "<Space>"),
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tysm for the PR @sp4ke! this is much better indeed.

I never really liked the keymaps screen but back then I wanted to maintain feature parity with telescope and if I recall correctly, this is how it used to look back then :-)

A couple of issues (I don't even know where this binds come from in my config):
- Empty binds display `nil` instead of empty string
- Unicode chars seem to mess up the alignment:

> Fzf-lua with this fork:
image


> Telescope:
image

sp4ke:
I added a fix to the empty description fields. 

However I have no idea how to manage the unicode problem as I don't use unicode fonts myself. Looking at lua's string formating docs I don't see anything that handles string format with multi-bytes characters.

sp4ke:
The empty binds should be easy to handle however quickly looking into the unicode character issue it will be annoying. Any tip for a string.format equivalent that can handle unicode ?

ibhagwan:
I found out these mappings come from [which-key.nvim](https://github.com/folke/which-key.nvim):

https://github.com/folke/which-key.nvim/blob/bd4411a2ed4dd8bb69c125e339d837028a6eea71/lua/which-key/keys.lua#L6

and decided to ignore `<Nop>` mappings for now, can probably be dealt with by searching for a multibyte occurance in the string and adjusting the format length accordingly but not really worth the effort, empty mappings are meaningless anyways.

Your last commit actually made this not work as the code tries to concatnate an nil with an empty string, I fixed that and added the `<Nop>` ignore, rebased and merged manually, tysm for your contribution!

sp4ke:
> Your last commit broke the provider as the code tries to concatnate an nil with an empty string, I fixed that and added the <Nop> ignore, rebased and merged manually, tysm for your contribution!

You're welcome !  ty for this plugin,  I used fzf.vim before and this one perfectly replaces it with the advantage of being hackable. 

Yeah for the concat I was afraid of the nil scenario but since it worked for me I thought lua handles it.

ibhagwan:
> You're welcome ! ty for this plugin, I used fzf.vim before and this one perfectly replaces it with the advantage of being hackable.

Ty @sp4ke <3

> Yeah for the concat I was afraid of the nil scenario but since it worked for me I thought lua handles it.

The “correct” way to do this is lua is to use `or`:

https://github.com/ibhagwan/fzf-lua/blob/84142bd2240e88c2c4bdb30f3d2ab07047a1a930/lua/fzf-lua/providers/nvim.lua#L285

PR #506, opened 2022-08-23
Allow moving from dirty buffers if they are open in other windows

idanarye:


ibhagwan:
Ty for the PR @idanarye!

PR #521, opened 2022-09-18
add rockspec

teto:
to allow an install via luarocks


ibhagwan:
Ty @teto! I'm not sure how this works but seems rather straight forward.

teto:
You can see some of the reasoning behind it at https://teto.github.io/posts/2022-06-22-neovim-plugin-luarocks-2.html . 
I've used https://github.com/teto/gh2rockspec to generate the rockspec. This will be used to package fzf-lua in nixpkgs and run the tests.

PR #522, opened 2022-09-20
fix: lazy loading of fzf_wrap

stevearc:
Got a report here about fzf-lua: https://github.com/stevearc/dressing.nvim/issues/64

The lazy loading was forwarding the function arguments, but not the return value. For `find_files` this is fine, but for `fzf_wrap` we expect it to return a function.


ibhagwan:
You’re correct, good catch!

PR #528, opened 2022-09-22
binary name is fdfind on debian and ubuntu based systems

Hitesh-Aggarwal:


ibhagwan:
According to: https://github.com/sharkdp/fd#on-ubuntu this won’t solve the problem as there is already a binary named `fd` therefore it won’t reach the new condition and use the existing `fd` binary.

Also, when you submit a PR it is better to not mix different issues, this commit https://github.com/ibhagwan/fzf-lua/pull/528/commits/7a548d09c12441ae600a01deea72c3c34c33cf14 is unrelated to this issue. If you believe there is a space missing (I’m not sure) then open a new PR/issue so this can be discussed.

Hitesh-Aggarwal:
Yeah, there is a binary called fd on ubuntu (provided by fdclone package). If that is installed, fzf_lua will pick use a completely wrong command (as fd is a file manager on ubuntu systems). But since majority of users don't have it installed, it works just fine. I have tested it on my machine (running linux mint). Another solution (probably better) is to check for 'fdfind' first and then 'fd'.

Talking about commit 7a548d0, I made this commit after create the PR.  Sorry about that, I didn't knew github worked that way.

ibhagwan:
Rebased and manually marged (removed the `ui.select` prompt change) - https://github.com/ibhagwan/fzf-lua/commit/316879662ff9c33ae4b1072f8238e4bba2999e0a

PR #540, opened 2022-10-13
add default action to paste register

gennaro-tedesco:
I would like to propose an action for the `register` provider: namely insert selected text at cursor position. Not sure if in scope, but I have had it in my config for a while and I thought it may make a nice default addition: browsing registers means the user more likely than not intends to paste some content.

https://user-images.githubusercontent.com/15387611/195696850-854eda14-06bf-458b-afe5-cce26b9e3e0e.mov

In the demo you can see that it works also for long text containing spaces and tabs, pasting as intended. ~I have not figured out yet how to make new line characters `\n` paste as such, I suspect it is because you parse the registers content literally as `\n` string (thus vim thinks this is just a character rather than a line break)?~

EDIT: I amended and simplified the logic by just fetching the register number and invoking the paste command in normal mode: it works under all circumstances now.

In my configuration I have it bound as
```
...
registers = {
  actions = {
    ["default"] = actions.paste_register,
  },
}
```
I haven't found the place in the code to actually bind it to defaults - if you consider it valid you could make the change.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -692,5 +692,11 @@ M.tmux_buf_set_reg = function(selected, opts)
     end
 end
 
+M.paste_register = function(selected)
+	local lines = vim.split(selected[1], "^%[.] ", false)
+	local text = table.remove(lines)
+	vim.api.nvim_put({text}, "c", true, true)
+end
+
 return M
 
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -693,10 +693,12 @@ M.tmux_buf_set_reg = function(selected, opts)
 end
 
 M.paste_register = function(selected)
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @gennaro-tedesco for the PR!

I manually rebase/merged this as I wanted to reduce this to one commit and also improve on the logic a bit and use neovim API instead of `vim.cmd`, I also made this the default action for registers and added it the action help when pressing `F1`:
```lua
M.paste_register = function(selected)
  local reg = selected[1]:match("%[(.-)%]")
  local ok, data = pcall(vim.fn.getreg, reg)
  if ok and #data>0 then
    vim.api.nvim_paste(data, false, -1)
  end
end
```

https://github.com/ibhagwan/fzf-lua/commit/025d8c5ef8495aef838ac8e6be47f8da1bd9cc2c

PR #549, opened 2022-10-22
fix(fn_selected): pass down opts to fn_selected in fzf_wrap

gjeusel:
Hey there, love your lua glue to fzf !

Would it be possible to pass down the `opts` to `fn_selected` in fzf_wrap so we can better customize it ?

My use case is to operate multi select on rg grep sending it to the quickfix list like so:

```lua
local function grep_sel_to_qf(selected, opts)
  local qf_list = {}
  for i = 1, #selected do
    local file = fzf_lua.path.entry_to_file(selected[i], opts)
    local text = selected[i]:match(":%d+:%d?%d?%d?%d?:?(.*)$")
    table.insert(qf_list, {
      filename = file.bufname or file.path,
      lnum = file.line,
      col = file.col,
      text = text,
    })
  end
  vim.fn.setqflist(qf_list, "r")
end

local function fn_selected_multi(selected, opts)
  if not selected then
    return
  end
  -- first element of "selected" is the keybind pressed
  if #selected <= 2 then
    fzf_lua.actions.act(opts.actions, selected, opts)
  else -- here we multi-selected
    local _, entries = fzf_lua.actions.normalize_selected(opts.actions, selected)
    grep_sel_to_qf(entries, opts)
    vim.cmd("cfirst")
  end
end

-- using the following fzf keymap:
--      ["ctrl-r"] = "select-all+accept", -- https://github.com/ibhagwan/fzf-lua/issues/324
```

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -101,7 +101,7 @@ M.fzf_wrap = function(opts, contents, fn_selected)
     opts.fn_selected = opts.fn_selected or fn_selected
     local selected = M.fzf(contents, opts)
     if opts.fn_selected then
-      opts.fn_selected(selected)
+      opts.fn_selected(selected, opts)
     end
   end)
 end


ibhagwan:
It’s possible, the question is why is it needed? I’m wondering why do you need to customize `fzf_wrap` rather than achieve the same goal by writing a custom `default` action?

Also, I’m not sure what your customization does differently then the default action `file_edit_or_qf`? It seems to try to achieve the same, if a single item is selected it is acted upon and if multiple items are selected they are sent to the qf - what am I missing?

gjeusel:
Please, do correct me if there is a simpler way to achieve my goal !

#### FZF multi select issue

But from my understanding, it's a bit of a pain to operate a multi-select using fzf.
The keybinding needs to send a `select-all+accept` to fzf.
And if there is any default action on that keybinding, it takes precedence over any input fed to fzf. Which will in turn only feed a simple `select+accept` to fzf, and send the single result to the action.

```lua
local fzf_lua = require("fzf-lua")

fzf_lua.setup({
  keymap = {
    fzf = {
      ["ctrl-r"] = "select-all+accept",
    },
  },
  grep = {
    actions = {
      ["ctrl-r"] = fzf_lua.actions.file_edit_or_qf,
    },
  },
})

fzf_lua.grep({ cwd = vim.loop.cwd(), search = "" })  -- once search refined for a a few files, press 'ctrl-r' and check if multi input was sent to the action.
```

#### Custom qflist action

I'm not using the builtin `actions.file_sel_to_qf` as:
- I want to specify an action to `vim.fn.setqflist` ('r' to replace, and not the default 'a' to append)
- I don't want to execute a `copen` at the end of it

ibhagwan:
> But from my understanding, it's a bit of a pain to operate a multi-select using fzf.
The keybinding needs to send a `select-all+accept` to fzf.

Unfortunately you can’t avoid this step even with your solution, there is no other way to get all items from fzf without sending `select-all`, so given that you still need to configure the fzf keybind why not use the same code you came up with as the action to `ctrl-r` (the signature of custom actions already contains `opts`):

```lua
fzf_lua.setup({
  keymap = {
    fzf = {
      ["ctrl-r"] = "select-all+accept",
    },
  },
  grep = {
    actions = {
      ["ctrl-r"] = function(selected, opts)
        local qf_list = {}
        for i = 1, #selected do
          local file = fzf_lua.path.entry_to_file(selected[i], opts)
          local text = selected[i]:match(":%d+:%d?%d?%d?%d?:?(.*)$")
          table.insert(qf_list, {
           filename = file.bufname or file.path,
           lnum = file.line,
           col = file.col,
           text = text,
          })
        end
        vim.fn.setqflist(qf_list, "r")
      end
    },
  },
})
```

gjeusel:
Precisely because when using the same keymap for `fzf` and `grep.actions`, the keymap from `grep.actions` takes precedence, and we receive only one element !

Please, do run the following:
```lua
fzf_lua.setup({
  keymap = {
    fzf = {
      ["ctrl-r"] = "select-all+accept",
    },
  },
  grep = {
    actions = {
      ["ctrl-r"] = function (selected)
        if #selected > 1 then
          vim.notify("Great Success !")
        else
          vim.notify("Damn, only one result.")  -- We always endup here !
        end
      end,
    },
  },
})
```

ibhagwan:
> Precisely because when using the same keymap for `fzf` and `grep.actions`, the keymap from `grep.actions` takes precedence, and we receive only one element !

You're correct!

> The only way I managed to make it work was to avoid setting any action on "ctrl-r", and only rely on the fzf keymap with a custom `fn_selected`.

Got it, you just gave me an idea to try and fix this scenario by not adding the action keybind to the `--expect` flag if it's already defined in `--bind` which would've solved the code snippet above where you define both an fzf keybind and a neovim action, doesn't seem to work though as when I remove the keybind from the `--expect` it sends an empty string in `selected[1]` as if the default keybind was pressed and not `ctrl-r`, try it with your original code snippet without definine `ctrl-r` as an action (only as an fzf bind) and you'll see it returns empty:
```lua
local function fn_selected_multi(selected, opts)
  if not selected then
    return
  end
  -- first element of "selected" is the keybind pressed
  print("keybind pressed", selected[1])  -- <== will return empty string if only an fzf bind is defined
  if #selected <= 2 then
    fzf_lua.actions.act(opts.actions, selected, opts)
  else -- here we multi-selected
    local _, entries = fzf_lua.actions.normalize_selected(opts.actions, selected)
    grep_sel_to_qf(entries, opts)
    vim.cmd("cfirst")
  end
end
```

I'll do some digging in the fzf manual again and see if it's possible to keep with the action with `--bind` while still defining a `--expect` for the same keybind, otherwise we can merge your PR though I'd still prefer this to work without any hacks or workarounds from the user's end.

ibhagwan:
Not sure if this is solveable via fzf, it seems that `--expect` is the only way to print the keybind pressed:
```
       --expect=KEY[,..]
              Comma-separated list of keys that can be used to complete fzf in
              addition to the default enter key. When this option is set, fzf
              will print the name of the key pressed as the first line of its
              output (or as the second line if --print-query is also used).
              The line will be empty if fzf is completed with the default
              enter key. If --expect option is specified multiple times, fzf
              will expect the union of the keys. --no-expect will clear the
              list.
```

ibhagwan:
Given I don't think this is possible I see no harm in merging this :-)

Ty and lmk if you can think of a better solution given what I posted above.

gjeusel:
Thank you very much !!!
 
 It's indeed a bit of a shame that the keypress is an empty string when the same keybind is used both in
 `--expect` and `--bind` indeed.
 
 And I perfectly understand that you would like to expose the **fzf-multi-select** to the users at a higher level.
 As this workaround of overriding the `fn_selected` only works if you know you want exactly one action to be applied when using the multi select - in my case, send to qflist.
 
 After some digging on my side, I came up with an in-between solution that might be acceptable:
 ```lua
 local fzf_lua = require("fzf-lua")

fzf_lua.setup({
  keymap = {
    fzf = {
      ["ctrl-m"] = "toggle-all",  -- not ctrl-a as it's for jumping to beginning of search query
    },
  },
  grep = {
    actions = {
      ["ctrl-r"] = function(selected)
        print(("Selected %s entries"):format(#selected))
      end,
    },
  },
})
 ```
 
 So the user might `toggle-all` with a first keypress, and then use a second keypress to choose its action.
 It's maybe worth adding a small section in the wiki called `fzf-multi-select`?
 I'll let you decide / do it if it's the case 👍 
 
 As for having one keypress for a multi-select action... I don't see any other way around than one of the followings:
 - opening an issue/discussion on the fzf github, asking for the reason why we got and empty string on the keypress when used in bind.
 - doing some hackish things when you spot the same bind in both options, removing the one in `--expect`, and replace the one in `--bind` with another fzf action being something like an `execute(echo 'ctrl-r' >> /tmp/path-to-outputtmpname)+select-all+accept`
 
Cheers 🎉

ibhagwan:
> So the user might toggle-all with a first keypress, and then use a second keypress to choose its action.
It's maybe worth adding a small section in the wiki called fzf-multi-select?
I'll let you decide / do it if it's the case

There already is a default keybind: `alt-a` (as you noted `ctrl-a` is start of line) for `toggle-all`, check out the [default options](https://github.com/ibhagwan/fzf-lua#default-options), the only caveat of the default mappings is the tables don’t get merged so once you defined `keymap.fzf` these will be your only fzf binds (as opposed to other tables where settings are “merged”) - this was done on purpose as keybinds are usually personal and opinionated and I didn’t users to have to define `<key> = false` to disable a bind.

> As for having one keypress for a multi-select action... I don't see any other way around than one of the followings …

I also believe the best route is opening an issue with fzf, I like this one the most:

> to modify the behavior of --expect to still execute the action in --bind while printing the name of the keypress in the case of a shared key binding.

PR #559, opened 2022-10-31
feat(run_command): provide user feedback in case of wrong command

micheleCTDE:
Won't happen often, but if a user mistypes a command they will be presented with some feedback like `[Fzf-lua] invalid command 'wrong_cmd_test'`


ibhagwan:
Technically will never happen but good change nontheless.

PR #560, opened 2022-10-31
code improvement(convert_user_opts): simplify the code for user option conversion

micheleCTDE:
Unless I am missing something, the code in this PR does exactly the same thing as the original code but it is simpler to read.
I guess `number` and `boolean` were part of the `switch` table for historical reasons.
Feel free to discard the PR if you prefer to keep the current code in case it could be useful in future.

### DIFF
--- a/lua/fzf-lua/cmd.lua
+++ b/lua/fzf-lua/cmd.lua
@@ -15,42 +15,14 @@ local bool_type = {
 }
 
 -- convert command line string arguments to
--- lua number boolean type and nil value
+-- lua boolean type and nil value
 local function convert_user_opts(user_opts)
-
-  local _switch = {
-    ["boolean"] = function(key, val)
-      if val == "false" then
-        user_opts[key] = false
-        return
-      end
-      user_opts[key] = true
-    end,
-    ["number"] = function(key, val)
-      user_opts[key] = tonumber(val)
-    end,
-    ["string"] = function(key, val)
-      if arg_value[val] ~= nil then
-        user_opts[key] = arg_value[val]
-        return
-      end
-
-      if bool_type[val] ~= nil then
-        user_opts[key] = bool_type[val]
-      end
-    end,
-  }
-
-  local _switch_metatable = {
-    __index = function(_, k)
-      utils.info(string.format("Type of %s does not match", k))
-    end,
-  }
-
-  setmetatable(_switch, _switch_metatable)
-
   for key, val in pairs(user_opts) do
-    _switch["string"](key, val)
+    if arg_value[val] ~= nil then
+      user_opts[key] = arg_value[val]
+    elseif bool_type[val] ~= nil then
+      user_opts[key] = bool_type[val]
+    end
   end
 end
 
@@ -77,7 +49,7 @@ local function run_command(args)
 
   if builtin[cmd] then
     builtin[cmd](opts)
-	else
+  else
     utils.info(string.format("invalid command '%s'", cmd))
   end
 end


ibhagwan:
> Unless I am missing something, the code in this PR does exactly the same thing as the original code but it is simpler to read.

It's possible, this is code I practically copy-pasted from Telescope a year ago, I'll test the changes and get back to you

ibhagwan:
I'm not sure if your new code handles failed type conversions properly, TBH I'm not even sure when this happens as I rarely use the command conversion (I usually use `:lua require...`), but it seems that Telescope still uses it and even added handling of table argument conversion:
https://github.com/nvim-telescope/telescope.nvim/blob/97847309cbffbb33e442f07b8877d20322a26922/lua/telescope/command.lua#L77

ibhagwan:
If anything, I would prefer this PR enables the ability to provide all types of arguments via the `:FzfLua` command and not just strings, for example, how would you send this using `:FzfLua`?
```lua
:lua require'fzf-lua'.files({ winopts = {relative="cursor"} })
```
I'm not well versed in vimscript but I think tables are `[]` so perhaps something like?
```vim
:FzfLua files winopts=[winopts=[relative=["cursor"]]
```

micheleCTDE:
The original code only uses the `_switch["string"]` function for all option pairs, where it checks for option values being nil or booleans. The functions `_switch["number"]` and `_switch["boolean"]` are never used AFAICT.

The [code](https://github.com/nvim-telescope/telescope.nvim/blob/97847309cbffbb33e442f07b8877d20322a26922/lua/telescope/command.lua#L133) in telescope uses all four types (number, boolean, string, table), hence the functions make sense in their code. Telescope seems to handle their options in a different way, checking the type and calling the corresponding function, while in fzf-lua we simply convert nil/boolean values and then pass the options to the respective fzf-lua command.

With that said, if you are not comfortable in merging this PR, we can abandon it.

ibhagwan:
> The original code only uses the _switch["string"] function for all option pairs, where it checks for option values being nil or booleans. The functions _switch["number"] and _switch["boolean"] are never used AFAICT.

Correct probably some hack I did early on to get this to work quickly and never bothered to properly fix.

> With that said, if you are not comfortable in merging this PR, we can abandon it.

If we’re already changing this part of code to work properly I’d rather improve it to handle all data types as opposed to dumbing it down.

micheleCTDE:
> If we’re already changing this part of code to work properly I’d rather improve it to handle all data types as opposed to dumbing it down.

Sounds good :+1: I will close this PR and we can probably look at what telescope is doing in more details :-)

micheleCTDE:
@ibhagwan 
I have been looking into telescope code in more detail. I don't yet have enough knowledge of the code of fzf-lua (still lot of code to study), so I would like to ask you how are options parsed in fzf-lua. For example if we pass  a`winopts` table as option to a fzf-lua command, are the contents parsed and validated somewhere in the code while executing the command? Since we are not doing that in the `convert_user_opts` method, I assume that is the case, but thought of asking since you know the code in much more detail.

ibhagwan:
Options aren’t validated, it can still raise an exception if you send string where table is due, etc, some options are parsed and additional logic is also applied which merges all options to one table (global, provider and call options), it all happens here:

https://github.com/ibhagwan/fzf-lua/blob/b1f1f76c13d6040b9b9b6ccdb59129c6c4885124/lua/fzf-lua/config.lua#L733

micheleCTDE:
ok, thanks for the answer. Telescope is doing something like this:
1. for some special keys (`split_keywords`), it will call `_switch["table"]` where it will try to evaluate the value passed in. By the way I am not even sure the telescope code is fully correct, because `loadstring` returns a function, so `eval, err = loadstring("return " .. val)` looks wrong to me (but I may be missing something??)
2. for `default_opts` (config values), it invokes `_switch` with the actual type and deal with it accordingly (numbers, booleans, strings, tables, other type (consider invalid)
3. for numbers, convert the option to a number
4. for everything else, treat it as a string option.

The question is what we want to do in fzf-lua. Point 4 is already there. Point 4 seems good to add. Point 1 seems very specific to some telescope options. Point 2 seems a bit redundant based on the facts that options are later checked and normalized in the rest of the code. And beside point 2 does not check for nil values.

How about we rework the current code to implement point 3 and 4? The code from this PR could be reused and extended to include point 3. What is your opinion?

ibhagwan:
Your analysis is correct, that was my impression when I looked at the Telescope  code.

The fact it’s done in a certain way in Telescope does not necessarily mean this is the “correct” way or it has no faults, it was just a hint as to what may be possible or the intention of the code, in fact, I’m pretty sure I can come up with a generic and more efficient way to do this.

Being totally honest this PR is in the “nice to have” category or less, this was never requested or even mentioned by users, “if it works don’t touch it”, this isn’t a glaring issue with the code and has never affected a single user, I’d argue there are better places to spend your time on this project’s code than this.

micheleCTDE:
>  “if it works don’t touch it”, this isn’t a glaring issue with the code and has never affected a single user, I’d argue there are better places to spend your time on this project’s code than this.

Sounds good, then let's leave this part of the code as is and settle here. I will continue go on with the studies of the rest of the plugin code.

PR #583, opened 2022-12-16
docs: add language annotations

comfortablynick:
Update docs to be able to use treesitter syntax highlighting

### DIFF
--- a/doc/fzf-lua.txt
+++ b/doc/fzf-lua.txt
@@ -91,14 +91,14 @@ INSTALLATION                                            *fzf-lua-installation*
 
 Using vim-plug <https://github.com/junegunn/vim-plug>
 
->
+>vim
     Plug 'ibhagwan/fzf-lua', {'branch': 'main'}
     " optional for icon support
     Plug 'nvim-tree/nvim-web-devicons'
 <
 Using packer.nvim <https://github.com/wbthomason/packer.nvim>
 
->
+>lua
     use { 'ibhagwan/fzf-lua',
       -- optional for icon support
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for this PR, it makes sense, the issue is this document is auto generated by [ts-vimdoc.nvim](https://github.com/ibhagwan/ts-vimdoc.nvim) so a one time fix wouldn’t be useful as it will get overwritten the next time README is updated.

This would have to be done in 2 different PRs, adding this functionality to `ts-vimdoc` and then updating the README accordingly.

comfortablynick:
Didn't know that. This won't be very useful, then!

I assume there's some way for your tool to take the language of the code from Markdown and translate it to vimdoc.

ibhagwan:
> Didn't know that. This won't be very useful, then!
> 
> I assume there's some way for your tool to take the language of the code from Markdown and translate it to vimdoc.

Of course there is (via tree-sitter which is already used to parse the markdown), it’s just has to be coded into the tool now :)

ibhagwan:
FYI, 

Added this functionality was added to `ts-vimdoc.nvim` in: https://github.com/ibhagwan/ts-vimdoc.nvim/commit/eb25fc912ff73cae9a003158a7413448ccdb20e5

Also added it as a CI workflow (https://github.com/ibhagwan/fzf-lua/commit/72cf10ea41819a95e29c114c744b6376642ce3e0) and applied to the main branch (https://github.com/ibhagwan/fzf-lua/commit/725276b860b041d5d93232d5ea212c901b4a00ff).

PR #644, opened 2023-02-14
Add filter for cwd to the buffer opts.

asmodeus812:
Add filtering for a given directory with the buffers list.


ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/d6bb1c02ef72ee64653e2ff3d6b8a451fc40bf9b

PR #647, opened 2023-02-15
Apply keymap when select.

liangxianzhe:
Hello, trying to contribute here. Hope it makes sense.

Though I notice you use the `│` instead of `|` as the divider. It took me a few minutes to figure it out :D

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -452,6 +452,11 @@ M.goto_jump = function(selected, opts)
   end
 end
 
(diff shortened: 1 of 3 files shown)


ibhagwan:
> Hello, trying to contribute here. Hope it makes sense.
> 
> Though I notice you use the `│` instead of `|` as the divider. It took me a few minutes to figure it out :D

Tysm @liangxianzhe, this is great!

Actually regarding the separator it wasn't me lol, it was contiributed by @blob42 in #504.

PR #650, opened 2023-02-16
Add more generic handling for headers for all different pickers.

asmodeus812:
This extends the default table of actions for the header builder, to support a more generic approach to all existing actions, by using a simple function reference to key the table of action definitions. This can then be setup / extended to handle common actions and in the future the action definitions table could also be overrideable by the user so he can choose which actions to actually show / hide. This also allows us to use the header builder function pretty much in all pickers, and not have specific implementations per picker as it was for buffers for example.


lua/fzf-lua/core.lua
```diff
@@ -569,12 +576,10 @@ M.set_header = function(opts, hdr_tbl)
       val = function()
         -- do not display header when we're inside our
         -- cwd unless the caller specifically requested
-        if opts.show_cwd_header == false or
-            not opts.show_cwd_header and
-            (not opts.cwd or opts.cwd == vim.loop.cwd()) then
-          return
+        if opts.show_cwd_header == true or opts.show_cwd_header then
```
ibhagwan, line 579: This condition is redundant, both evauluate to the same thing, this change also messes up showing `cwd` header by default when using on any folder outside of `cwd` - I will restore this in the manual merge.

lua/fzf-lua/core.lua
```diff
@@ -569,12 +576,10 @@ M.set_header = function(opts, hdr_tbl)
       val = function()
         -- do not display header when we're inside our
         -- cwd unless the caller specifically requested
-        if opts.show_cwd_header == false or
-            not opts.show_cwd_header and
```
ibhagwan, line 573: Here the 2nd condition makes sense because `opts.show_cwd_header == false` is not the same as `not opts.show_cwd_header` which evaulates to `true` when the variable equals both `false` and `nil` (undefined), I will change it to `== nil` so it's clearer.

ibhagwan:
Ty @asmodeus812 for the contribution!

https://github.com/ibhagwan/fzf-lua/commit/178709f0839359ffbd44fcaca64558212abe2493 - made minor modifications (which I commented about above) and rebase-merged manualy.

PR #656, opened 2023-02-17
added option to provide marks to builtin marks-provider

abenz1267:
I've added logic in order to be able to provide marks to the builtin marks-provider. This way you can filter the marks to what you need.

```
marks = {
    marks = "abcdef"
}
```


ibhagwan:
Ty @abenz1267!

Minor modification, the code you wrote works perfectly  but I would change this to a one-liner, from:
```lua
  if opts.marks ~= nil then
    marks = vim.fn.execute("marks " .. opts.marks)
  else
    marks = vim.fn.execute("marks")
  end
```
to:

```lua
local marks = vim.fn.execute(string.format("marks %s", opts.marks and opts.marks or ""))
```

Dw if you don’t get to it I will merge manually soon.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/83c4cf8b91a7725cf36d1217d1220d6bf958c014 - done, tysm @abenz1267!

Issue #669, opened 2023-02-25
Question: Combining results from lsp_refs , lsp_defs


sandangel:
### Info

Hi, I think I asked this question before, now I'm trying to implement it.

I'm trying to reuse the code to query lsp results from fzf-lua but [the function is not exposed](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/providers/lsp.lua#L358)

I have this so far. 

```lua
      vim.keymap.set('n', 'gd', function()
        local lsp = require('fzf-lua.providers.lsp')
        lsp.definitions()
        -- lsp_typedefs = { "fzf-lua.providers.lsp", "typedefs" },
        -- lsp_references = { "fzf-lua.providers.lsp", "references" },
        -- lsp_definitions = { "fzf-lua.providers.lsp", "definitions" },
        -- lsp_declarations = { "fzf-lua.providers.lsp", "declarations" },
        -- lsp_implementations = { "fzf-lua.providers.lsp", "implementations" },
      end, opts)

```

Is it possible that we can expose those function or make it easier to reuse? Like exposing the functions those returning the results LSP only (not open fzf windows or something else), then I can combine them together and feed them to `fzf_exec`.

- Operating System:
- Shell:
- Terminal:
- `nvim --version`:
- `fzf --version`:

- [ ] The issue is reproducible with `mini.sh`

<details>
<summary>fzf-lua configuration</summary>

```lua
require('fzf-lua').setup({
})
```
</details>

ibhagwan:
`definitions` is exposed publicly:
https://github.com/ibhagwan/fzf-lua/blob/15e9b1f1494dc1facaffe9543281d015c43b9eba/lua/fzf-lua/providers/lsp.lua#L373

but it’s not returning a simple set of results as your code suggests, it’s a bit more complicated with callback handlers (for the async option), your best bet is copying parts of the code and modifying it to woke shnc only and combine the items in one big table, or alternatively (but more complex), use the async method and keep the pipe open until all LSP callbacks from all your calls are complete.

sandangel:
> alternatively (but more complex), use the async method and keep the pipe open until all LSP callbacks from all your calls are complete.

I think  this one would be ideal. Could you share with me which code should I copy from?

ibhagwan:
You'll have to follow this function, explore what it does and then copy the relevant parts:
[...]

PR #671, opened 2023-02-27
feat: add lsp finder

sandangel:
I added the function for #669  . It's working as expected, but the UI was lagging a bit. so I think the async version is better.

Also I think I can combine `set_lsp_finder_fzf_fn` with `set_lsp_fzf_fn`, checking when method is a table or a string.

I guess we can also improve check_capabilities. When `lsp_handler.method` is a table, but 1 item is not supported by lsp, we can remove that item from `method`. 

Is this a good start? Will you open to accept this PR if I make the above improvement?

### DIFF
--- a/lua/fzf-lua/init.lua
+++ b/lua/fzf-lua/init.lua
@@ -28,7 +28,6 @@ do
   if not vim.g.fzf_lua_server then
     vim.g.fzf_lua_server = vim.fn.serverstart()
   end
-
 end
 
 local M = {}
@@ -110,7 +109,7 @@ function M.setup(opts, do_not_reset_defaults)
   -- backward compatibility for 'fzf_binds'
   if opts.fzf_binds then
     utils.warn("'fzf_binds' is deprecated, moved under 'keymap.fzf', " ..
-      "see ':help fzf-lua-customization'")
+    "see ':help fzf-lua-customization'")
     globals.keymap.fzf = opts.fzf_binds
   end
   -- do not merge, override the bind tables
@@ -215,6 +214,7 @@ do
     spell_suggest = { "fzf-lua.providers.nvim", "spell_suggest" },
     filetypes = { "fzf-lua.providers.nvim", "filetypes" },
     packadd = { "fzf-lua.providers.nvim", "packadd" },
+    lsp_finder = { "fzf-lua.providers.lsp", "finder" },
     lsp_typedefs = { "fzf-lua.providers.lsp", "typedefs" },
     lsp_references = { "fzf-lua.providers.lsp", "references" },
     lsp_definitions = { "fzf-lua.providers.lsp", "definitions" },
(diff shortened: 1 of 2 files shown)


sandangel:
also what are you using for code formatting? I'm using vim.lsp.buf.format get a different format.

ibhagwan:
Ty @sandangel for the PR.

> Is this a good start? Will you open to accept this PR if I make the above improvement?

It’s a good start but if I were to merge this I would have to do this differently which is I haven’t done it yet, this is just one use case out of many and each of these requires new code which duplicates inner code, in addition it just combines the results without differentiating the source of entry (whether it cam from references, definitions, etc).

I would prefer a more generic approach where I you can combine as many providers as you wish and optionally add a prefix so you can distinguish the source of the entry, this obviously would take much more effort and I still haven’t decided on doing it.

> I guess we can also improve check_capabilities. When lsp_handler.method is a table, but 1 item is not supported by lsp, we can remove that item from method.

The reason the code is still there was backward comparability before lspconfig dropped support for older versions of neoivm, I agree it can safely be removed which I will do at one point.

ibhagwan:
> also what are you using for code formatting? I'm using vim.lsp.buf.format with lua_ls but get a different format.

Regarding this, I use the `lua_ls` formatter but it had lots of changes recently as well as some bugs (https://github.com/CppCXY/EmmyLuaCodeStyle/issues/89) and I haven't updated `.editorconfig` yet so the formatting result is different.

sandangel:
how about we introduce this new lsp_finder first for combining lsp related information only. At the point when we decide to work on a more generic version then we can refactor it? So it will be a small improvement over time instead of a big effort that we hesitate to do.
 
I think I add differentiating the source of entry. And all the improvements for the current code I mentioned earlier. Then we can merge. What do you think?

sandangel:
I can add the async version and help maintain this code. does it sound good?

ibhagwan:
> I can add the async version and help maintain this code. I will also improve it over time. does it sound good?

I thought about it and I’d rather not, the LSP code is due for a refactor and removal of deprecated settings (as you pointed out), I prefer doing it all at once together with adding a generic approach for combining providers (of all kinds, not just LSP)  rather than have to this addition into account as well.

I can’t promise when I’ll do it but it’s on the table.

sandangel:
Thank you for saying that. Looking forward to it 😀

sandangel:
Awesomeeeeee. Thanks alot.

sandangel:
I will close this PR then. and again, thank you very much for supporting this. i think many users can consider this as a killer feature for them to switch fzf-lua, which already a lot faster than telescope.

ibhagwan:
Completed in https://github.com/ibhagwan/fzf-lua/commit/a81d18ee591635ce537bfa045ed1edf7dbe30cf3

PR #674, opened 2023-03-06
feat(keymaps): sort keymaps alphabetically

liangxianzhe:
It looks nicer and allows better tiebreak when filtering.


ibhagwan:
Ty @liangxianzhe!

Merged manually and added `--tiebreak=index` to default `fzf_opts`.

PR #676, opened 2023-03-07
fix(colorschemes): not load colorschemes if passed in.

liangxianzhe:
Well, this is a breaking change and I'm not sure if you will like it or not.

What I want to do is to create a custom command to show a subset of colorschemes. It is impossible to do without introducing this breaking change or adding a new parameter.

I think the new behavior is easier to use. If people want to extend the list, they can always use `vim.fn.getcompletion` themselves.


ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/78e44ff485bd02cce53454780a405e5b789bcc1c - merged.

Actually your change makes more sense and is more consistent with other filters, TBH I don't know of anyone using this option so it's not really breaking :-)

PR #679, opened 2023-03-08
fix: make opening file path relative to vims cwd (closes #678)

### LINKED ISSUE #678: Opening a file (e.g. via git_files) always opens file with absolute path.
### Info

- Operating System: ubuntu/nix
- Shell: zsh
- Terminal: st
- `nvim --version`: v0.8.3
- `fzf --version`: 0.38.0

<details>

```lua
require("fzf-lua").setup({
    fzf_bin = "fzf-tmux",
    fzf_opts = {
        ["--border"] = "rounded",
    },
    fzf_tmux_opts = {
        ["-p"] = "",
        ["-y"] = "1",
        ["-w"] = "95%",
        ["-h"] = "75%",
    },
    winopts = {
        preview = {
            hidden = "hidden", -- I don't like a preview, keep space for selection
        },
    },
})

require("fzf-lua").register_ui_select()

```
</details>

### Description 

Opening a file (e.g. via git_files) always opens file with absolute path.
I would expect the file to open relativ to current working dir of neovim instance.

The issue with absolute paths is that other plugins will work on current buffer path and will not properly react to the working dir anymore.
I'll handle in an PR with a proposal.

This is probably releated to #341 and b273a358a7a33bebcf6e5c1ef1adc25439083023.

DanielSiepmann:


DanielSiepmann:
I'm not a lua expert. And I use this plugin and fzf since a week. So there might be issues with my proposal.

ibhagwan:
> I'm not a lua expert. And I use this plugin and fzf since a week. So there might be issues with my proposal.

Unfortunately this will break almost everything when using the plug-in on anything other than `cwd`.

I’ll take a look it your issue, it’s still not clear to me why editing using full paths causes issues, AFAIK he buffer name is always full path no matter how it was opened.

In any event, if this is to be fixed it should be fixed as part of the open file action.

DanielSiepmann:
> Unfortunately this will break almost everything when using the plug-in on anything other than `cwd`.
I tested with builtin, buffers and btags, beside git_files. They all still work for me. I don't see why this should break, as it would strip off the cwd, which should be fine if it doesn't exist?

> why editing using full paths causes issues, AFAIK he buffer name is always full path no matter how it was opened.
` expand('%:r')` will expand to the full path instead of relative path, causing trouble in my setup.

`set statusline+=\ %f` also shows the "full" path with `~` instead of relative path, consuming unnecessary space for information I don't care about.

I didn't had the issue with telescope and ctrlp, which I used before.

I've the patch applied in my setup and will comment if I encounter any issue. But maybe my explanation already helps? I'm also fine with having my own patch, but thought this might be an improvement. I won't be disappointed if it would be closed without merging :)

lua/fzf-lua/path.lua
```diff
@@ -234,6 +234,7 @@ function M.entry_to_file(entry, opts, force_uri)
   if cwd and #cwd > 0 and not isURI and
       not M.starts_with_separator(stripped) then
     stripped = M.join({ cwd, stripped })
+    stripped = M.relative(stripped, vim.fn.getcwd())
```
DanielSiepmann, line 237: Maybe we should add a comment like:
`-- #678: strip off vims cwd`

ibhagwan:
> I tested with builtin, buffers and btags, beside git_files. They all still work for me. I don't see why this should break, as it would strip off the cwd, which should be fine if it doesn't exist?

This will not break using default options, this will break when using FzfLua with the `cwd` option, it will also break when using providers with previews outside of the cwd (LSP symbols, etc) and many others issues I can’t forecast ATM, guaranteed :)

Responding to the rest in #678.

ibhagwan:
> In any event, if this is to be fixed it should be fixed as part of the open file action.

https://github.com/ibhagwan/fzf-lua/commit/592ab9a26176a7f9ebd43faf91d1f5d066e0aa07

PR #680, opened 2023-03-09
feat(symbols): add space to child symbols.

liangxianzhe:
Also removes `[]` from `symbo_style 2 (icon only)`.

This gives a nice tree-like view for document_symbols.

symbol


ibhagwan:
Ty @liangxianzhe, looks great :)

I have two minor comments:
1. I’m not sure how prone to changes are the internal LSP until functions, I’d rather rely on the public `lsp.util` APIs than to replicate internal code and call `_` functions.
2. I’m also unsure about having this as default and non optional, users being users some have different preferences and would like to not have an indent in a fuzzy search interface as it can look weird after starting to type

That said, I’d love to accept this (perhaps even make it default) as long as users have an option to revert to the previous look and feel, the unchanged version should also call the original `vim.lsp.util.symbols_to_items` just in case something changes in the API so users can have a workaround until a bug fix is available.

What do you think?

liangxianzhe:
Hi, see the last commit.

1. Changed to not rely on `_` internal function, instead, inline it.
2. The `child_prefix` option can be used to restore the original non-indented look, the user just needs to set it to empty or nil.
3. Changed to use `vim.lsp.util.symbols_to_items` if `child_prefix` is not set.

On a side note, there are some issues with the existing symbol icon resolution logic. I'm planning to send another PR once this is in.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/fd825cc6937c774fd7d8dd8573100e0ce3fbaa00 - few changes:

- Removed the exception for `symbol_style=2`, it's incosistent with the rest of the options and changes default behavior, if someone wishes to override that it's now possible due to `opts` being sent as 2nd argument (left that intact)
- changed `child_prefix` default to `true`, users don't need to be aware of fzf/UTF woes, and the characters you chose have caused an issue in the past with filenames that contain those (very rare), instead used `utils.nbsp` which has proven track record now. You can still set `child_prefix` to whatever you want if you wish to control that (don't see why but it's still possible).

ibhagwan:
> On a side note, there are some issues with the existing symbol icon resolution logic. I'm planning to send another PR once this is in.

Tysm!

liangxianzhe:
Makes sense, thanks!

ibhagwan:
> Makes sense, thanks!

Actually, the space char "fix" needed to be reverted, it's only meant to be used for prefixes and will cause issues with anything that comes after the file.

https://github.com/ibhagwan/fzf-lua/commit/a553b614f1b89fdbf645efef977f4b7aa0fa6c19

liangxianzhe:
Hi @ibhagwan , just notice that `utils.nbsp` will break the preview window for indented items (mac, alacritty), while `\xc2\xa0` works fine. I saw you played with these magical codes before, any idea how to debug this?

liangxianzhe:
Actually, saw your fix. Thanks!

ibhagwan:
> Actually, saw your fix. Thanks!

Yes, this was my bad :)

I needed a character that isn’t visible and isn’t space and Thad nobody uses, it is used to distinguish all the file prefix markers (icons, git icons, buffer num, etc), it works very well for anything that comes **before the file**:
https://github.com/ibhagwan/fzf-lua/blob/a553b614f1b89fdbf645efef977f4b7aa0fa6c19/lua/fzf-lua/path.lua#L230

The file end is then marked with `:` and anything that comes after is valid, we could equally use regular space here to make the indent but `A0` will work just as fine.

PR #682, opened 2023-03-11
feat(symbols): better symbol kind icons.

liangxianzhe:
Addresses a few issues. See attached images:

1. The existing method of using CmpItemKind is not right.

CmpItemKind was designed for completion, not for symbols. It includes things like Snippet, but doesn't cover all the symbols defined in LSP specs, such as String, Array, and Object.

Additionally, fetching icons from the protocol object is not the best practice in my mind. The default protocol object is just a symbol name and id. There are a few plugins using this practice, but many others (cmp, aerial, symbol-outline) ask users to set icons directly.

Providing some default icons is something debatable. We can also not provide any default icons, and rely on the user to set them manually.

2. The highlight groups are changed.

`TSFunction` is already removed from Treesitter implementation. The default one now is `@function`.

3. A minor issue when substituting `[Kind]`.

It should only substitute the first match, otherwise, if the symbol name includes brackets (e.g. `[-nth=foo]`), it will be substituted too.

Before:
before

After:
after


ibhagwan:
Ty @liangxianzhe!

Left a few minor comments in the code.

> CmpItemKind was designed for completion, not for symbols. It includes things like Snippet, but doesn't cover all the symbols defined in LSP specs, such as String, Array, and Object.

I’m not a big fan of having to dup code or maintain a list (of symbols) where it’s defined elsewhere, but I wasn’t aware of the above, one issue the approach you’ve taken is that now if someone customized their completion icons it won’t be picked up automatically and they’ll have to go to the README figure out where to change the setting and redefine their icon preferences.

Isn’t it better we first check `vim.lsp.protocol.CompletionItemKind` and only define the extra icons we are missing?

> Providing some default icons is something debatable. We can also not provide any default icons, and rely on the user to set them manually.

I really don’t like the idea of having to ask the users to configure stuff, I purposefully made it so one doesn’t even have to call `setup`, my philosophy is that plugins should “just work” with sane defaults (including icons) and users only have to dig deeper if they’re interested in advanced customization.

> The highlight groups are changed.

Ty, I wasn’t aware of this change, I assume these cover the new types too (String, Number, etc).

liangxianzhe:
> Isn’t it better we first check vim.lsp.protocol.CompletionItemKind and only define the extra icons we are missing?

It seems to me that setting icons into CompletionItemKind is a hack that is not widely used. If you look at the [code](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/lsp/protocol.lua#L66), CompletionItemKind is just a simple map without any icons.

If we look at nvim-cmp, it only provides a [format function](https://github.com/hrsh7th/nvim-cmp/wiki/Menu-Appearance#basic-customisations). Hacking CompletionItemKind to include icons was done only if people use lspkind.nvim, or set it themselves. Even if they do, we cannot assume they only set icons, they can totally set icon plus kind name into CompletionItemKind. It will not work well with the `symbol_style` option.

Given all these, I'm not a fan of relying on CompletionItemKind, or extracting "icons" from it.

> I really don’t like the idea of having to ask the users to configure stuff

Me too, that's why I added default icons in this PR. Though if people want the same icons across cmp, fzf, aerial, etc, they need to set them manually anyway. At this moment, there is no solution that can automatically unify them.

> I assume these cover the new types too (String, Number, etc).

Yes.

ibhagwan:
> Given all these, I'm not a fan of relying on CompletionItemKind, or extracting "icons" from it.

Makes sense, I used my anecdotal “evidence” looking into how people configure their neovim and it was widely used but I get your point and it doesn’t seem reliable.

> Me too, that's why I added default icons in this PR. Though if people want the same icons across cmp, fzf, aerial, etc, they need to set them manually anyway. At this moment, there is no solution that can automatically unify them.

A bit of an annoyance to have to configure yet another set of icons, one day perhaps a unified icon map will arrive in neovim and we can rethink this, until then, I’m with you in support of having default icons (as opposed to not having any and forcing the user to read the docs), can `lspkind` icons be considered the “standard” here in the same way `nvim-web-devicons` is for files (I’m assuming you copied from `lspkind`)?

liangxianzhe:
I copied from `aerial.nvim`, which is copied from `lspkind`. The `lspkind` itself doesn't cover all Lsp symbols since it is mostly for CompletionItemKind. I know, it is a little bit messy.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/c0427c71a5fca05b31f010ee5d2996aa567259b2 - Ty @liangxianzhe!

PR #687, opened 2023-03-17
Make grep input cancelable

### LINKED ISSUE #686: Canceling the input of grep does works the same way as passing no input
Here
https://github.com/ibhagwan/fzf-lua/blob/d666abc60fff12af9e91812666d42ff7f2555934/lua/fzf-lua/providers/grep.lua#L129

if the user closes the input the grep runs anyway,  which is a bit annoying. IMHO one should be allowed to abort the operation (by pressing escape or <C-c>), for an empty pattern it's still possible to hit enter.

leiserfg:


leiserfg:
This fixes   #686

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/8ca21b779f61b484ab1b6612fcbc707624982a27 - ty @leiserfg!

PR #698, opened 2023-03-23
Add support for nvim 0.9

oswalpalash:


ibhagwan:
Not sure I understand what this PR does, `:lua print(vim.fn.has("nvim-0.8"))` returns `1` on both `0.8` and `0.9` nightly (as 0.9 "has" 0.8 in it), in addition, you're defining a `local fallback` variable within the `if-then` scope, this makes no sense.

What problem are you trying to solve?

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -712,7 +712,13 @@ function Previewer.buffer_or_file:do_syntax(entry)
         -- limit treesitter manual attachment to 0.8 instead (0.7.2 also errs)
         -- Error executing vim.schedule lua callback:
         --   vim/filetype.lua:0: attempt to call method 'gsub' (a nil value)
-        local fallback = vim.fn.has("nvim-0.8") == 0
+        local fallback = 0
+        if ( vim.fn.has("nvim-0.8") == 0 ) then
+          local fallback = 1
```
ibhagwan, line 717: This does nothing, a local `fallback` variable is limited to the `if-then` scope.

ibhagwan:
Closing, feel free to reopen after clarification of the code and purpose of this PR.

PR #703, opened 2023-03-27
feat(git_status) add git_reset action

liangxianzhe:
I guess it was designed for git_status? Not using `action.resume` to be safe.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -592,6 +592,7 @@ end
 M.git_reset = function(selected, opts)
   local cmd = path.git_cwd({ "git", "checkout", "HEAD", "--" }, opts)
   git_exec(selected, opts, cmd)
+  vim.cmd("e!")
 end
 
 M.git_stash_drop = function(selected, opts)
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -241,6 +241,7 @@ M.defaults.git = {
     actions     = {
       ["right"] = { actions.git_unstage, actions.resume },
       ["left"]  = { actions.git_stage, actions.resume },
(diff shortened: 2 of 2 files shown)


ibhagwan:
The reason I didn’t add this until now by default is it was incomplete, `e!` is only good for the current buffer whereas the power of the action comes from being able to reset multiple selection.

In addition, I would add the interactive header (for example,  the “`ctrl-x` to close” in buffers ), otherwise the risk is too big someone coming from another plugins which uses `ctrl-x` to open in a split and lose all their changes.

liangxianzhe:
> e! is only good for the current buffer whereas the power of the action comes from being able to reset multiple selection.

Then stash_apply has the same issue?

> would add the interactive header

Or maybe pop up some confirmation box like stash_apply?

ibhagwan:
> > e! is only good for the current buffer whereas the power of the action comes from being able to reset multiple selection.
> 
> Then stash_apply has the same issue?

Probably, even better reasoning to find a solution for both.

> 
> > would add the interactive header
> 
> Or maybe pop up some confirmation box like stash_apply?

perhaps an optional popup? I would assume some users would get annoyed with that popup after a while, regardless, the header should be added (can be disabled).

liangxianzhe:
> find a solution for both
A quick one would be `M.vimcmd_file("e!", selected, opts)`, though it is not ideal since it opens and jumps to the files. A better solution will need more work.

ibhagwan:
>  A better solution will need more work.

which is the main reason it isn’t yet default

liangxianzhe:
I'm closing this for now and will reopen once I have more time. Thanks!

ibhagwan:
@liangxianzhe, see #730, while the file reload issue is still not solved with it I figured it’s still worth the benefit and we can solve the reload at a later time.

PR #722, opened 2023-04-08
fix(diagnostic): buffer might have become invalid

tomtomjhj:
Since diag_results is used in in vim.schedule()d callback, the buffers may have become invalid when the callback is executed. So the buffers' validity should be checked.

### DIFF
--- a/lua/fzf-lua/providers/diagnostic.lua
+++ b/lua/fzf-lua/providers/diagnostic.lua
@@ -132,6 +132,9 @@ M.diagnostics = function(opts)
 
   local preprocess_diag = function(diag, bufnr)
     bufnr = bufnr or diag.bufnr
+    if not vim.api.nvim_buf_is_valid(bufnr) then
+      return nil
+    end
     local filename = vim.api.nvim_buf_get_name(bufnr)
     -- pre vim.diagnostic (vim.lsp.diagnostic)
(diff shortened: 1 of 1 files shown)


tomtomjhj:
Sorry for the lack of the context. This may happen when you :bwipeout some buffers, but unfortunately I don't have a stable reproducer due to this bug's nature. The last time I encountered this bug, I was editing LaTeX files with texlab, renamed some file using fern.vim (which :bwipeout the buffer with the old name, I guess), got some weird diagnostics from texlab, recompiled the project, and ran fzf-lua's diagnostics_workspace. At that point, fzf-lua raise the following error: 
```
Error executing vim.schedule lua callback: ...vim/plugged/fzf-lua/lua/fzf-lua/providers/diagnostic.lua:135: Invalid buffer id: 136
stack traceback:
    [C]: in function 'nvim_buf_get_name'
    ...vim/plugged/fzf-lua/lua/fzf-lua/providers/diagnostic.lua:135: in function 'preprocess_diag'
    ...vim/plugged/fzf-lua/lua/fzf-lua/providers/diagnostic.lua:166: in function <...vim/plugged/fzf-lua/lua/fzf-lua/providers/diagnostic.lua:165>
```

ibhagwan:
Ty @tomtomjhj, that’s quite the scenario :)

merged.

PR #741, opened 2023-05-03
fix(icons): match new Nerd Fonts 3.0.0 release

XenoPhex:
See the following links for more information:
- https://github.com/ryanoasis/nerd-fonts/issues/365
- https://github.com/ryanoasis/nerd-fonts/issues/1059#issuecomment-1404891287


ibhagwan:
Is this backward compatible or does it force all users to use v3.0 to see the new icons?

Perhaps anecdotal but most users bother updating the font regularly (unless it comes with the distro) so this might have unintended consequences for existing users not being able to see some icons.

XenoPhex:
Unfortunately this is not backwards compatible AFAIK. So I'm unsure what the best course of action is when trying to maintain support for older [font] versions.

From experience - most people that use Nerd Fonts likely use some automated package manager to install them (so you probably don't need to worry about the system font use case). For people on MacOS - this means they're likely using `brew` which is pretty aggressive about updates. For people on other platforms 🤷🏽 

I'm personally of the opinion that people should do their best to stay reasonably up to date / within the last few releases - regardless of major/minor/patch update. So if anything, it might be worth waiting a few days (or weeks?) to give people time to update. The only issue is holding off _disincentives_ people from updating / causes some pain for those that have.

ibhagwan:
> The only issue is holding off disincentives people from updating / causes some pain for those that have.

AFAIK all of the changes in this commit are **customizable via setup** so I’m pretty sure this can wait until we figure out what’s the right thing to do here.

XenoPhex:
Yeah that's fair. I can do the same until they're updated on main.

Thank you for the responsive reply!

mhanberg:
For those wanting something to copy and paste, this worked for me

```lua
      require("fzf-lua").setup {
        lsp = {
          symbols = {
            symbol_icons = {
              File = "󰈙",
              Module = "",
              Namespace = "󰦮",
              Package = "",
              Class = "󰆧",
              Method = "󰊕",
              Property = "",
              Field = "",
              Constructor = "",
              Enum = "",
              Interface = "",
              Function = "󰊕",
              Variable = "󰀫",
              Constant = "󰏿",
              String = "",
              Number = "󰎠",
              Boolean = "󰨙",
              Array = "󱡠",
              Object = "",
              Key = "󰌋",
              Null = "󰟢",
              EnumMember = "",
              Struct = "󰆼",
              Event = "",
              Operator = "󰆕",
              TypeParameter = "󰗴",
            },
          },
        },
      }
```

XenoPhex:
For those looking for a fuller list of icons that popular tools are using, feel free to use mine - I updated it recently to finish the 3.0.0 release: https://github.com/XenoPhex/nvim/blob/42192ef0c637c3ae6e274df4b7b2df633577a72c/lua/utils/icons.lua#L43-L160

ibhagwan:
> Unfortunately this is not backwards compatible AFAIK. So I'm unsure what the best course of action is when trying to maintain support for older [font] versions.

@XenoPhex, I looked into this, there's not much we can do here when it comes to backward compatibility, given nvim-web-devicons also updated to nerdfots v3 I guess we should merge this and if people need to use the older fonts they can always either pin commit customize `symbol_icons`, ref:
https://github.com/nvim-tree/nvim-web-devicons/pull/264

PR #749, opened 2023-05-07
fix(commands): `ex_run_cr` without showing putting the command into Vim cmdline, should close #747

### LINKED ISSUE #747: Fire commands immediately after selection
### Description 

Is it possible to fire the command immediately after I select the command from the `:FzfLua commands`, instead of putting the selected command into the command mode, which requires hitting the enter again?

nyngwang:
I also tested that `stopinsert` is not needed by removing it and tried `:Inspect`. It didn't go into insert mode without it.

Preview: (no cmdline in after the selection comparing to the original one.)

https://user-images.githubusercontent.com/24765272/236675075-942e297a-c03d-485a-b0df-9cc1ab2a0692.mov

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -403,8 +403,8 @@ M.ex_run = function(selected)
 end
 
 M.ex_run_cr = function(selected)
-  local cmd = M.ex_run(selected)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Merged, ty @nyngwang!

PR #750, opened 2023-05-08
feat: add logger

linrongbin16:
1. add logger module
2. setup logger in init setup
3. add description in README

Manual tested.
cmdline log (left-bottom corner):
image

file log:
[fzf-lua.log](https://github.com/ibhagwan/fzf-lua/files/11418659/fzf-lua.log)


ibhagwan:
Ty @linrongbin16!

As I mentioned in the other issue I welcome a PR for windows support, a logger module without extensive logging/debugging calls isn’t useful IMHO and I personally don’t need it for development, you can use it in your own fork until you finalize the windows support but I don’t intend to blanket add this to the plug-in as this is useful only for developers.

I also don’t intend to spam the code with debugging calls, there are barely any issues and the code is mature and stable.

linrongbin16:
OK

PR #758, opened 2023-05-15
refactor: use treesitter functions from upstream

gpanders:
Drop (at least one) dependency on nvim-treesitter and simply use the vim.treesitter API provided by upstream Nvim.

Note that nvim-treesitter's `ft_to_lang` is already (informally) deprecated, and the long term roadmap of nvim-treesitter is to remove the module system and stop serving as a dependency for plugins, so a change like this will be needed _eventually_ anyway.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -7,9 +7,6 @@ local api = vim.api
 local uv = vim.loop
 local fn = vim.fn
 
--- is treesitter available?
-local __has_ts, __ts_configs, __ts_parsers
-
 local Previewer = {}
 
 Previewer.base = Object:extend()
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @gpanders! I can’t check ATM as traveling, which version added this API?

gpanders:
`vim.treesitter.start()` was added in 0.9.

If you'd like to support versions older than that, you could use

```lua
local ok, parser = pcall(vim.treesitter.get_parser, bufnr, ft)
if not ok then
  return false
end

vim.treesitter.highlighter.new(parser)
return true
```

instead (this is exactly what `vim.treesitter.start` does).

ibhagwan:
Ty @gpanders, will test and merge this PR if works as expected.

gpanders:
It requires that the parser for a filetype be installed, but that is also true when using `nvim-treesitter`, is it not?

As a test, I changed it locally to this:

```lua
local ts_attach = function(bufnr, ft)
  local ok, err = pcall(vim.treesitter.start, bufnr, ft)
  assert(ok, err)
  return true
end
```

For filetypes which I have a parser (C, Lua), there is no error and highlighting works. For filetypes which I do not have a parser (e.g. YAML), I see an error. But this is exactly what I would expect. If I don't have a parser installed it should fallback to regex syntax highlighting.

ibhagwan:
> It requires that the parser for a filetype be installed, but that is also true when using nvim-treesitter, is it not?

That was my thought too, I have parser installed for all my files, for some odd reason the old code works but when using `vim.treesitter.start` it fails, since it works for you I’ll do some more digging.

gpanders:
I suggest keeping an eye on the `main` branch of `nvim-treesitter`, which will eventually become the default (or at least merged into `master`): https://github.com/nvim-treesitter/nvim-treesitter/tree/main

The code as-is will fail when using that branch because it removes `ft_to_lang`, which is why I looked into this issue and created this PR in the first place.

No immediate action is needed, but as I said it's worth keeping an eye on so that you can stay ahead of any breakages when that branch is eventually merged.

ibhagwan:
> I suggest keeping an eye on the `main` branch of `nvim-treesitter`, which will eventually become the default (or at least merged into `master`): https://github.com/nvim-treesitter/nvim-treesitter/tree/main
> 
> The code as-is will fail when using that branch because it removes `ft_to_lang`, which is why I looked into this issue and created this PR in the first place.
> 
> No immediate action is needed, but as I said it's worth keeping an eye on so that you can stay ahead of any breakages when that branch is eventually merged.

Agreed, we can leave this PR open until we figure out the alternative, we can also monitor Telescope as they'd be facing the same issue.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/74e8200a350e1e6a4f3b3d3dbc7c87fc7ddaa2df

@gpanders, merged manually with backward compatibility (did not remove the `0.8` code) and a couple more changes, you can't use the `ft` directly with `vim.treesitter.start` as it doesn't always match, you first have to call `vim.treesitter.language.get_lang(ft)` to acquire the right ts language name.

The new implementation is:
https://github.com/ibhagwan/fzf-lua/blob/74e8200a350e1e6a4f3b3d3dbc7c87fc7ddaa2df/lua/fzf-lua/previewer/builtin.lua#L691-L703

You still get credit for the entire commit btw :-)

PR #777, opened 2023-06-10
Fix "-.*" and "--.*" queries for st and mt versions of live_grep

deponian:
Hi!

Thank you for this awesome plugin!
I've read some of the past issues and I can't help but note your thoroughness and kindness with which you maintain this repository!

I stumbled upon a bug with `live_grep` command. It doesn't work for queries like `-info`, `--help` and other things that can be interpreted as command line option.

I saw that you use `utils.rg_escape()` to avoid these issues but unfortunately it doesn't work when user (like me) adds `--fixed-strings` options for `rg` and wants to find only exact matches.

I've fixed it using `rg's` `-e` flag for both single-thread and multi-thread cases of `live_grep`.

It still doesn't work for `grep` command, because I couldn't fix it without changing code too much.

Is there any reason you chose to use `utils.rg_escape()` instead of `-e` flag?

### DIFF
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -287,7 +287,7 @@ M.preprocess = function(opts)
       search_query = search_query:gsub("%%", "%%%%")
       -- reset argvz so it doesn't get replaced again below
       opts.cmd = opts.cmd:gsub(argvz,
-        glob_args .. vim.fn.shellescape(search_query))
+        glob_args .. "-e " .. vim.fn.shellescape(search_query))
     end
   end
 
@@ -297,7 +297,7 @@ M.preprocess = function(opts)
     opts.cmd = opts.cmd:gsub("{argv.*}",
       function(x)
         local idx = x:match("{argv(.*)}")
-        return vim.fn.shellescape(argv(idx))
+        return "-e " .. vim.fn.shellescape(argv(idx))
       end)
   end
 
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @deponian for the kind words!

This isn't a bug, this is "by design", you just have to manually escape regex special characters when you input the query, so for example if I wanna search for `--info` I would use the query `\-\-info`, this also works with the `live_grep_glob` variant:

> Searching for `--info` within the fzf-lua repository in all files that aren't `*.lua`:
image

Note you can also perform this search directly from the command line but you have to add `no_esc=true` in order to prevent the call to `rg_escape` (works for both `live_grep` and `grep`):
```lua
:lua require("fzf-lua").grep({ search=[[\-\-info -- !*.lua]], no_esc=true })
:lua require("fzf-lua").live_grep_glob({ search=[[\-\-info -- !*.lua]], no_esc=true })
```

ibhagwan:
> Is there any reason you chose to use `utils.rg_escape()` instead of `-e` flag?

There are a few reasons for that, although `rg` is the preferred grepper this plugin also works with `grep` or `ag`, not all support the same options and work the same, not using any hardcoded opttions gives the users the flexibility to use any command they wish using the `cmd` argument, e.g.:
```lua
:lua require("fzf-lua").grep({ cmd = "ag ..."  })
```

In addition, this also enables the users to search for regex patterns directly (using `no_esc`) and not rely on crutches or auto-escaping of certain characters, for exmaple, searching for a function that has `require` in name:
```lua
:lua require("fzf-lua").grep({ search=[[require.*\(.*\)]], no_esc=true })
```

deponian:
Thank you for clarification! Sorry for bothering you.

ibhagwan:
> Thank you for clarification! Sorry for bothering you.

No worries @deponian, feel free to open an issue if you have addittional questions.

Issue #343, opened 2022-02-11
Oldfiles takes long to open


shyun3:
When I enter `:FzfLua oldfiles`, it can take about a second or two to open. But `:o` pulls up the old files immediately. There are about 2400 files in there right now. On another system, there are about 170 old files and both `:o` and `:FzfLua oldfiles` open immediately. Could the performance of this be improved?

My `shada` setting in Neovim is `!,'80000,<50,s10,h`. I am also using the `include_current_session` option for fzf-lua oldfiles.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/8591f3c8464dd1baaf801c5e8acf6d663c56cf6d this commit should improve the speed significantly.

`oldfiles` was one of the first providers I implemented, TBH without much thought for performance, with the latest changes the interface is async and sohuld be more responsive, additioanlly I disabled the filesystem test for files (was probably part of the issue) unless the user specifically requests by passing `stat_file = true`:
```lua
-- only send this if you wish to remove files that no
-- longer exist on disk as this can be quite costly
:lua require'fzf-lua'.oldfiles({ stat_file=true })
```

Try it out and let me know if this solves your issue?

ibhagwan:
Alright, I think I've got this done properly now, latest commit makes sure the interface opens regardless of the amount of files, just tested it with artficially inflating  my oldfiles up to 1M files, the UI always stays responsive as results are pouring in, took about 7 seconds to finish with `stat_file=false` and roughly double the time with `stat_file=true` (which is the default).

If you're still having issues feel free to reopen this.

shyun3:
This is much more responsive, thanks!

ibhagwan:
Great! tysm for the update :-)

PR #790, opened 2023-06-18
wip: reducing the times of invoking fs_stat in oldfiles

haolian9:
hi, while reading providers/oldfiles.lua, i begin to wonder if we can reduce the times of invoking fs_stat.

and the very first question always is: is it a real problem? and it's worth it?
honestly i'm not sure, i am just having fun. well, maybe yes. the default length of v:oldfiles is 100, and these files might be stored in different disks: ssd, hdd, usb stick.

at the very begin, and just like [this commit](https://github.com/ibhagwan/fzf-lua/commit/9b418128d7a0ebc2a740b11938c69f75eabb9c3a), i thought we can cache the results of fs_stat.

as i dig deeper, i figured some other approaches, bases on several facts:
1. `:h 'oldfiles' -> `The List can be modified, but this has no effect on what is stored in the |shada| file later.`
2. `:h :oldfiles` -> `This list is read on startup and only changes afterwards with \`:rshada!\`.`

with `1`, we process v:oldfiles once, then store the results to v:oldfiles for later use, and add a special mark to it for checking if it has been changed by others. while this could not be acceptable to users.

with `2`, we process v:oldfiles once, and store it in providers.oldfiles module, and invalid it when `:rshada`. in which cmdlineleave and cmdwinleave events and `quote:` can be used to watch if :rshada is being ran.

thanks for your reading, i hope these texts have not wasted mush time from you, and them did have some valid points.

---

after all this, i would just delegate these concerns to end users and let them offer a source of provider.oldfiles, haha.

### DIFF
--- a/lua/fzf-lua/providers/oldfiles.lua
+++ b/lua/fzf-lua/providers/oldfiles.lua
@@ -4,6 +4,34 @@ local make_entry = require "fzf-lua.make_entry"
 
 local M = {}
 
+local file_existence_checker
+do
+  ---@type {[string]: boolean}
+  local facts = {}
+
+  local function check(fpath)
+    -- todo: what if fpath is not absolute?
+    -- todo: what if the filesystem changes either inside or outside nvim?
+    if facts[fpath] == nil then
+      local stat, _, err = vim.loop.fs_stat(fpath)
+      facts[fpath] = not (stat == nil and err == "ENOENT")
+    end
+    return facts[fpath]
+  end
+
+  ---@diagnostic disable-next-line: unused-local
+  local function nocheck(fpath)
+    return true
+  end
+
+  ---@param do_stat boolean
+  ---@return fun(fpath: string): boolean
+  function file_existence_checker(do_stat)
+    do_stat = do_stat or true
+    return do_stat and check or nocheck
+  end
+end
+
 M.oldfiles = function(opts)
   opts = config.normalize_opts(opts, config.globals.oldfiles)
   if not opts then return end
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @haolian9.

> and the very first question always is: is it a real problem? and it's worth it?

This is the real question, IMHO, it’s never been a problem, certainly not for the default of 100 files.

I did test this with up to 100k entries, you can see the remnants of my debug prints before and after:
https://github.com/ibhagwan/fzf-lua/blob/b58799714513083bd646f0b716e29f224ebba48e/lua/fzf-lua/providers/oldfiles.lua#L53-L60

When originally improving performance of `oldfiles` in #343, I tested up to 1M entries, although slower, even 1M wasn’t an issue as fzf is being async fed (using the `coroutine`) and while I’m usually very open to performance enhancements this does come with the caveat of having to manage the cache as old files are never cleared, if someone decides to delete/rename files the cache needs to be updated which gets us back to square 1 (imagine someone “lives” inside neovim, they’re bound to make fs modifications).

Although probably not a huge issue I also dislike having unnecessary memory hogs inside the plug-in, which is why I went with a circular cache for the shell functions ID cache:

https://github.com/ibhagwan/fzf-lua/blob/b58799714513083bd646f0b716e29f224ebba48e/lua/fzf-lua/shell.lua#L9-L25

Bottom line, I would be against adding such a change as this has the potential of being a “voodoo bug” one day after I’ve forgotten we added this “forever cache”.

Much appreciate the effort!

PR #791, opened 2023-06-21
fix: bad unicode in UI

### LINKED ISSUE #672: [BUG] Unicode `REPLACEMENT CHARACTER`(efbfbd) in Buffers
### Info

- Operating System: MacOS
- Shell: zsh
- Terminal: Iterm2
- `nvim --version`: v0.8.2
- `fzf --version`: 0.37.0 (brew)

<details>

```lua
require('fzf-lua').setup({
    winopts = {
      height = 0.85, -- window height
      width = 0.85, -- window width
      preview = {
        horizontal = "right:50%",
        scrollbar = false,
      },
    },
    files = {
      prompt = "Files> ",
      rg_opts = "--color=never --files --follow -g '!.git'",
      fd_opts = "--color=never --type f --follow --exclude .git",
    },
    grep = {
      prompt = "Rg> ",
      input_prompt = "Grep For> ",
    },

})
```
</details>

### Description 

Sometimes it comes, sometimes it doesn't.

liangxianzhe:
│ -> |. Notice the difference?

#672.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -470,7 +470,7 @@ M.goto_jump = function(selected, opts)
 end
 
 M.keymap_apply = function(selected)
-  local key = selected[1]:match("[│]%s+(.*)%s+[|]")
+  local key = selected[1]:match("[|]]%s+(.*)%s+[|]")
   vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(key, true, false, true), "t", true)
 end
 
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -470,7 +470,7 @@ M.goto_jump = function(selected, opts)
 end
 
 M.keymap_apply = function(selected)
-  local key = selected[1]:match("[│]%s+(.*)%s+[│]")
+  local key = selected[1]:match("[│]%s+(.*)%s+[|]")
   vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(key, true, false, true), "t", true)
 end
 
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1305,8 +1305,8 @@ function Previewer.autocmds:populate_preview_buf(entry_str)
   self._is_vimL_command = false
   if entry.path == "<none>" then
     self._is_vimL_command = true
-    entry.path = entry_str:match("[^:]+│")
-    local viml = entry_str:match("[^│]+$")
+    entry.path = entry_str:match("[^:]+|")
+    local viml = entry_str:match("[^|]+$")
     local lines = vim.split(viml, "\n")
     local tmpbuf = self:get_tmp_buffer()
     vim.api.nvim_buf_set_lines(tmpbuf, 0, -1, false, lines)
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -301,7 +301,7 @@ M.keymaps = function(opts)
     if type(keymap.rhs) == "string" and #keymap.rhs == 0 then
       return
     end
-    keymap.str = string.format("%s │ %-40s │ %s",
+    keymap.str = string.format("%s | %-40s | %s",
       utils.ansi_codes[modes[keymap.mode] or "blue"](keymap.mode),
       keymap.lhs:gsub("%s", "<Space>"),
       keymap_desc or "")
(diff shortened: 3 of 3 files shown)


ibhagwan:
Ty for the PR @liangxianzhe, but I don’t think it makes sense to workaround this issue by changing the Unicode chars, if fzf has an issue with these characters we should find the exact issue and it should be fixed upstream.

More so, this doesn’t explain why this happens in `buffers` which doesn’t use the pipe character.

liangxianzhe:
Hi, I cannot repro this in fzf. I think the issue is more likely due to unicode doesn't pass through the pipes very well between lua and fzf, so more like a neovim issue. If file names contain weird unicode, it could still occur for `buffers`, but it is hard to get a consistent repro. This PR at least improves current situation.

liangxianzhe:
FYI here is a smaller repro, press the key and enter a few "a" then delete them. There are some randomness where the issue occurs though.
```lua
map("n", "fp", function()
  require('fzf-lua').fzf_exec({ [[
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
aaa │ aaa
  ]] })
end, { desc = "Test" })
```

Passing the same string to fzf cli works fine.

liangxianzhe:
Actually sounds like https://github.com/neovim/neovim/issues/16245

Even so, I would suggest to get this PR in, to avoid future mistakes parsing this.

ibhagwan:
> Actually sounds like [neovim/neovim#16245](https://github.com/neovim/neovim/issues/16245)
> 
> Even so, I would suggest to get this PR in, to avoid future mistakes parsing this.

Great find @liangxianzhe! this does seem like the issue.

I’m not still not sure if this PR is the right choice, it’s gonna look a bit different for users as the pipe won’t connect in between lines and is the main reason the character was chosen.

I’m also unsure if this solves it completely, I don’t have any pipe chars on filenames and still experienced it with `buffers` so this may not cover everything, it’s just a workaround.

liangxianzhe:
@ibhagwan The neovim issue applies to a few other unicode too. I suspect the `nbsp` could cause the issue for buffer.

Not sure when the neovim issue will be fixed. Before that, it is up to you to choose between bad unicode vs less nicer bar. I'm fine either way since I use my local fzf-lua.

ibhagwan:
> @ibhagwan The neovim issue applies to a few other unicode too. I suspect the `nbsp` could cause the issue for buffer.

Nbsp night be an issue too, given the issue (I suspect anything past `0x2000`) it used to be `A0` but some users had filenames with it which created issues, you can configure it via setup (from the README):
```lua
require'fzf-lua'.setup {
  -- uncomment if your terminal/font does not support unicode character
  -- 'EN SPACE' (U+2002), the below sets it to 'NBSP' (U+00A0) instead
  nbsp = '\xc2\xa0',
}
```

> Not sure when the neovim issue will be fixed. Before that, it is up to you to choose between bad unicode vs less nicer bar. I'm fine either way since I use my local fzf-lua.

It does seem to be a long running issue, I’m not sure what to do about this yet, I’m also unsure why the issue stopped occurring on my system, perhaps there’s a better way to avoid this, needs some more research.

nyngwang:
> I’m also unsure if this solves it completely, I don’t have any pipe chars on filenames and still experienced it with buffers so this may not cover everything, it’s just a workaround.

I also don't have the pipe chars(proposed in this PR) in my case `:FzfLua tabs`:

- I found that changing the terminal font doesn't help (I have tried four fonts locally, all with nerd-font patched)



- It's **NOT** something that randomly appears, it always happens if it can happen (see my scrolling demo below).

https://github.com/ibhagwan/fzf-lua/assets/24765272/ad9994dc-72fa-41da-b602-61e092803eef

ibhagwan:
Ty @nyngwang, that’s very helpful, does seem to match https://github.com/neovim/neovim/issues/16245 exactly, also reinforces my thinking that this PR won’t help.

Unfortunately I don’t think we can so much here other than wait for https://github.com/neovim/neovim/issues/16245 to be fixed.

@liangxianzhe, closing this PR, ty for clearly identifying the issue.

liangxianzhe:
@nyngwang Your promp character for tabs are unicode. Try change it back to ">" and it might fix your issue. Using less unicode consistently fix the repro on my side.

liangxianzhe:
Actually all the file icons are unicodes, which could cause the issue too.

ibhagwan:
> Does Neovim/fzf(I'm not sure who controls this) provide some API to "refresh" the terminal buffer every time a user scrolls it (and when it first starts)?

@nyngwang, I don’t know of anything like that, its possible that `:redraw` will “fix” this (try to resize the neovim window and see if that removes the number characters) but that would require triggering a redraw with every scroll/preview which is a bit heavy handed and far from a solution.

I genuinely think our efforts should be focused in expediting the neovim issue rather than trying to force fzf-lua to work around what is **clearly a neovim bug**.

nyngwang:
@liangxianzhe I have removed all icons and got this:

Issue #796, opened 2023-06-25
Extraneous preview window title highlights


0xcharly:
### Info

- Operating System: macOS Ventura 13.4
- Shell: Fish
- Terminal: Kitty
- `nvim --version`: `NVIM v0.10.0-dev-569+gfa0a25dcb`
- `fzf --version`: `0.42.0 (brew)`

```lua
require('fzf-lua').setup({
  winopts = {
    preview = {
      scrollbar = false,
    },
  },
})
```

### Description 

Title highlights are not cleared when `winopts.preview.scrollbar = false`, resulting in highlight artifacts that erroneously apply to the preview window border instead, as demonstrated in the following screenshot:

<img width="2179" alt="Screenshot 2023-06-25 at 12 32 34 PM" src="https://github.com/ibhagwan/fzf-lua/assets/5804569/78f9ccd9-79a3-4d42-a011-d220d3b3a30f">

0xcharly:
Tentative fix in https://github.com/ibhagwan/fzf-lua/pull/797.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/7aff4a3a089c65c6e60646945d8a5c8116078061

@0xcharly  - should solve your problem, lmk if otherwise.

ibhagwan:
@0xcharly if you don’t mind me asking, what settings you used to get the look and feel in your screenshot?

I believe @RayZ0rr was after something similar in #760 :)

0xcharly:
I've replied on #760 directly. Let me know if you want me to make it a profile.

ibhagwan:
> I've replied on #760 directly. Let me know if you want me to make it a profile.

It does look great, would love to accept a PR with your profile, I’m just not sure how to craft this without the theme overrides, maybe with a custom `winopts.border` using “fat” characters?

PR #797, opened 2023-06-25
fix(win): extraneous preview window title highlights

0xcharly:
When `winopts.preview.scrollbar` is `false`, the highlights of the preview window's title are not cleared, therefore overlapping with each other when navigating up and down the list of results.

This commit moves the code responsible for clearing the highlights into `FzfWin:set_title_hl()` to avoid leftover highlight artifacts.

Fixes https://github.com/ibhagwan/fzf-lua/issues/796.

| Before | Screenshot 2023-06-25 at 12 32 34 PM |
| --- | --- |
| After | Screenshot 2023-06-25 at 12 33 31 PM |

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -917,19 +917,17 @@ function FzfWin.win_leave()
   _self:close()
 end
 
-function FzfWin:clear_border_highlights()
-  if self.border_winid and vim.api.nvim_win_is_valid(self.border_winid) then
-    vim.fn.clearmatches(self.border_winid)
-  end
-end
-
 function FzfWin:set_title_hl()
-  if self.winopts.__hl.title and self._title_len and self._title_len > 0 then
-    pcall(vim.api.nvim_win_call, self.border_winid, function()
-      fn.matchaddpos(self.winopts.__hl.title, { { 1, self._title_position, self._title_len + 1 } },
-        11)
-    end)
-  end
+  pcall(vim.api.nvim_win_call, self.border_winid, function()
+    -- Clear all highlights before adding new ones.
+    fn.clearmatches()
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for raising this issue @0xcharly, the border code is a bit of a mess, although this PR fixes the issue in the OP it does seem to mess up the hls when `scrollbar = "border"`.

I will refactor the border code so it makes more logical sense and solve this issue along the way.

In any event I intend to retire this code soon and use the new`nvim_open_win` properties `title_xxx` (same as the main window).

0xcharly:
Thanks for the quick answer, and apologies for the inadequate fix.

I'm happy to have a look a it, if you want, though I'm not sure I understand/can reproduce the problem you're referring to. With and without this commit, the `border` color seems to ignore `FzfLuaScrollBorderEmpty` and `FzfLuaScrollBorderFull`, whereas `float` respects `FzfLuaScrollFloatEmpty` and `FzfLuaScrollFloatFull`.

So I'm not seeing a change of behavior here, but I'm probably not looking at the right highlight. Do let me know if I'm missing something.

Either way, I understand if you'd rather not invest too much time and effort into this now if you're planning on refactoring this entirely soon. I'm happy to use `scrollbar = "float"` in the meantime.

ibhagwan:
Not your fault @0xcharly, as part of me testing the PR I discovered a few more issues I’d like to address, one of which is `FzfLuaScrollBorderFull` being ignored, another is border scroll bar being drawn when the file content is smaller than the number of lines which doesn’t require a border to be drawn.

I’ll make sure it all gets addressed and the function calls make logical sense.

0xcharly:
Gotcha, thanks for confirming and looking into it.

Cheers

PR #872, opened 2023-09-13
fix(completion): sort completion items for the :FzfLua command

wookayin:
Completion items for arguments to the `:FzfLua` command are now sorted in the alphabetical order. Previously it was given in a random order.


ibhagwan:
Ty for the contribution @wookayin!

I’ll test this briefly and merge.

PR #873, opened 2023-09-14
feat(input): Use asynchronous vim.ui.input with coroutine

wookayin:
This is an attempt continued from 6ee73fd, trying to use `vim.ui.input`
for asking user inputs via `fzf-lua.utils.input()`.

Previously in commit 6ee73fd (later reverted in 5d8dd50), the use of
`vim.ui.input` was incorrect, because it's an asynchronous function
but `utils.input()` is meant to be a synchronous function.
When dressing.nvim is used for `vim.ui.input`, `utils.input()` will
immediately return with `nil` before user enters text in the prompt;
it would result in the grep command being cancelled.

A possible fix to this would be to use a coroutine, which can "wrap"
an asynchronous function so that it can be used like a synchronous
function. To make it possible, command functions (e.g., FzfLua grep)
that uses `utils.input()` inside should be wrapped with coroutines.
Note that `utils.input()` will fall back to the default behavior
with the synchronous `vim.fn.input()` when coroutine is unavailable.

We enable the use of `vim.ui.input()` for prompting an user input
in neovim 0.9.0+ or higher, because the behavior of `vim.ui.input()`
on aborting the input by 'Ctrl-C' has been changed since neovim 0.9.0.

### DIFF
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -110,7 +110,7 @@ local get_grep_cmd = function(opts, search_query, no_esc)
   return command
 end
 
-M.grep = function(opts)
+M.grep = utils.wrap_coroutine(function(opts)
   opts = config.normalize_opts(opts, config.globals.grep)
   if not opts then return end
 
@@ -174,7 +174,7 @@ M.grep = function(opts)
   opts = core.set_header(opts, opts.headers or { "actions", "cwd", "search" })
   opts = core.set_fzf_field_index(opts)
   core.fzf_exec(contents, opts)
-end
+end)
 
 -- single threaded version
 M.live_grep_st = function(opts)
--- a/lua/fzf-lua/providers/tags.lua
+++ b/lua/fzf-lua/providers/tags.lua
@@ -182,7 +182,7 @@ M.btags = function(opts)
   return tags(opts)
 end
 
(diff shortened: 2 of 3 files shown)


wookayin:
This continues an attempt from #686 to use `vim.ui.input()` (or dressing.nvim). /cc @leiserfg

ibhagwan:
Hi @wookayin and ty for the contribution!

I like the approach, I thought about doing this before but I was reluctant to add yet another coroutine as fzf-lua already runs in a coroutine (and I wasn't sure what kind of woes I may encounter with nested coroutines):
https://github.com/ibhagwan/fzf-lua/blob/2a4d16f64bb2f9be6f471470eb1653a3b1d8be90/lua/fzf-lua/core.lua#L170

I guess the real question here is this worth the effort for 3 minor prompts (grep, tags and commit actions)?

wookayin:
had closed this accidentally. We might need to check more carefully whether a nested coroutine would work without any further problems. For its necessity -- yes, if one would want to use `dressing.nvim` for better prompt, so why not?

ibhagwan:
> if one would want to use dressing.nvim for better prompt, so why not?

As long as there’s no downside I’m ok with it.

I responded to specific lines of code in the commit with questions, do you see these?

ibhagwan:

PR #875, opened 2023-09-14
Fix two bugs (whitespace and multi-line desc) in keymaps

wookayin:
> fix(keymaps): Fix keymaps error when desc is a multi-line string

The textual description of a keymap can have more than one lines,
in which case the previewer was emitting errors. An easy fix here would
be to join all the lines (so that 1 keymap entry == 1 line in fzf).

BEFORE:

image

AFTER:

image

> fix(keymaps): Fix whitespace handling in keymap_apply action
Need to exclude trailing whitespaces to get the lhs keycode.

For example, an entry

    "n │ c                                        │ (desc)"

should be resolved to `c`, not `c                                    `.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -472,7 +472,8 @@ M.goto_jump = function(selected, opts)
 end
 
 M.keymap_apply = function(selected)
-  local key = selected[1]:match("[│]%s+(.*)%s+[│]")
+  -- extract lhs in the keymap. The lhs can't contain a whitespace.
+  local key = selected[1]:match("[│]%s+([^%s]*)%s+[│]")
   vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(key, true, false, true), "t", true)
 end
 
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -306,6 +306,8 @@ M.keymaps = function(opts)
     if type(keymap.rhs) == "string" and #keymap.rhs == 0 then
       return
     end
+    -- desc can be a multi-line string, normalize it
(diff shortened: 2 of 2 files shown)


wookayin:
BTW, although off-topic here, regarding the keymap_apply action, I'm not sure if executing an `imap` wouldn't make much sense. Also vmaps, xmaps, etc. Junegunn's original fzf.vim `Maps` command also shows normal-mode mappings (`n`).

ibhagwan:
Ty for these @wookayin, looks great, I’ll merge this after you remove the leftover debug line.

> BTW, although off-topic here, regarding the keymap_apply action, I'm not sure if executing an `imap` wouldn't make much sense.

Not sure I understand, right now the “apply” action just feeds the key sequence, what do you mean regarding the `imap`?

> Also vmaps, xmaps, etc. Junegunn's original fzf.vim Maps command also shows normal-mode mappings (n).

Aren’t you seeing normal mode mappings in the interface?

wookayin:
Oh wait, actualy this is wrong, rhs can contain whitespaces. Let me fix this soon
```
local key = selected[1]:match("[│]%s+([^%s]*)%s+[│]")
```

ibhagwan:
> Oh wait, actualy this is wrong, rhs can contain whitespaces. Let me fix this soon
> 
> ```
> local key = selected[1]:match("[│]%s+([^%s]*)%s+[│]")
> ```
> 
> e.g. `<cmd>lua require'fzf-lua'<CR>`

Need to also consider `<plug>` mappings.

wookayin:
My bad, it was about lhs. In vim keymaps lhs can't contain any whitespaces right?

ibhagwan:
> My bad, it was about lhs. I fixed that now. In vim keymaps lhs can't contain any whitespaces right?

Not as an actual space `%s+`, it would have to be defined as `<space>`.

PR #876, opened 2023-09-15
feat(keymaps) UI Improvement and more options for keymaps

duqcyxwd:
This PR aim to improve `keymaps`:
1. UI improvement, add fzf header line and show key mapp content
2. Add config options to support filter by mode. e.g: `lua require'fzf-lua'.keymaps({modes = {'n'}})`
3. Filter vim internal mapping that is not useful for end user, eg: `<Plug>`, `<SNR>` 

Example:
Before:
image

After:
image

### DIFF
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -291,6 +291,8 @@ M.keymaps = function(opts)
   opts = config.normalize_opts(opts, config.globals.keymaps)
   if not opts then return end
 
+  local formatter = opts.formatter or "%s │ %-14s │ %-33s | %s"
+  local key_modes = opts.modes or { "n", "i", "c", "v", "t" }
   local modes = {
     n = "blue",
     i = "red",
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for the contribution @duqcyxwd!

https://github.com/ibhagwan/fzf-lua/commit/bb704a36bc219c3e9aea2628968cafd4b50057e7

Manually merged (original autor is kept) with tiny fixes:
- The `|` char wasn't the correct one (it was a longer unicode char)
- Seperated `--header-line` into it's own key in `fzf_opts`
- Fixed the keymap previewer entry parser

wookayin:
This broke the multiline string issue again (described in #875 

image

ibhagwan:
> This broke the multiline string issue again (described in #875)
> 
> image
> when a keymap has desc, etc. so it has more than two lines in the `:map` (or `:imap`, etc.) output, e.g.:
> 
> ```
> i  -           * <Lua 788: some/path/to/module.lua:110>   
>                  autopairs end pair: 
>                  autopairs start pair: 
> ```

Oops, sorry @wookayin, I can fix it later when I’m near the PC but you’re also welcome to submit a new PR if you wish.

wookayin:
Thanks. Also I don't think we should exclude `<Plug>` mappings completely, because it's often useful to look at what implementation functions it's actually mapped to in a chain without leaving the UI.

e.g. `cs -> <Plug>Csurround -> :<C-U>call <SNR>102_changesurround()<CR>`

Perhaps excluding them by default would make a sense, but at least we could have some options as a configuration point.

ibhagwan:
> Perhaps excluding them by default would make a sense, but at least we could have some options as a configuration point.

Absolutely in favor of an include option, what do you think about `<SNR>` mappings?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/5fa59836c8df511dafd676c43a89ae4b47c4629a

@wookayin, this should fix the multiline string desc, also replaced the hardcoded `<SNR>` and `<Plug>` filtering with `ignore_patterns` (lua regex array, similar to `file_ignore_patterns`), the default being:
```lua
require("fzf-lua").setup({
  keymaps = {
    ignore_patterns = { "^<SNR>", "^<Plug>" }
  }
})
```

To disable:
```lua
:lua require("fzf-lua").keymaps({ ignore_patterns=false })
-- or
:FzfLua keymaps ignore_patterns=false
```

PR #877, opened 2023-09-17
fix(highlights): uppercase prompt

MariaSolOs:
A bit pedantic, but I think it makes sense to use the same casing as the other defaults :)


ibhagwan:
> A bit pedantic, but I think it makes sense to use the same casing as the other defaults :)

I can’t believe I missed this :)

Thank you again @MariaSolOs!

Issue #878, opened 2023-09-17
General insert mode completion (fzf-lua.complete)


wookayin:
This would be a feature request or a discussion thread. I'm trying to implement *custom* insert mode completion (#688) as in the original fzf.vim.

<details>
   <summary>Info (not quite relevant)</summary>

- Operating System: macOS
- Shell: zsh
- Terminal: iTerm2 
- `nvim --version`: nvim-0.9.2
- `fzf --version`: 0.39.0
</details>

### In fzf.vim ...

Previously I had
```vim
imap <C-x><C-l><c-l>   <plug>(fzf-complete-line-allfiles)
imap <expr> <plug>(fzf-complete-line-allfiles)   fzf#vim#complete(Fzf_completeline_rg('', ''))

function! Fzf_completeline_rg(pattern, rg_filetype)
  let l:pattern = empty(a:pattern) ? '""' : shellescape('\b' . a:pattern. '\b')
  let l:filetype = empty(a:rg_filetype) ? ' ' : printf('-t %s ', shellescape(a:rg_filetype))

  return fzf#wrap({
        \ 'source': 'rg --column --line-number --no-heading --color=always ' . l:filetype . ' ' . l:pattern,
        \ 'options': ['--delimiter', ':', '--ansi', '--nth', '3..', '--no-multi'],
        \ 'reducer': { lines -> trim(join(split(lines[0], ':')[3:], ':')) },
        \ })
endfunction
```

which works like `:FzfLua lines` but the default action being inserting the selected line at the cursor rather than jumping to the selected line.

### fzf-lua.complete

The module `fzf-lua.complete` provides some useful built-in completions, such as `M.line` (it's same as `bline` at the moment though), `M.bline`, `M.file`, etc. However for implementing custom insert-mode completions, I find it's quite complicated. Rather than providing the raw fzf command through [`fzf-lua.complete..fzf_complete`](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/complete.lua#L98C1-L98C15), I hope I can reuse existing providers with modified opts.

For example, `fzf-lua.complete.bline` is implemented like (see https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/complete.lua#L146)

```lua
M.bline = function(opts)
  opts = config.normalize_opts(opts, config.globals.complete_bline)
  opts = set_cmp_opts_line(opts)
  return require "fzf-lua.providers.buffers".blines(opts)
end
```
[...]

PR #879, opened 2023-09-17
fix(core): display errors that might happen during actions

wookayin:
Errors happened in `fn_selected` (typically actions upon selection)
are somehow not displayed. To prevent a silent failure, we catch errors
and notify users. See also #878 for more context.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -171,7 +171,15 @@ M.fzf_wrap = function(opts, contents, fn_selected)
     opts.fn_selected = opts.fn_selected or fn_selected
     local selected = M.fzf(contents, opts)
     if opts.fn_selected then
-      opts.fn_selected(selected, opts)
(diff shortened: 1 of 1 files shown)


ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/335f908b912504015fefeeccacb4c325029d2bfc

Ty @wookayin!  Merged manually (author is kept).

Replaced `vim.notify` with `utils.err` which now also uses `vim.notify`, this part of utils was created back on the nightly build of 0.5 and was using `echo`, it was about time to use `vim.notify` instead so fixed that as well: https://github.com/ibhagwan/fzf-lua/commit/51a4cc0c2ad509147d7c3ce6d115e03de1c985e4

wookayin:
Thanks. I see 51a4cc0c2ad509147d7c3ce6d115e03de1c985e4 has replaced utils.err implementation with `vim.notify`.

PR #883, opened 2023-09-24
fix!: Always shellescape fzf_opts["--header"] and don't override for history providers

wookayin:
### fix(core)!: always shellescape fzf_opts["--header"]

It was previously not clearly documented or required whether `fzf_opts["--header"]` must be shell-escaped or may take a form of raw strings.

**BREAKING CHANGE**: Like `fzf_opts["--prompt"]`, `fzf_opts["--header"]` will be automatically shell-escaped (for the sake of consistency) so normal string can be directly fed to `opts.fzf_opts["--header"]`.

Although this may not break any built-in commands, any user-custom opts that have a *shell-escaped* header fzf_opts may result in a wrong double escaping.

This commit also adds an unit test for this behavior on `--header`.

### fix(actions): Do not override header if it's already given in opts

This allows to override header for command_history and search_history.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,11 @@ PLENARY-REPO=https://github.com/nvim-lua/plenary.nvim.git
 
 .PHONY: test
 test:
-	nvim --headless --noplugin -u tests/minimal_init.vim -c "lua require('plenary.test_harness').test_directory('tests/', { minimal_init = 'tests/minimal_init.vim', sequential = true, timeout = 120000 })"
+	nvim --headless --clean --noplugin -u tests/minimal_init.vim -c "lua require('plenary.test_harness').test_directory('tests/', { minimal_init = 'tests/minimal_init.vim', sequential = true, timeout = 120000 })"
 
 .PHONY: test-file
 test-file:
-	nvim --headless --noplugin -u tests/minimal_init.vim -c "lua require('plenary.busted').run(vim.loop.cwd()..'/'..[[$(FILE)]])"
+	nvim --headless --clean --noplugin -u tests/minimal_init.vim -c "lua require('plenary.busted').run(vim.loop.cwd()..'/'..[[$(FILE)]])"
 
 
 .PHONY: plenary
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -438,6 +438,10 @@ M.build_fzf_cli = function(opts)
       opts.fzf_opts[flag] = opts.fzf_opts[flag]
     end
   end
+  -- shell escape the header. Note: '--header=', '--header', '' are all different!
+  if opts.fzf_opts["--header"] then
+    opts.fzf_opts["--header"] = libuv.shellescape(opts.fzf_opts["--header"])
+  end
   opts.fzf_opts["--bind"] = M.create_fzf_binds(opts.keymap.fzf)
   if opts.fzf_colors then
     opts.fzf_opts["--color"] = M.create_fzf_colors(opts)
@@ -516,8 +520,9 @@ M.build_fzf_cli = function(opts)
     end
     if v then
       v = v:gsub(k .. "=", "")
-      cli_args = cli_args ..
-          (" %s%s"):format(k, #v > 0 and "=" .. v or "")
+      -- assumes v are well normalized and shell-escaped
+      local value_repr = (#v > 0 and "=" .. v or "")
+      cli_args = cli_args .. (" %s%s"):format(k, value_repr)
     end
   end
   return cli_args .. extra_args
@@ -769,7 +774,7 @@ M.set_header = function(opts, hdr_tbl)
     end
   end
   if hdr_str and #hdr_str > 0 then
-    opts.fzf_opts["--header"] = libuv.shellescape(hdr_str)
+    opts.fzf_opts["--header"] = hdr_str   -- escape in build_fzf_cli
   end
   return opts
 end
--- a/lua/fzf-lua/providers/dap.lua
+++ b/lua/fzf-lua/providers/dap.lua
@@ -145,7 +145,7 @@ M.breakpoints = function(opts)
   end
 
   if opts.fzf_opts["--header"] == nil then
-    opts.fzf_opts["--header"] = vim.fn.shellescape((":: %s to delete a Breakpoint")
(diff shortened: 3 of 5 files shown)


wookayin:
You might not like a breaking change here, but I find the specification on `fzf_opts` really confusing. IMHO, every single entry in `fzf_opts` should be normal text rather than shell-escaped strings, and `core.build_fzf_cli` should take care of all the escaping hassles (except for raw flags such as `fzf_raw_args`).

If you decide to NOT introduce any breaking change for some reason (which I think is inevitable at some point though), we should at least document it and promise precisely what we would expect about `fzf_opts`. For now it has been only implicit, so I think it'd be OK to fix in the right way.

The fix for command_history and search_history is actually independent of this breaking change, so I can also have it as a separate PR if you'd like (or you can just easily cherry-pick it).

ibhagwan:
> You might not like a breaking change here, but I find the specification on fzf_opts really confusing. IMHO, every single entry in fzf_opts should be normal text rather than shell-escaped strings

You’re correct, I prefer to avoid breaking changes as much as possible, having avoided most of breaking changes until now potentially created the confusion.

I agree it makes more sense to have **everything** in fzf_opts she’ll escaped but that would:
(a) introduce a breaking change I’d like to avoid
(b) remove the flexibility of sending non shell escaped strings (for rebinds, preview actions, etc).

if you wish to continue with this PR it would have to be much more than just for headers, will have to take into account breaking changes either by working around them in the code or properly notifying users who would otherwise be affected.

I will not introduce something that breaks functionality by raising an exception and forcing the user to figure out what went wrong in the update and having to read docs, etc.

ibhagwan:
> The fix for command_history and search_history is actually independent of this breaking change, so I can also have it as a separate PR if you'd like (or you can just easily cherry-pick it).

If that’s your wish I can do that.

wookayin:
> I agree it makes more sense to have everything in fzf_opts shell escaped but that would:
(a) introduce a breaking change I’d like to avoid
(b) remove the flexibility of sending non shell escaped strings (for rebinds, preview actions, etc).

I think this is an interesting point to discuss. While I think **everything** in fzf_opts shouldn't be assumed to be shell-escaped, may I ask you for the rationale/reason why you think they would need to be escaped? In either way, we need to **be consistent** and well-document the behavior.

Let's use this thread for discussing, I'm fine with not merging this at this very moment (but please feel free to cherry-pick 5970fce).

ibhagwan:
> may I ask you for the rationale/reason why you think they would need to be escaped?

While I agree there is not always a need to escape I try to minimize the overhead for users and having to know that if the value contains spaces or special chars you need to shell escape requires reading the docs and understanding the underlying mechanics, that means understanding a binary named fzf is called and it has flags that are sent via the shell, etc, etc. You have to understand that most users aren’t sophisticated, some don’t even know what is fzf and how to use it, they know they have to install it and that’s it, most functionality comes out of the box without any need to customize.

My philosophy is such that I prefer most options (not just fzf_opts) to work in a straight forward manner that is intuitive and doesn’t always require reading docs, knowing you have to call `shellescape` is a step I’d like to save the users from.

Optimally this knowledge should be “hidden”, users can specify any value they want and it would **just work**, fzf-lua should be the one deciding if we need to escape or not and on top of that we should provide an option to not escape for advanced customization (similar to grep `no_esc = true`.

wookayin:
Oh I see, I just got your point wrong. I misunderstood it like everything that are passed to `fzf_opts` needs to be shell-escaped but actually what you meatn by "I agree it makes more sense to have everything in fzf_opts shell escaped" is just exactly the same way as I was thinking of --- we pass the un-escaped values and it's escaped in `build_fzf_cli`. Sorry for the mistake!

Yes I also 100% agree with you, when an advanced user would need to pass raw CLIs through `fzf_raw_args`, etc., they need to take care of shellescape because it's a raw flag.

> remove the flexibility of sending non shell escaped strings (for rebinds, preview actions, etc).

I think this can be dealt with `fzf_raw_args` (either in the form of string or table), so it'd be fine.

Overall I think it's more worth doing breaking change at some point to correct the wrong decision to make using fzf-lua easier. It makes all the options straightforward and intuitive. As you said, normal users without much customization would be fine as all the built-in features will work as-is, and those who have used `fzf_opts` with *manual* shell escapement are probably advanced users so they can also catch up with the upstream changes. The impact of this breaking change is not like an error, but only would be double escapement.

ibhagwan:
> Oh I see, I had gotten your point wrong

Replying on the mobile device isn’t always as clear as desirable :)

I have another suggestion, how about a new function `utils.needs_shellescape`? All fzf_opts would be run through this function and will only be escaped if needed (I.e not already escaped or a single value option like `--header-lines=1`?

This avoids the potentially breaking change and would also introduce the least amount of new potentially breaking code (some of the issues can be extremely voodoo, I wouldn’t discount wrapping `”1”` wouldn’t suddenly break some obscure edge case on fish shell lol).

ibhagwan:

Perhaps this one needs a new PR since the command history headers were fixed.

ibhagwan:
@wookayin, you'd be happy to hear this was addressed in https://github.com/ibhagwan/fzf-lua/commit/869367cab172cbedc4428bab330bb4e0ba2f5c3b (currently only available on the [windows branch](https://github.com/ibhagwan/fzf-lua/tree/windows)).

In addition, users can also use the "shortcut" option `{ header = ... }` in the same manner `prompt` and `query` are used.

None of the `fzf_opts` require escaping anymore, all are automaticlly escaped, for backward compat, the code also detects if any flag is already escaped and will alert the user to remove the escaping (while still working as expected):
https://github.com/ibhagwan/fzf-lua/blob/869367cab172cbedc4428bab330bb4e0ba2f5c3b/lua/fzf-lua/core.lua#L522-L525

Aside from escaping normalization being long overdue this was pretty much a must for windows. Due to its hellish escaping logic the users are no longer able to use `vim.fn.shellescape`.

For more info on windows support and the escaping logic, see https://github.com/ibhagwan/fzf-lua/pull/1001#issuecomment-1913446946.

ibhagwan:
@wookayin, FYI, with the windows branch (will merge sometimes next week), I made it so `fzf_opts` need not be escaped anymore, it is automatically escaped, you’re welcome to test it and lmk what you think.

PR #885, opened 2023-09-24
fix(actions): Do not override header if it's already given in opts

wookayin:
See #883. Separated from #883.


ibhagwan:
Ty @wookayin!

PR #888, opened 2023-09-26
chore: remove redundant bufname checks

MariaSolOs:
`nvim_buf_get_name` always returns a string, so checking that the result isn't falsy it's not needed :)

### DIFF
--- a/lua/fzf-lua/providers/files.lua
+++ b/lua/fzf-lua/providers/files.lua
@@ -43,7 +43,7 @@ M.files = function(opts)
   if not opts then return end
   if opts.ignore_current_file then
     local curbuf = vim.api.nvim_buf_get_name(0)
-    if type(curbuf) and #curbuf > 0 then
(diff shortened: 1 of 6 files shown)


lua/fzf-lua/providers/files.lua
```diff
@@ -43,7 +43,7 @@ M.files = function(opts)
   if not opts then return end
   if opts.ignore_current_file then
     local curbuf = vim.api.nvim_buf_get_name(0)
-    if type(curbuf) and #curbuf > 0 then
+    if #curbuf > 0 then
```
MariaSolOs, line 46: About this change: `type()` also always returns a string, so `type(curbuf)` is always true.

ibhagwan: Yah I have no idea why it's there, this can never be false, maybe I forgot the `== "string"` :-)

PR #889, opened 2023-09-28
fix(git_status): untracked files should be displayed correctly

wookayin:
Problem: The git status of untracked files in an (untracked)
directory were not correctly recognized as '?'.

Consider a scenario where
```
$ git status --porcelain=v1     # comment
A  TODO                         # staged
?? foo/                         # untracked
```

```
$ git status --porcelain=v1 -u
A  TODO
?? foo/1.md
?? foo/2.md
```

In this case, foo/1.md and foo/2.md were not recognized correctly as "?"
when included/listed in any normal `Files`-like command, or even in `GitFiles`
with a custom option is used to include untracked files as well, e.g.:

```
fzf_lua.git_files {
  cmd = "git ls-files --exclude-standard --cached --modified --others --deduplicate"
}
```

Solution: Use `-u` (`--untracked-files`) option for git status.

Note:
- Use git status --porcelain command, this is always a preferred
way to generate machine-readable git status.
- Also, the logic for git status parsing has been improved.

Refs: https://git-scm.com/docs/git-status

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -207,7 +207,7 @@ M.defaults.files = {
   cwd_prompt_shorten_len = 32,
   cwd_prompt_shorten_val = 1,
   fzf_opts               = { ["--info"] = "default", },
-  git_status_cmd         = { "git", "-c", "color.status=false", "status", "-s" },
+  git_status_cmd         = { "git", "status", "--porcelain=v1" },
   find_opts              = [[-type f -not -path '*/\.git/*' -printf '%P\n']],
   rg_opts                = "--color=never --files --hidden --follow -g '!.git'",
   fd_opts                = "--color=never --type f --hidden --follow --exclude .git",
@@ -231,9 +231,7 @@ M.defaults.git = {
   },
   status = {
     prompt      = "GitStatus> ",
-    -- override `color.status=always`, techincally not required
-    -- since we now also call `utils.strip_ansi_coloring` (#706)
-    cmd         = "git -c color.status=false status -su",
+    cmd         = "git status --porcelain -u",
     previewer   = "git_diff",
     file_icons  = true and M._has_devicons,
     color_icons = true,
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -223,15 +223,20 @@ M.get_diff_files = function(opts)
   if not cmd then return {} end
   local ok, status, err = pcall(utils.io_systemlist, path.git_cwd(cmd, opts))
   if ok and err == 0 then
+    -- Parse git status, expecting --porcelain=v1 format (or git status --short)
+    -- https://git-scm.com/docs/git-status#_short_format
     for i = 1, #status do
       local line = status[i]
-      local icon = line:match("[MUDARCT?]+")
-      local file = line:match("[^ ]*$")
+      local icon, file = line:match("^([ MTADRCU?][ MTADRCU?]) (.*)$")
+      -- Note: file that contains space might "be quoted"
       if icon and file then
         -- Extract first char, staged if not space or ? (32 or 63)
         local first = #line > 0 and string.byte(line, 1)
         local is_staged = first ~= 32 and first ~= 63 or nil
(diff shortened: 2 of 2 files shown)


wookayin:
Additional Note :

> Porcelain Format Version 1
>
> Version 1 porcelain format is similar to the short format, but is guaranteed not to change in a backwards-incompatible way between Git versions or based on user configuration. This makes it ideal for parsing by scripts. The description of the short format above also describes the porcelain format, with a few exceptions:

> - The user’s color.status configuration is not respected; color will always be off.
> - The user’s status.relativePaths configuration is not respected; paths shown will always be relative to the repository root.

(https://git-scm.com/docs/git-status#_short_format)

Also:

- Git status parsing might be slower if the repository has a *huge* number of untracked files. This might cause some performance issue. To optimize, one could combine the output of `git ls-files` and `git status` (without `-u`) and treat all the missing entry as `?`, but I think this is an overkill.

- Unfortunately `M.status.files.git_status_cmd` is not configurable. This value in `opt` will be ignored, and is used in the subprocess (not in the main thread).

- But I think actually we don't need all the commands and configs hyper-configurable. Git statuses are supposed to be in the exact same format, so it doesn't make a lot of sense to make it fully configurable. So I think it'd be better to hard-code in `make_entry`. What do you think?

ibhagwan:
> Git status parsing might be slower if the repository has a huge number of untracked files.

This is the main reason I didn’t do it this way, the performance impact is too high, since this command is run before the actual command (fd, rg, etc) this is also unexpected.

If you wish to see untracked you can use `git_status` instead, this is too much of a potential impact just for the files git icons.


> But I think actually we don't need all the commands and configs hyper-configurable. Git statuses are supposed to be in the exact same format, so it doesn't make a lot of sense to make it fully configurable. So I think it'd be better to hard-code in make_entry. What do you think?

You’re correct in the sense the customization here isn’t needed and was a bit of an overreach by me when writing the initial code (that’s why the customization isn’t working as I never needed to use this option) but perhaps we’ve found the purpose for this - I prefer to avoid this high impact change but I’ll accept this without changing the default and whoever is willing to take a risk with performance can change the default?

wookayin:
I think there's actually no reason to allow any kind of customization here for the git status command. If performance is the concern, I think doing directory matching (similar to prefix matching but aware of path segments) is the only way without the same default cmd being used (i.e. no `-u`).

ibhagwan:
> I think there's actually no reason to allow any kind of customization here for the git status command. If performance is the concern, I think doing directory matching (similar to prefix matching but aware of path segments) is the only way without the same default cmd being used (i.e. no `-u`). Let me work on that in my next free cycle.

If we’re not using `-u` I think it’s too much of an effort for just adding for icons to the files command. It’s not supposed to be a full on git_status replacement.

wookayin:
For the normal files command yes, maybe, but in my need I wanted to have a custom `:GitFiles!` which extends `:GitFiles` (or `:FzfLua git_files`) with the `--untracked-files` option. In this context, untracked files that are NOT displayed by are confusing and wrong as it feels like tracked, clean files.

An alternative option is to leave the default git_status parsing incorrect as-is, but allow overriding it runtime (https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/core.lua#L533-L551). But actually I more feel that we should remove this kind of hyperconfiguration; it's confusing because the option exists, so I'd prefer to writing some overkill logic to understand the git status output. It's non-trivial but not so difficult actually :P

ibhagwan:
Ty @wookayin! This looks good but I have a slight concern the new get_diff_info might be a bit costly as it essentially doubles the looks ups for each file, what do you think? Might be worth doing some perf testing here.

wookayin:
Yes, exactly, I also noted that this will be not performant. I will get back soon with some optimization and profiling results.

wookayin:
TODO: another caveat is that "git status --porcelain" will return falsy (non-zero exit code) if the repository is clean-clean.

ibhagwan:
> TODO: another caveat is that "git status --porcelain" will return falsy (non-zero exit code) if the repository is clean-clean.

What do you mean by "clean-clean", just tested it a repo a fresh git clone (i.e no changes at all) and both commands return with exit code `0`.

ibhagwan:

PR #891, opened 2023-09-29
Add marks_line mode

damianoognissanti:
Now "marks" will take you to the line and column of the mark, while "marks_line" will take you to the beginning of the line.


ibhagwan:
Ty @damianoognissanti!

TBH I think it makes more sense to just change it to go to the column, if extra customization is needed we can always add it as an option flag rather than adding yet another provider just for a slight change in the default action behavior.

PR #892, opened 2023-09-29
fix: marks should go to line and column

damianoognissanti:
Now "marks" will take you to the line and column of the mark, not just the start of the line.


damianoognissanti:
I made a new attempt :-)

ibhagwan:
Ty @damianoognissanti!

PR #894, opened 2023-09-29
fix: the previewer shows the wrong column for marks (off by one).

damianoognissanti:
The previewer shows the wrong column for marks (off by one). 
(This was the case even before #892)


damianoognissanti:
Wait! The wrong column is also shown in the previewer. Let me just fix that too...

PR #895, opened 2023-09-29
fix: the previewer shows the wrong column for marks

damianoognissanti:
The previewer shows the wrong column for marks (off by one).
(This was the case even before #892)


ibhagwan:
Ty @damianoognissanti!

PR #897, opened 2023-10-02
feat(git_stash): Allow `search` to search stash list for regex keyword

wookayin:
This is equivalent as `git stash list -G<regex>` in the command line.

Example:

    :lua require("fzf-lua").git_stash { search = 'pattern' }

    :FzfLua git_stash search=(some|regex|pattern)

This will be useful to quickly find the stash for some keyword **by content**. DEMO:

image

... when you have really many clutters on the stash list :P

### DIFF
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -1,5 +1,6 @@
 local core = require "fzf-lua.core"
 local path = require "fzf-lua.path"
+local libuv = require "fzf-lua.libuv"
 local utils = require "fzf-lua.utils"
 local config = require "fzf-lua.config"
 local shell = require "fzf-lua.shell"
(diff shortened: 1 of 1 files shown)


wookayin:
Idea: how to highlight the "grep string" in the preview window?

wookayin:
Another idea is to search through the "content" for a keyword:

Ref: https://stackoverflow.com/questions/42555410/is-it-possible-to-search-through-git-stash-items

```bash
for i in $(git stash list --format="%gd"); do
	git stash show -p $i | grep -H --label="$i" "$@"
done
```

Although it'd be quite easy to set this up with a custom user command, but if you'd more like this behavior (which is arguably more useful) I could also consider working again for such a "stash-grep" like behavior.

ibhagwan:
Ty @wookayin for the contribution!

> Idea: how to highlight the "grep string" in the preview window?

The current stash previewer is  using the native fzf previewer which limits us to shell command (currently `git --no-pager stash show --patch --color {1}`).

I haven't done proper research so it might be possible with a custom `git` command, another option that might be possilble is with  [git-delta](https://github.com/dandavison/delta) as pager (really nice for commits).

If you really want this there's another option which is to create a custom previewer that uses a neovim buffer (i.e. the "builtin" previewer), set it to display the stash as diff and then highlight whichever lines you want using `matchadd`, or `search`, etc

PR #911, opened 2023-10-22
Generate unique name for RPC server

oskarkook:
Using `vim.fn.serverstart()` can cause conflicts when different nvim instances have the same PID. This can happen if nvim runs sandboxed in its own PID namespace, for example when using Flatpak. The two different instances can have the same PID in their own namespace.

As noted in the docs, neovim will use the default path of `stdpath("run").."/{name}.{pid}.{counter}"`, however if the PID is the same, the RPC servers will clash.

To accomodate this use-case, the RPC server will be started with a reasonably unique name by using a timestamp in its name. This avoids name clashes.


ibhagwan:
> for example when using Flatpak. The two different instances can have the same PID in their own namespace.

Had no ideas this was possible, ty for the contribution @oskarkook!

PR #941, opened 2023-11-30
(complete-path): fix E5108, add support for "../path"

jqhr:
For example.
For relative path "../views", it will find for path related with current file. Regardless of whether the file is in the current directory or not。
84_1701345536_raw

### DIFF
--- a/lua/fzf-lua/complete.lua
+++ b/lua/fzf-lua/complete.lua
@@ -67,6 +67,20 @@ local function find_file_path(cursor_before_line)
   return nil
 end
 
+-- get search prompt
+local function get_prompt(filepath)
+  local cwd = vim.fn.getcwd()
+  -- if path include "-", replace it
+  local pattern = cwd:gsub("-", "")
+  local input_string = filepath:gsub("-", "")
+  local relative_path = input_string:gsub(pattern, "")
+  local prompt = relative_path:gsub(vim.env.HOME, "~")
+  if prompt:sub(1, 1) == "/" then
+    prompt = "~" .. prompt:sub(1)
+  end
+  return prompt
+end
+
 -- forward and reverse match spaces and single/double quotes
 -- and attepmpt to find the top level existing directory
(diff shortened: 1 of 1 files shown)


jqhr:
Some code reference from [cmp-path](https://github.com/hrsh7th/cmp-path/blob/main/lua/cmp_path/init.lua)

ibhagwan:
Tysm for this contribution @jqhr! Looks great, I’ll do some testing and merge.

Btw, there are still a few debug prints in the commit :)

ibhagwan:
So far a couple of issues with this PR:
1. it doesn't set the prompt to the correct working directory.
>before the PR completing `~/.config/nvim`:
image
>Aftet the PR:
image

3. Completion in normal mode doesn't set the cursor to the end of the path
>normal mode complete before the PR:
image
>After the PR (also note the botched path after completing with the cursor at the end of the path):
image

4. Noticed wrong cursor placement when completing from insert mode.

5. Issues completing when the cursor is on dot or slash (mangled paths as result)

jqhr:
Ok, I will fix those issues later

jqhr:
For promot and cursor position, normal mode search have beed fixed.
iShot_2023-11-16_21 25 52 mp4

ibhagwan:
@jqhr, https://github.com/ibhagwan/fzf-lua/commit/a1a2d0f42eaec400cc6918a8e898fc1f9c4dbc5f - this commit fixes the relative parent usecase with `../`, I'm still willing to accept improvements and fixes with this PR, just make sure this is tested properly and **has the same expected behavior prior to the PR**.

jqhr:
Oh. good, this commit looks better.
Actually, at first, I just want to solve `complete_path` could fix "../" problem, I don't familar with any feature about `complete_path`. My pr has many break changes. 
Now it's even better to have an official solution.
I will close those pr

ibhagwan:
Ty @jqhr!

PR #948, opened 2023-12-10
fix(lsp_code_actions): typo in preview message

MariaSolOs:
Super nit, but given that this message is displayed in the preview window of non-previewable code actions I couldn't resist :D


ibhagwan:
Oops, Ty @MariaSolOs!

Issue #800, opened 2023-06-27
bat previewer not found on linux


ramrom:
### Info

- Operating System: Ubuntu
- Shell: zsh
- Terminal: xterm
- `nvim --version`: 0.9.0
- `fzf --version`: 0.29

- [ ] The issue is reproducible with `mini.sh`

<details>
<summary>fzf-lua configuration</summary>

```lua
require('fzf-lua').setup({
  winopts = { preview = { default = "bat" } }
})
```
</details>

### Description 
The mini test uses your default previewer. this is only an issue with the bat previewer. an on linux.

On osx it works. 
on ubuntu, i see the error `sh: 1: bat: not found`

My guess is that it's because the binary is called `batcat` and not `bat`. it's called `bat` on osx.

(on linux i have alias for bat -> batcat)

ibhagwan:
The error message is self explanatory, the `sh` shell can’t find the executable `bat`, if `batcat` is your executable use setup to set `previewers.bat.cmd` to `batcat`.

ramrom:
aah dangit, I was scouring the README and i missed that part of it. 

Maybe the cmd should be a default? I guess not many people run linux.

thanks!

ibhagwan:
> I guess not many people run linux.

I'm on Linux :-)

![image](https://github.com/ibhagwan/fzf-lua/assets/59988195/571eb578-1580-4330-a8a8-32a23ae36b4f)

yao-weijie:
> The error message is self explanatory, the `sh` shell can’t find the executable `bat`, if `batcat` is your executable use setup to set `previewers.bat.cmd` to `batcat` on Linux:
> 
> ```lua
> require('fzf-lua').setup({
>   winopts = { preview = { default = "bat" } },
>   previewers = { bat = {
>     cmd = vim.loop.os_uname().sysname == "Darwin" and "bat" or "batcat"
>   } }
> })
> ```

Actually on Debian series distros, the bat cmd is `batcat`

```
❯ apt show bat
Package: bat
Version: 0.24.0-1
Built-Using: rust-ansi-colours (= 1.1.1-1), rust-onig-sys (= 69.8.1-1), rustc (= 1.70.0+dfsg1-1)
Priority: optional
Section: utils
Source: rust-bat
Maintainer: Debian Rust Maintainers <pkg-rust-maintainers@alioth-lists.debian.net>
Installed-Size: 5,299 kB
[...]

PR #960, opened 2023-12-18
fix(bat): adjust bat default cmd batcat/bat(#800)

yao-weijie:
I'm sorry I made a mistake that I forgot to use `vim.fn.executable()==1`,  this commit works well, please re-merge this commit instead.


ibhagwan:
Already fixed in https://github.com/ibhagwan/fzf-lua/commit/407323ade0ebca1bb4f26453c1aacbb246654139.

ty @yao-weijie!

Issue #957, opened 2023-12-17
Question: Using `:` as fzf delimiter is a problem on Windows, what would be the prefered curse of action? (Windows support PR)


TheLeoP:
### Description 

For similar reasons to the ones described in #956, using `:` as the fzf delimiter (or part of the fzf delimiter) it's a problem on Windows.

For example, the `FzfLua autocmds` uses `:` as a delimiter by default.

What would be the preffered curse of action? Would it be ok to simply change the delimiter?

TheLeoP:
As a reference, using `:` as a delimiter (the defautl at the moment), `FzfLua autocmds` looks like this:

![image](https://github.com/ibhagwan/fzf-lua/assets/53507599/b2b0c241-71f7-4bc8-9fc7-49242d651822)

and without any delimiters it looks like this:

![image](https://github.com/ibhagwan/fzf-lua/assets/53507599/f6c17c3b-f83c-4d71-a75d-4372ce8aadbc)

ibhagwan:
That’s a good question, not just `autocmds`, all line/column delimiters use `:` as delimiter (so basically all file type entries, buffers, grep, etc), I think there could be an easy solution for that which it to increase the fzf field index expression by 1 on windows (configurable), so for example on Linux/Mac line is always `{2}` on windows it will be `{3}`.

The alternative would be a full refactor and new approach to delimiters and also having to transform rg/grep output which IMHO would also incur a heavy performance cost.

TheLeoP:
> The alternative would be a full refactor and new approach to delimiters and also having to transform rg/grep output which IMHO would also incur a heavy performance cost.

I think this option is out of the picture.

> I think there could be an easy solution for that which it to increase the fzf field index expression by 1 on windows (configurable), so for example on Linux/Mac line is always {2} on windows it will be {3}.

Should this be changed in the appropiate default configs? Or inside the `set_fzf_field_index` function? I can see that the function isn't called in every provider that uses `:` as delimiter, should every provider that uses it call this function?

ibhagwan:
> Should this be changed in the appropiate default configs? Or inside the set_fzf_field_index function? I can see that the function isn't called in every provider that uses : as delimiter, should every provider that uses it call it?

I think perhaps we can use this opportunity to normalize the usage of the `:` delimiter and the field index separators and have a single source function that sets it including adjusting for windows, we can then call this function while setting config.defaults.

[...]

Issue #954, opened 2023-12-14
Would a PR trying to add windows support be welcomed?


TheLeoP:
### Description 

I'll like to try adding windows support based on the work already done by #772. I can't promise anything, but since I'm mostly using windows and I've heard such nice things about fzf-lua, I'd like to give it a try.

ibhagwan:
Go for it @TheLeoP, as long as it’s throughly tested it will be welcomed :)

GitMurf:
I see this Issue was closed. A couple questions:

1. does this imply Windows support is now available? From digging around I do not think so yet, but wanted to validate.
2. if no Windows support yet, @TheLeoP it looks like you are taking on the task of trying to get Windows working? It would be nice if you created a GH Issue to follow the development similar to this one from the past (that was dropped): #751 
3. if you created an open "discussion/progress" Issue for windows support (as suggested in Item 2) I would be happy to follow along and help provide testing as I am a Windows user very interested in this.

Appreciate the hard work!

ibhagwan:
> I see this Issue was closed. A couple questions:
> 
> 1. does this imply Windows support is now available? From digging around I do not think so yet, but wanted to validate.
> 2. if no Windows support yet, @TheLeoP it looks like you are taking on the task of trying to get Windows working? It would be nice if you created a GH Issue to follow the development similar to this one from the past (that was dropped): [Windows support ongoing questions #751](https://github.com/ibhagwan/fzf-lua/issues/751)
> 3. if you created an open "discussion/progress" Issue for windows support (as suggested in Item 2) I would be happy to follow along and help provide testing as I am a Windows user very interested in this.
> 
> Appreciate the hard work!

1. Correct, no windows support as of now.
2. A GH issue people can subscribe to get notifications is a good idea

TheLeoP:
@GitMurf just in case you didn't get pinged, #961 is tracking the progress of adding windows support

GitMurf:
@TheLeoP is this still being tracked? Did it land? It looks like the PR was closed without merge? Thanks in advance for any updates!

ibhagwan:
> @TheLeoP is this still being tracked? Did it land? It looks like the PR was closed without merge? Thanks in advance for any updates!

Windows support was added long ago @GitMurf lol

https://github.com/ibhagwan/fzf-lua/pull/961

https://github.com/ibhagwan/fzf-lua/commit/a5b32c30ece070d0417c2f41751fa11248c4e5de

[...]

PR #961, opened 2023-12-20
Add windows support [WIP]

TheLeoP:
This draft PR is intended to track the progress of adding windows support [as suggested](https://github.com/ibhagwan/fzf-lua/issues/954#issuecomment-1863871390) in #954.

@GitMurf 

TODO:
- [ ] Test escaping with every command (things like nested double quotes or passing lua strings through `cmd.exe` have proven to bee really tricky). The most complicated command yet has been `FzfLua live_grep` because `cmd.exe` loves eating double quotes.
    - [x] Check why `FzfLua live_grep` doesn't work with regexes (probably an escaping problem) and fix it. What does insert the `^`?
    - [x] Make `FzfLua live_grep` work for simple searches.
- [x] ~Handle separators as per #957 because windows loves `:`~ It looks like this isn't a path problem, but a `cmd` problem (again)
- [x] Handle separators as per #957 because windows loves `:` (for real this time). Solving an issue with colons in `make_entry.file` made the pickers act weird unless I manually increase the `--with-nth` value
- [ ] Do not only check if os is windows, check if the current shell is `cmd.exe` (low priority, can be done at the end).
- [ ] Check problems related with preview using shell commands
    - [ ] alternatives to git_diff using `sh -c`
- [x] Check live grep `exec_empty_query` conditioning
    - [x] Is this the same thing that makes `FzfLua live_grep` always fail before writing anything? Probably yes
- [ ] Add support for quickstart on Windows (?)
- [x] Why do commands like `FzfLua blines` or `FzfLua autocmds` show repeated entries?
- [ ] Use `vim.fs.normalize` on windows where appropiate
- [ ] Why do `ctrl-n` and `ctrl-p` do not work in buffers?


ibhagwan:
This is a great start @TheLeoP!

If you haven’t yet, I think it’s also worthwhile checking out fzf’s wiki regarding known limitations on windows and making sure we mention these in fzf-lua if/where applicable:
https://github.com/junegunn/fzf/wiki/Windows

TheLeoP:
@ibhagwan I hadn't found that info yet, thanks. Would this have to be added to the README, the neovim documentation, the wiki or all three?

ibhagwan:
> @ibhagwan I hadn't found that info yet, thanks. Would this have to be added to the README, the neovim documentation, the wiki or all three?

I think it’s worth a mention on the README, I linked this so we can explore what of these limitations affects fzf-lua in functionality in case we have to disable it by code.

TheLeoP:
I have a question regarding style. It seems like the current `stylua.toml` results in a style that gives warnings for lua-ls (since it has style cheking enabled). Is this intended?

ibhagwan:
> I have a question regarding style. It seems like the current `stylua.toml` results in a style that gives warnings for lua-ls (since it has style cheking enabled). Is this intended?

I have stylua configured as I tested it but I eventually settled on the formatted that comes with `lua_ls` and is configured in `.editorconfig`.

GitMurf:
@TheLeoP awesome! When you are ready for a windows user to test, let me know and I can test and try to help out in that way :)

GitMurf:
> - [ ] Test escaping with every command (things like nested double quotes or passing lua strings through `cmd.exe` have proven to bee really tricky). The most complicated command yet has been `FzfLua live_grep` because `cmd.exe` loves eating double quotes.

Curious have you tried with PowerShell instead of cmd?

ibhagwan:
@TheLeoP, given all the styling changes it’s a bit difficult for me to focus on the meaningful changes but from what I could gather it looks like amazing work so far! At one point when you feel the code is ready it might be worthwhile to merge it into a separate branch so we can work on it together and maybe even call for windows testers on Reddit.

TheLeoP:
@ibhagwan Sorry, I hadn't realized how much styling noice stylua caused while I had it turn on. I tried to reduce most of it.

> At one point when you feel the code is ready it might be worthwhile to merge it into a separate branch so we can work on it together and maybe even call for windows testers on Reddit.

Sounds great, I'll let you know when I think the PR is ready.

TheLeoP:
> Curious have you tried with PowerShell instead of cmd?

Yep. Escaping it's easier because powershell has both `'` and `"`, and sane esacaping rules. But, it's slower to start a powershell session than a `cmd` one and fzf internally uses `cmd` as shell on windows (at least by default).

@GitMurf

ibhagwan:
> @ibhagwan Sorry, I hadn't realized how much styling noice stylua caused while I had it turn on. I tried to reduce most of it.

No worries, I’ll handle it either way, if later we feel a different and more opinionated styling makes sense (such as stylua) I’m not against it.

> > At one point when you feel the code is ready it might be worthwhile to merge it into a separate branch so we can work on it together and maybe even call for windows testers on Reddit.
> 
> Sounds great, I'll let you know when I think the PR is ready.


Ty @TheLeoP, looking forward to it!

TheLeoP:
The string used in `--delimiter` is a regex, right? So, for example `["--delimiter"] = "'[:]'"`, is matching all characters inside the `[]`, in this case only `:`, right? Are the `'` intended to be part of the regex? Or is it because of a shell thing on posix shells?

I think that I was wrong in #957 because, apparently, removing the pair of `'` from the delimiter on Windows seems to solve the problem I was having.

ibhagwan:
> The string used in `--delimiter` is a regex, right? So, for example `["--delimiter"] = "'[:]'"`, is matching all characters inside the `[]`, in this case only `:`, right? Are the `'` intended to be part of the regex? Or is it because of a shell thing on posix shells?

Correct it’s a regex, from [man fzf] [-d](https://www.mankier.com/1/fzf#-d)
```man
-d --delimiter=STR
Field delimiter regex for --nth and --with-nth (default: AWK-style)
```

> Are the `'` intended to be part of the regex? Or is it because of a shell thing on posix shells?

I’ve experimented with different delimiters, if it’s just one char the single ticks can probably be removed but will have to be tested again.

> I think that I may have been wrong in #957 because, apparently, removing the pair of `'` from the delimiter on Windows seems to solve the problem I was having.

If it works on other platforms too by all means remove it.

TheLeoP:
> If it works on other platforms too by all means remove it.

Windows seems to require `"` instead of the `'`, for now I feel more confortable doing this change only on windows.

TheLeoP:
By the way, I'm realizing that some commands, like `blines` are showing duplicate entries. Do you have an idea of whay may be causing this?

Picture as reference:

image

ibhagwan:
> > If it works on other platforms too by all means remove it.
> 
> Windows seems to require `"` instead of the `'`, for now I feel more confortable doing this change only on windows.

Replacing `'` with `”` is a pretty safe change, in fact that’s what `libuv.shellescape` does for fish shell# (due to issues documented in the code).

ibhagwan:
> By the way, I'm realizing that some commands, like `blines` are showing duplicate entries. Do you have an idea of whay may be causing this?
> 

Strange indeed, no idea why this happens, perhaps something due to like 230 in fzf.lua:
```lua
handle_contents = vim.schedule_wrap(function()
```

Any specific reason you decided to `vim.schedule_wrap` the function?

TheLeoP:
> Strange indeed, no idea why this happens, perhaps something due to line 230 in fzf.lua:

I'll look into it tomorow.

> Any specific reason you decided to `vim.schedule_wrap` the function?

Since the Windows implementation uses libuv `pipe`s, the callbacks couldn't call neovim api functions and I was getting a lot of errors. But I'll also double check that tomorrow.

TheLeoP:
I couldn't resist taking a quick look and it seems that removing the `vim.schedule_wrap` does solve the problem (but calling API functions still throws errors)

ibhagwan:
> > Strange indeed, no idea why this happens, perhaps something due to line 230 in fzf.lua:
> 
> I'll look into it tomorow.
> 
> > Any specific reason you decided to `vim.schedule_wrap` the function?
> 
> Since the Windows implementation uses libuv `pipe`s, the callbacks couldn't call neovim api functions and I was getting a lot of errors. But I'll also double check that tomorrow.

I see, without windows this only affects some use-cases as mentioned here (regarding `vim.schedule`:
https://github.com/ibhagwan/fzf-lua/wiki/Advanced#lua-function-as-contents

ibhagwan:
> I couldn't resist taking a quick look and it seems that removing the `vim.schedule_wrap` does solve the problem (but calling API functions still throws errors)

Lucky guess. :)

Now that makes me even more curious why… why would scheduling something outside of vim’s “fast event” cause this duplication.

Perhaps there is another way to rid of the errors without scheduling? If you point out to me which calls are throwing the errors I can think of something.

TheLeoP:
> If you point out to me which calls are throwing the errors I can think of something.

Depends on the command used. In `FzfLua blines`, for example, its this line:

https://github.com/ibhagwan/fzf-lua/blob/407323ade0ebca1bb4f26453c1aacbb246654139/lua/fzf-lua/providers/buffers.lua#L25

ibhagwan:
> > If you point out to me which calls are throwing the errors I can think of something.
> 
> Depends on the command used. In `FzfLua blines`, for example, its this line:
> 
> https://github.com/ibhagwan/fzf-lua/blob/407323ade0ebca1bb4f26453c1aacbb246654139/lua/fzf-lua/providers/buffers.lua#L25

Got it, so basically every `vim.` call, anywhere, one solution would be to `vim.schedule` only the affected providers contents generation but that’s just a bandaid, anyone using fzf_exec on windows will face similar issues.

TheLeoP:
> Got it, so basically every `vim.` call, anywhere, one solution would be to `vim.schedule` only the affected providers contents generation but that’s just a bandaid, anyone using fzf_exec on windows will face similar issues.

Yeah, that's what I was thinking. Tomorrow I'll try to understand why scheduling this causes the lines to be duplicated and hopefully I'll be able to find a better solution. Thanks for your time c:

ibhagwan:
> > Got it, so basically every `vim.` call, anywhere, one solution would be to `vim.schedule` only the affected providers contents generation but that’s just a bandaid, anyone using fzf_exec on windows will face similar issues.
> 
> Yeah, that's what I was thinking. Tomorrow I'll try to understand why scheduling this causes the lines to be duplicated and hopefully I'll be able to find a better solution. Thanks for your time c:

Is it possible like 385 in fzf.lua should be moved up inside the if clause? Not sure about the flow yet but is it possible it’s called twice like this (once from the pipe callback and a second time at said line)?

TheLeoP:
> Is it possible line 385 in fzf.lua should be moved up inside the if clause? Not sure about the flow yet but is it possible it’s called twice like this (once from the pipe callback and a second time at said line)?

That was exactly the problem, thanks

ibhagwan:
> > Is it possible line 385 in fzf.lua should be moved up inside the if clause? Not sure about the flow yet but is it possible it’s called twice like this (once from the pipe callback and a second time at said line)?
> 
> That was exactly the problem, thanks

Great!

TheLeoP:
Is the `buffer` (picker? Telescope calls them pickers but I'm not sure what are they called here) supposed to do nothing for `ctrl-n` and `ctrl-p`?

ibhagwan:
> Is the `buffer` (picker? Telescope calls them pickers but I'm not sure what are they called here) supposed to do nothing for `ctrl-n` and `ctrl-p`?

There are no default binds setup for ctrl-p/n for any provider (picker), if you setup `- -history` these will be bind to next/prev history, otherwise these will be inherited from `FZF_DEFAULT_OPTS`.

TheLeoP:
In other pickers I'm able to go up and down the result list using `ctrl-n` and `ctrl-p`, but not in `buffers`

ibhagwan:
> In other pickers I'm able to go up and down the result list using `ctrl-n` and `ctrl-p`, but not in `buffers`

Explore the come binds with debug=true,  I can go next/prev with ctrl-p/n in buffers.

ibhagwan:
@TheLeoP, it's probably wise to dedup and save in a static the `is_windows` var, I've noticed a few locations where the condition is wrong (missing the `== 1` part) which obviously causes issues on non-windows platforms:
image


Perhaps a static `__IS_WINDOWS` under `utils` below these lines?
https://github.com/ibhagwan/fzf-lua/blob/407323ade0ebca1bb4f26453c1aacbb246654139/lua/fzf-lua/utils.lua#L11-L14

ibhagwan:
Looks great @TheLeoP, the only file I wouldn’t change is `libuv.lua` since it’s loaded a stand-alone (to proxy shell cmdw) file it shouldn’t contain any requires to other files, while it does work due to the require hack I added it’s probably safer without a first line `require`.

TheLeoP:
> Looks great @TheLeoP, the only file I wouldn’t change is `libuv.lua` since it’s loaded a stand-alone (to proxy shell cmds) file it shouldn’t contain any requires to other files, while it does work due to the require hack I added it’s probably safer without a first line `require`.

Sorry, I made the change without thinking.

ibhagwan:
Ty @TheLeoP! Let me know when you feel it’s stable enough for me to start trading on Linux/Mac.

ibhagwan:
@TheLeoP, I wanted to start testing this, I'm also trying to get my hands on a windows VM so I can help you with this PR, in the meantime to make this easlier I carefully rebased on main and a few tiny modifications (reduced extra styling noise).

This way I can also add commits if I see stuff that needs fixing, the next commits you add would you be able to rebase and add them on top of the `windows` branch?

https://github.com/ibhagwan/fzf-lua/tree/windows

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/d94ca001c83ea04188a74eb21f447265c78ed493 - this commit adds a wrapperfor the `vim.fs.normalize` calls reducing some lines of code.

TheLeoP:
Just to be clear. Currently, all of the changes on this PR are already on the `windows` branch, right?. Should I close this PR and open a new one targetting the `windows` branch and continue from there?

ibhagwan:
> Just to be clear. Currently, all of the changes on this PR are already on the `windows` branch, right?. Should I close this PR and open a new one targetting the `windows` branch and continue from there?

Correct, all changes from this PR were rebase-merged into a single commit (under your authorship).

> Should I close this PR and open a new one targetting the windows branch and continue from there?

Preferably we can rebase this PR onto the new branch but if you find opening a new PR easier I have no problem with that.

ibhagwan:
@TheLeoP, something interesting (still WIP), I was able to retire the complex escaping logic you added for `cmd /c` quoting. unlike `sh -c`, with `cmd -c` you can keep adding arguments after the `fzf` command, for example in UNIX `sh -c 'fzf' --ansi` isn't valid, the entire command needs to be in inside the quotes.

However with `cmd` you can write the command as;
```cmd
cmd /c fzf --ansi --prompt "foo> "
```
Which means we can pass the arguments as table elements directly to `jobstart|termopen`, this is some rough WIP and lots of things still fail (previews, etc), but I got the command `:FzfLua files previewer=false` with this commit **without calling the special windows escape function**: https://github.com/ibhagwan/fzf-lua/commit/9e198d6e49932434f9721b712de482991c6eb49e

TheLeoP:
> something interesting (still WIP), I was able to retire the complex escaping logic you added for cmd /c quoting

That's great, escaping was proving to be complex and this may be a more stable option in the long run.

Also, I'm sorry for not being able to contribute to the PR lately. Next week I'll try to put time into it. Thanks for all the help :3

ibhagwan:
> > something interesting (still WIP), I was able to retire the complex escaping logic you added for cmd /c quoting
> 
> That's great, escaping was proving to be complex and this may be a more stable option in the long run.
> 
> Also, I'm sorry for not being able to contribute to the PR lately. Next week I'll try to put time into it. Thanks for all the help :3

Ty @TheLeoP and no worries there is no rush on this, I’m also traveling this week so I will only be able to continue next weekend.

ibhagwan:
@TheLeoP, I think I’ve nailed it, it’s still not perfect but it seems we have to use a very specific logic for shell escaping on windows, we definitely can’t use `vim.fn.shellescape` on windows which simple doubles the double quotes.

The logic you created was in fact a patch to try and fix the bad quoting caused by using `vim.fn.shellescape`, with the new logic which I will link below all we have to do us simply make sure all the occurrences of `vim.fn.shellescape` are replaced with the adjusted (for windows) `libuv.shellescape`.

Instead of explaining again what I discovered I’ll just link the solution, it’s quite an interesting read about window' super retarded quote escaping/stripping logic:
https://stackoverflow.com/questions/6714165/powershell-stripping-double-quotes-from-command-line-arguments

And here is the implementation which I briefly tested and works with actions, even git_status stage/unstafe works:
https://github.com/ibhagwan/fzf-lua/blob/d4266bd71596a4463c47a55a3915ef958daaac1b/lua/fzf-lua/libuv.lua#L492-L546

TheLeoP:
I tried to force-push the windows branch into this PR, but it got closed automatically (?), so I'll open a new PR. Shold I continue with the same TODO list? I haven't take a deep look into the changed of the windows branch yet

ibhagwan:
> I tried to force-push the windows branch into this PR, but it got closed automatically (?), so I'll open a new PR. Shold I continue with the same TODO list? I haven't take a deep look into the changed of the windows branch yet

Maybe because I've been force pushing into the branch repeatedly as I'm working on it as we speak, now that I know you're on it I won't override the commits.

I will rebase-merge the branch so that it has 2 commits, one for my changes and another for yours and we can build on that, I believe that would be most straight forward.

> I haven't take a deep look into the changed of the windows branch yet

Quite a lot of changes, I rewrote the `path` module to be cross platform, while the changes are bigger I believe this will benefit us in future changes.

> Shold I continue with the same TODO list?

We should probably have a combined TODO list.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/76268b158f15957021c11d8f8750bdace0a03d53

@TheLeoP - this is the latest rebased commit in the windows branch, I won't overwrite it from now on, only add new changes.

Note the path module was rewritten and there are now tests under `test\path_spec.lua` which you can run if you have the `plenary.nvim` plugin installed with `:PlenaryBustedDirectory tests`

PR #976, opened 2023-12-31
fix: fzf-lua exit when there are hlgroups with attr fg/bg=fg/bg

Bekaboo:
Bug:    Some colorschemes (e.g. the builtin colorscheme 'torte') set
        fg/bg=fg/bg for hlgroups. If we pass --color=fg:fg to fzf, it
        will make fzf to exit immediately.

Fix:    Detect this in function `utils.hexcol_from_hl()`

To reproduce the bug, set nvim colorscheme to "torte": `:color torte` then open fzf-lua.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -445,6 +445,11 @@ end
 function M.hexcol_from_hl(hlgroup, what)
   if not hlgroup or not what then return end
   local hexcol = synIDattr(hlgroup, what)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for the fix @Bekaboo, I modified the PR a bit to cover other cases where the colorscheme is set to a color that has no corresponding value in the colormap.

https://github.com/ibhagwan/fzf-lua/commit/4f2ba97e7fd3429342963b8ae205e35e75d5ca16

Issue #599, opened 2023-01-07
How can I add/modify header?


fjchen7:
### Description 

I implement an action to switch oldfiles between `cwd_only=true` and `cwd_only=false`

```lua
{
  oldfiles = {
    actions = {
      ["tab"] = function(_, opts)
        opts.cwd_only = not opts.cwd_only
        fzf.oldfiles(opts)
      end
    }
  }
}
```

How can I add header about this keymap? Just like what live_grep does:
![image](https://user-images.githubusercontent.com/10106636/211144316-2107586c-d1a3-46db-99b5-50837c666475.png)

ibhagwan:
Use Fzf’s `header` option:
```lua
  oldfiles = {
    fzf_opts = { ["--header"] = "custom header" ]  }
    …
  }
```

fjchen7:
It doesn't work. Window dismissed immediately when calling `FzfLua oldfiles`.

ibhagwan:
> ```lua
> fzf_opts = { ["--header"] = "custom header" ]  }
> ```

My bad I answered from my phone and there's an extra `]`, on top of it I didn't mention this requires escaping if the line has spaces/special chars:
```lua
oldfiles = {
    fzf_opts = { ["--header"] = [["custom header"]]  }
    …
  }
```

You can also try it as a one liner from the `:` prompt:
```lua
:lua require("fzf-lua").oldfiles({ fzf_opts = { ["--header"] = [["custom header"]]  } })
```

fjchen7:
That work, thank! 

However I find that some commands still use default value even if I set `header`, like `files` or `git_files`. I thinks this problem may occurs for other fifth options. 

And if I set a value, I am not sure if the default one will be loaded. The override rule seems different for different config value.

Anyway this issue is solved. Thanks again.

ibhagwan:
> That work, thank!
> 
> However I find that some commands still use default value even if I set `header`, like `files` or `git_files`. I thinks this problem may occur for other fzf options.
> 
> And if I set a table config value, I am not sure if the default one will be loaded. The override rule seems different for different config value.
> 
> Anyway this issue is solved. Thanks again.

Some commands back default header lines, for example for files, grep the header is reserved for the `cwd` and  search query, this can still be overridden option `no_header = true`.

PR #977, opened 2024-01-01
feat: expose core.ACTION_DEFINITIONS to allow custom header sections

Bekaboo:
I want to add two custom actions `switch_provider` and `switch_cwd` and have them shown in the fzf header like what the `live_grep` or the `git_status` provider has:

```lua
---Switch provider while preserving the last query and cwd
---@return nil
local function switch_provider()
  local opts = {
    query = fzf.config.__resume_data.last_query,
    cwd = fzf.config.__resume_data.opts.cwd,
  }
  fzf.builtin({
    actions = {
      ['default'] = function(selected)
        fzf[selected[1]](opts)
      end,
      ['esc'] = function()
        fzf.resume(opts)
      end,
    },
  })
end

---Switch cwd while preserving the last query
---@return nil
local function switch_cwd()
  local opts = fzf.config.__resume_data.opts or {}

    -- Remove old fn_selected, else selected item will be opened
    -- with previous cwd
    opts.fn_selected = nil
    opts.cwd = opts.cwd or vim.uv.cwd()
    opts.query = fzf.config.__resume_data.last_query

    vim.ui.input({
      prompt = 'New cwd: ',
      default = opts.cwd,
      completion = 'dir',
    }, function(input)
      if not input then
        return
      end
      local stat = vim.uv.fs_stat(input)
      if not stat or not stat.type == 'directory' then
        print('\n')
        vim.notify(
          '[Fzf-lua] invalid path: ' .. input .. '\n',
          vim.log.levels.ERROR
        )
        return
      end
      opts.cwd = vim.fs.normalize(input)
    end)

    -- Adapted from fzf-lua `core.set_header()` function
    if opts.cwd_prompt then
      opts.prompt = vim.fn.fnamemodify(opts.cwd, ':~:.')
      local shorten_len = tonumber(opts.cwd_prompt_shorten_len)
      if shorten_len and #opts.prompt >= shorten_len then
        opts.prompt =
          path.shorten(opts.prompt, tonumber(opts.cwd_prompt_shorten_val) or 1)
      end
      if not path.ends_with_separator(opts.prompt) then
        opts.prompt = opts.prompt .. path.separator()
      end
    end

    fzf.resume(opts)
end
```

I searched through the the issue and found [this](https://github.com/ibhagwan/fzf-lua/issues/599) that provides a method to set custom headers, but as it says it does not change the header of some providers that already have their default header, and it is also kind of cumbersome if user wants to add a action header for multiple providers (e.g. the `switch_cwd` actions can be applied to all providers that inherits the `files` actions, including `files`, `git_files`, `git_status`, `grep`, `lsp`, `oldfiles`, `quickfix`, `loclist`, `tags`, `btags`, `args`)

After some searching and reading (using fzf-lua's live_grep :)) I found the default headers are defined in the `ACTION_DEFINITIONS` variable in `core.lua`. It will be much more convenient if we can expose it to users so that they can add custom actions to the table and fzf-lua will automaitcally pick it up.

For example, to add the `switch_cwd` custom action, one only need to

```lua
local fzf = require('fzf-lua')
local actions = require('fzf-lua.actions')
local core = require('fzf-lua.core')
local path = require('fzf-lua.path')

---Switch cwd while preserving the last query
---@return nil
function actions.switch_cwd()
  local opts = fzf.config.__resume_data.opts or {}

    -- Remove old fn_selected, else selected item will be opened
    -- with previous cwd
    opts.fn_selected = nil
    opts.cwd = opts.cwd or vim.uv.cwd()
    opts.query = fzf.config.__resume_data.last_query

    vim.ui.input({
      prompt = 'New cwd: ',
      default = opts.cwd,
      completion = 'dir',
    }, function(input)
      if not input then
        return
      end
      local stat = vim.uv.fs_stat(input)
      if not stat or not stat.type == 'directory' then
        print('\n')
        vim.notify(
          '[Fzf-lua] invalid path: ' .. input .. '\n',
          vim.log.levels.ERROR
        )
        return
      end
      opts.cwd = vim.fs.normalize(input)
    end)

    -- Adapted from fzf-lua `core.set_header()` function
    if opts.cwd_prompt then
      opts.prompt = vim.fn.fnamemodify(opts.cwd, ':~:.')
      local shorten_len = tonumber(opts.cwd_prompt_shorten_len)
      if shorten_len and #opts.prompt >= shorten_len then
        opts.prompt =
          path.shorten(opts.prompt, tonumber(opts.cwd_prompt_shorten_val) or 1)
      end
      if not path.ends_with_separator(opts.prompt) then
        opts.prompt = opts.prompt .. path.separator()
      end
    end

    fzf.resume(opts)
end

core.ACTION_DEFINITIONS[actions.switch_cwd] = { "Switch CWD" }

fzf.setup({
  actions = {
    files = {
      ['alt-c'] = actions.switch_cwd,
    },
  },
})
```

to have the custom header shown in all providers that inherits from `files`:

image

`files` provider

image

`git_status` provider

Also changing the default separator to "|" to align with the default separator of `git_status`.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -10,7 +10,7 @@ local make_entry = require "fzf-lua.make_entry"
 
 local M = {}
 
-local ACTION_DEFINITIONS = {
+M.ACTION_DEFINITIONS = {
   -- list of supported actions with labels to be displayed in the headers
   -- no pos implies an append to header array
   [actions.toggle_ignore]     = { "Disable .gitignore", fn_reload = "Respect .gitignore" },
@@ -748,7 +748,7 @@ M.set_header = function(opts, hdr_tbl)
       hdr_txt_str = "",
       val = function()
         if opts.no_header_i then return end
-        local defs = ACTION_DEFINITIONS
+        local defs = M.ACTION_DEFINITIONS
         local ret = {}
         for k, v in pairs(opts.actions) do
           local action = type(v) == "function" and v or type(v) == "table" and (v.fn or v[1])
@@ -768,7 +768,7 @@ M.set_header = function(opts, hdr_tbl)
             table.insert(t, i)
           end
           t[1] = (opts.header_prefix or ":: ") .. t[1]
-          return table.concat(t, opts.header_separator or "")
+          return table.concat(t, opts.header_separator or "|")
         end)() or nil
       end,
     },


ibhagwan:
I have no problem with these changes @Bekaboo, the only “issue” I have with it is that I don’t like this code and there needs to be a better interface for this but that’s on me :)

Your new functions look great!

ibhagwan:
Btw, personally I would’ve used another fzf-lua interface to choose the new directory, perhaps `files` with a custom command for `fd --type d`.

Bekaboo:
Thanks for accpeting this PR.

> Btw, personally I would’ve used another fzf-lua interface to choose the new directory, perhaps files with a custom command for fd --type d.

True, I just write this quickly and vim.ui.input() is simpler to use. I will definately try to replace it with some fzf functions when I get time :)

ibhagwan:
@Bekaboo, FYI `ACTION_DEFINITION` header set behavior changed with https://github.com/ibhagwan/fzf-lua/commit/718be95a88f6200137c17d64999e947afba07151, I can't guarantee it will stay like this forever but it's better than before, my plan is to remodel the actions/keybinds and the headers associated with them but I can't commit when it will be done.

Bekaboo:
@ibhagwan Thanks for letting me know!

ibhagwan:
> @ibhagwan Thanks for letting me know!

Hi @Bekaboo, FYI, there is another change incoming (probably the last in a while) that might affect the code you came up with, due to the work required to support windows I refactored but parts of the code, for example the path module was changed to support both windows and UNIX style paths (+added proper tests) and the core module changes the way it passes area to the headless wrapper via `mt_cmd_wrapper` it now encodes the option using base64 (to minimize escaping issues) and fzf_opts no longer require escaping any option and many more changes…

The new code is already published under [the windows](https://github.com/ibhagwan/fzf-lua/tree/windows) branch and so far it seems to be working great with no issues, I plan to merge the code sometime next week if no major issues are found.

You’re one of the few users (that I’m aware) who make use of fzf-lua internals (not just the API, which remained unchanged), I can updated you when I do that so you can pin the previous commit if you need more time to adjust.

As a side note, I also want to refactor the `ACTION_DEFINITIONS` as I dislike that part of the code, needs to be a more generic interface anyone can use.

Bekaboo:
@ibhagwan Thanks, I am happy to see that fzf-lua is being constantly improved!

ibhagwan:
> @ibhagwan Thanks, I am happy to see that fzf-lua is being constantly improved!

Ty @Bekaboo! Merged.

PR #979, opened 2024-01-02
feat: add new opt `ls_cmd` to buffer provider; small fix to bufnr alignment

Bekaboo:
## What?

This PR adds a new option `ls_cmd` to the buffer provider.

If `ls_cmd` is specified, buffer list will be get from the output of `:ls`
instead of `vim.api.nvim_list_bufs()`, for example:

```lua
require('fzf-lua').buffers({ ls_cmd = 'ls a+' })
```

will get buffer lists from the output of `ls a+` (active buffers with
modification).

Users can use `:ls[!] [flags]` command with arbitrary flags supported by
nvim, e.g.

- `{ ls_cmd = 'ls!' }` -> get list of "unlisted" buffers
- `{ ls_cmd = 'ls F' }` -> get list of terminal buffers with finished job,
  etc.

In addition, if `ls_cmd` is specified, buffer flags will be grepped directly
from the `ls` output. In this way, it can show more information than the current implementation, which only has '%' and '#' flags.

This PR also includes a small fix of the alignment issue of buffer numbers: in the function `gen_buffer_entry()` variable `max_bufnr_w` is set to `26 + #tostring(max_bufnr)`, I don't know where the number 26 comes from but seems like it's the length of the ANSI code in the bufnr. However if the user sets the `FzfLuaBufNr` highlight group to a different color, the ansicode length will be different or not exist at all, and the alignment will look weird:

**Before the fix:**

When `FzfLuaBufNr` is linked to 'Normal' (correct):

image

When `FzfLuaBufNr` is linked to 'Operator' (wider than normal)
image

When `FzfLuaBufNr` is cleared (way too wide):
image

**After fix:**

When `FzfLuaBufNr` is linked to 'Normal':

image

When `FzfLuaBufNr` is cleared:

image

When `FzfLuaBufNr` is linked to 'Operator'

image

## Why?

I want to use the fzf buffers to completely replace nvim's builtin command `:ls`/`:buffers`/`:files`, but found that it does not support some flags of these commands, see `:h :ls`. This PR gives us the ability to use those flags with fzf buffer provider.

Another benefit is that we now have a way to get more buffer flags from `:ls` instead of calculating them on our own to mimic the behavior of `:ls`

`require('fzf-lua').buffers({})` before this PR:

image

`require('fzf-lua').buffers({ ls_cmd = 'ls' })` after this PR:

image

Actual output of `:ls`:

image

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -425,6 +425,7 @@ M.defaults.buffers = {
   ignore_current_buffer = false,
   no_action_set_cursor  = true,
   cwd_only              = false,
+  ls_cmd                = nil,
   cwd                   = nil,
   fzf_opts              = { ["--tiebreak"] = "index", },
   _actions              = function() return M.globals.actions.buffers end,
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -59,8 +59,10 @@ end
 local populate_buffer_entries = function(opts, bufnrs, tabh)
   local buffers = {}
   for _, bufnr in ipairs(bufnrs) do
-    local flag = (bufnr == core.CTX().bufnr and "%")
-        or (bufnr == core.CTX().alt_bufnr and "#") or " "
+    local flag = (opts.buf_flags and opts.buf_flags[bufnr])
+        or (bufnr == core.CTX().bufnr and "%")
+        or (bufnr == core.CTX().alt_bufnr and "#")
+        or " "
 
     local element = {
       bufnr = bufnr,
@@ -94,9 +96,9 @@ local populate_buffer_entries = function(opts, bufnrs, tabh)
     -- DON'T FORCE ME TO UPDATE THIS HACK NEOVIM LOL
     local future = os.time({ year = 2100, month = 1, day = 1, hour = 0, minute = 00 })
     local get_unixtime = function(buf)
-      if buf.flag == "%" then
+      if buf.flag:match("%%") then
         return future
-      elseif buf.flag == "#" then
+      elseif buf.flag:match("#") then
         return future - 1
       else
         return buf.info.lastused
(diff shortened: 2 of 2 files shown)


ibhagwan:
Ty @Bekaboo for this!

Unfortunately I'd have to pass on this PR, this is a very rare edgecase and is way too confusing for most users, they have no idea what is a backend or how buffers are listed and which buffers are missing/hidden, I like to try to keep things simple and this isn't for very advanced neovim users. Moreso, I do not wish to maintain this adittional complexity (solve issues, etc).

> This PR also includes a small fix of the alignment issue of buffer numbers

If you wish to submit  a separate PR for this fix I'll be glad to accept it.

Bekaboo:
> they have no idea what is a backend

Sorry the backend staff is just from my previous config to switch providers, I prefer calling them backend. It is not related to this issue! Sorry for confusion.

Bekaboo:
Please consider accepting this patch after reading the changes. It is really not completed at all. Thanks in advance!

ibhagwan:
> > they have no idea what is a backend
> 
> Sorry the backend staff is just from my previous config to switch providers, I prefer calling them backend. It is not related to this issue! Sorry for confusion.

Even without the backend, I don't wish to support a new provider that adds new markers to the buffers list that create confusion, most users have no idea what these mean or ever looked at the output of `:ls a+`.

Moreso, this will also have to be ported to work for `tabs` as well (which also uses the same formatting, I'd prefer to avoid it altogether as you will probable be the only user ever to use this :-)

Bekaboo:
Feel free to close then. Thanks for your response

Bekaboo:
> Moreso, this will also have to be ported to work for tabs as well 

AFAIK `:tab` does not accept any argument or flags.

ibhagwan:
> > Moreso, this will also have to be ported to work for tabs as well
> 
> AFAIK `:tab` does not accept any argument or flags so this has nothing to do with tabs provider?

`tabs` still uses `nvim_list_bufs` and the same formatting and buffer markers (`#`, `%`), that's what I meant.

ibhagwan:
> > tabs still uses nvim_list_bufs and the same formatting and buffer markers (#, %), that's what I meant.
> 
> Yes, but this PR does not affect that. Buffer provider still uses `nvim_list_bufs()` if `ls_cmd` is not provided. And if it is provided it only affects the `buffer` provider but not the `tab` provider, which still uses `nvim_list_bufs()` regardless how buffer provider behaves.

Yes I understood that but I like to have feature parity between the two (in terms of formatting at least).

In any event, I really appreciate the effort but as I mentioned I would like to avoid having that extra complexity unless absolutely required.

Bekaboo:
Alright then, but can you check the patch and make the fix to the bufnr alignment logic? It is less than 10 lines of code and I don't want to bother making another PR :)

ibhagwan:
> Alright then, but can you check the patch and make the fix to the bufnr alignment logic? It is less than 10 lines of code and I don't want to bother making another PR :)

Absolutely, I just never want to prevent people from being counted as contributors :-)

ibhagwan:
@Bekaboo, simpler fix using the already existing `utils.ansi_escseq_len` (renamed from `ansi_col_len`):

https://github.com/ibhagwan/fzf-lua/commit/ad892f7787624171074d53c00eda42b952a37c72

Bekaboo:
Nice, thanks for the fix!

PR #987, opened 2024-01-07
fix(oldfiles): exclude unlisted buffers

danilshvalov:
At the moment, when the `include_current_session` option is enabled, extra files get into the oldfiles list (for example, files that got into the preview, but were not opened). It seems that only those files that were actually opened by the user should be added to oldfiles.


ibhagwan:
Ty for the fix @danilshvalov!

PR #988, opened 2024-01-07
fix(oldfiles): properly process opened buffers

danilshvalov:
In PR #987, I created a new bug by fixing the old one. Now, when the buffer is deleted, the file stops being shown in oldfiles. To fix this error, we had to slightly change the logic of getting opened buffers. Now all buffers that the user opens are saved in a table, which is then used in the picker.

The idea and part of the implementation was taken from [`telescope-recent-files`](https://github.com/smartpde/telescope-recent-files).

### DIFF
--- a/lua/fzf-lua/providers/oldfiles.lua
+++ b/lua/fzf-lua/providers/oldfiles.lua
@@ -1,37 +1,98 @@
-local core = require "fzf-lua.core"
-local utils = require "fzf-lua.utils"
-local config = require "fzf-lua.config"
-local make_entry = require "fzf-lua.make_entry"
+local core = require("fzf-lua.core")
+local config = require("fzf-lua.config")
+local make_entry = require("fzf-lua.make_entry")
 
 local M = {}
 
+M._recent_buffers = {}
+M._recent_counter = 0
+
+vim.api.nvim_create_autocmd("BufEnter", {
+  pattern = "*",
+  callback = function()
+    local buffer = vim.api.nvim_get_current_buf()
+    local file = vim.api.nvim_buf_get_name(buffer)
+    if file ~= "" then
+      M._recent_buffers[file] = M._recent_counter
+      M._recent_counter = M._recent_counter + 1
+    end
+  end,
+})
+
 M.oldfiles = function(opts)
   opts = config.normalize_opts(opts, "oldfiles")
-  if not opts then return end
+  if not opts then
+    return
+  end
 
   local current_buffer = vim.api.nvim_get_current_buf()
   local current_file = vim.api.nvim_buf_get_name(current_buffer)
-  local sess_tbl = {}
-  local sess_map = {}
+
+  local result_list = {}
+  local result_map = {}
+  local old_files_map = {}
+
+  local function add_recent_file(file)
+    if not file or #file == 0 or result_map[file] then
+      return
+    end
+
+    local fs_stat = not opts.stat_file and true or vim.loop.fs_stat(file)
+    if not fs_stat then
+      return
+    end
+
+    table.insert(result_list, file)
+    result_map[file] = true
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @danilshvalov, I’m currently traveling so didn’t have time to look at this thoroughly but at first glance I don’t like seeing an autocmd defined within a provider file, I try to keep these to a minimum and only add one if absolutely necessary and certainly not with a scope of `*`.

There has to be a better solution, if you can please accurately tell me the problem you’re trying solve and how to reproduce this issue I’m sure I can find a better solution that doesn’t involve a generic autocmd.

We already have a lot of info within `core.__CTX`, perhaps it can be utilized (I’m just not sure what we’re trying to solve here since I’m not using `include_current_session`):
https://github.com/ibhagwan/fzf-lua/blob/5890b84b07f842da292ee9067b7363ac2af02d92/lua/fzf-lua/core.lua#L187-L197

danilshvalov:
I need to save all the files that the user opened in this session. Is there a way to do it differently, a better way? @ibhagwan

ibhagwan:
> I need to save all the files that the user opened in this session. Is there a way to do it differently, a better way? @ibhagwan

I’m not sure yet. Can we start from why is this necessary and what is the exact problem we’re trying to solve? Namely how are you utilizing `include_current_session` in your workflow?

Once I can understand the need more accurately (and the bug this is attempting to solve) I’ll be able to better address your question.

danilshvalov:
> I’m not sure yet. Can we start from why is this necessary and what is the exact problem we’re trying to solve? Namely how are you utilizing `include_current_session` in your workflow?

In my workflow, I often navigate between already opened files using `oldfiles`. If I open a new file, I also want to see it in `oldfiles` so that I can quickly switch to it.

I know that this can be done through `buffers`, but when using `oldfiles`, you don't need to think about whether this file was opened earlier or not.

danilshvalov:
> Why do we need to do this? If the file is open it will appear on buffers and if the user closed the buffer, the file will automatically be added to old files, isn’t that the case?

It seems that the file will be added to oldfiles only after restarting neovim. That's the problem.

ibhagwan:
> > Why do we need to do this? If the file is open it will appear on buffers and if the user closed the buffer, the file will automatically be added to old files, isn’t that the case?
> 
> It seems that the file will be added to oldfiles only after restarting neovim. That's the problem.

I see, well that’s a `shada` issue, I’d rather find a better solution for session management, I’d rather avoid managing the sessions file inside fzf-lua, it feels out of scope.

I’m sure there is a way of force a shada file update manually, if we find how to do to you can add that command to buffer close autocmd.

ibhagwan:
From `:help shada-read-write`:

```vimdoc
MANUALLY READING AND WRITING				*shada-read-write*

Two commands can be used to read and write the ShaDa file manually.  This
can be used to exchange registers between two running Vim programs: First
type ":wsh" in one and then ":rsh" in the other.  Note that if the register
already contained something, then ":rsh!" would be required.  Also note
however that this means everything will be overwritten with information from
the first Vim, including the command line history, etc.
```

Can you try adding an autocmd for either `BufDelete` or `BufWipeout` that calls `:wsh` and see if that solves your issue?

danilshvalov:
It seems to be working:

```lua
vim.api.nvim_create_autocmd({ "BufDelete", "BufWipeout" }, {
  pattern = "*",
  command = "wshada",
})

vim.keymap.set("n", "<somekey>", function()
  vim.cmd("rshada!")
  require("fzf-lua").oldfiles()
end)
```

Thank you, I'm closing the PR.

ibhagwan:
Nice touch on calling `rshada!` I didn’t realize it was also needed to refresh:)

PR #990, opened 2024-01-09
fix: make `opts` optional in `fzf-lua.utils.fzf_version`

wookayin:
... so that `require("fzf-lua.utils").fzf_version()` can work, no need
to call like `require("fzf-lua.utils").fzf_version({})`.


ibhagwan:
Ty @wookayin!

wookayin:
You are always super fast to respond. Thanks too!

PR #991, opened 2024-01-09
refactor(utils)!: use vim.system for running process if available

wookayin:
Use the new `vim.system()` API on neovim 0.10 or higher, in place of
`vim.fn.system()`. It's more performant by using libuv (`vim.loop`),
though only slightly and known to have less glitches.

### DIFF
--- a/lua/fzf-lua/init.lua
+++ b/lua/fzf-lua/init.lua
@@ -138,8 +138,6 @@ function M.setup(opts, do_not_reset_defaults)
       opts[gopt] = nil
     end
   end
-  -- set lua_io if caller requested
-  utils.set_lua_io(opts.lua_io)
   -- set custom &nbsp if caller requested
   if opts.nbsp then utils.nbsp = opts.nbsp end
   -- store the setup options
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -859,46 +859,27 @@ function M.keymap_set(mode, lhs, rhs, opts)
   end
 end
 
--- speed up exteral commands (issue #126)
-local _use_lua_io = false
-function M.set_lua_io(b)
-  _use_lua_io = b
-  if _use_lua_io then
-    M.warn("using experimental feature 'lua_io'")
-  end
-end
-
-function M.io_systemlist(cmd, use_lua_io)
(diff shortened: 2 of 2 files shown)


wookayin:
Two remarks:

- The `use_lua_io` parameter is actually not being used, and it even doesn't work! Note the line `cmd == "string"` (not `type(cmd) == "string").
   - I find this implementation is vulnerable to injection, and no longer used. This was introduced in #126 for speed concerns, but we never use given "shell" cmdline (actually we *should never* use it), how about removing this completely? It's an internal utility so it would be fine.
 
- You might wonder why I wanted to have this feature in the first place, given that `vim.fn.system({"cmd", "--args"})` just works. Well, actually there are some subtle glitches of neovim: when `vim.fn.system()` is running, it may have some unwanted effects on TUI. For example, the cursor blinks on startup if `vim.fn.system()` is being called during initialization (what a weird bug).
  - See this GIF; the cursor blinks during plugin initialization, when `fzf-lua.utils.git_version()` is called: <br/> 
  - This is some bug that might have to be fixed on neovim core, but since the bug is around in neovim 0.9.x stable, it'd be good to have some fix on fzf-lua's side too. Actually `vim.system()` should be always preferred if available in neovim 0.10+.
  - Somehow this doesn't happen with the new libUV-based `vim.system` implementation.

ibhagwan:
Ty @wookayin, this is great, I had no idea about this change, I also much prefer a libuv based solution.

> The use_lua_io parameter is actually not being used and it even doesn't work! Note the line cmd == "string" (not type(cmd) == "string").

I agree, this can be safely removed, it was never used (as you astutely pointed out) and the concern for performance raised in #126 turned out of be a different issue.

As for removing it, we can add it to this PR, have a new PR or I can remove it myself tomorrow - whichever you prefer, lmk.

> You might wonder why I wanted to have this feature in the first place

ty for this explanation!

wookayin:
I think it'd be better to address here together. Hope I didn't break anything lol

ibhagwan:
> I think it'd be better to address here together.

Looks great @wookayin, much better without the never actually used code :)

> Hope I didn't break anything lol

We’ee about to find out since I’m merging it rn :)

PR #998, opened 2024-01-13
feat: save changes automatically with nohidden and autowriteall

guidocella:


ibhagwan:
Ty @guidocella, I wasn’t aware of this option.

PR #1000, opened 2024-01-14
Encode tab_cwd in bas64 and add it to tabs buffer entry as field 2

lbrayner:
With `tab_cwd` in entry, one can do something like:

```lua
fzf.tabs({
  fzf_opts = {
    ["--preview"] = [['echo "Tab #"{1}": $(echo {2} | base64 -d -)"']],
    ["--preview-window"] = "down,1",
  }
})
```

And get:

Screenshot from 2024-01-14 13-11-10
Screenshot from 2024-01-14 13-11-20
Screenshot from 2024-01-14 13-11-25

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -450,7 +450,7 @@ M.defaults.tabs = {
   },
   fzf_opts    = {
     ["--delimiter"] = "'[\\):]'",
-    ["--with-nth"]  = "2..",
+    ["--with-nth"]  = "3..",
   },
   _cached_hls = { "buf_nr", "buf_flag_cur", "buf_flag_alt", "tab_title", "tab_marker" },
 }
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -371,11 +371,12 @@ M.tabs = function(opts)
           return msg, hl
         end
 
+        local tab_cwd_tilde = path.HOME_to_tilde(tab_cwd)
         local title, fn_title_hl = opt_hl("tab_title",
           function(s)
             return string.format("%s%s#%d%s", s, utils.nbsp, t,
               (vim.loop.cwd() == tab_cwd and ""
-                or string.format(": %s", path.HOME_to_tilde(tab_cwd))))
+                or string.format(": %s", tab_cwd_tilde)))
           end,
           utils.ansi_codes[opts.hls.tab_title])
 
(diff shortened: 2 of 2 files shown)


ibhagwan:
Looks great @lbrayner!

I’m wondering, is the base64 encoding needed? I’m currently working on adding windows support and would rather avoid extra dependencies (base64 binary to decode).

What are we really missing if we add the cwd to the entry as is (without base64)?

lbrayner:
> Looks great @lbrayner!
> 
> I’m wondering, is the base64 encoding needed? I’m currently working on adding windows support and would rather avoid extra dependencies (base64 binary to decode).
> 
> What are we really missing if we add the cwd to the entry as is (without base64)?

@ibhagwan Hey! base64 is just to avoid the hassle of escaping characters. So to avoid the bas64 dependency, you could either try to escape characters (on Unix, I think everything is allowed except /, right), or use Neovim RPC invocations, like you do with the reload functions. What do you think?

ibhagwan:
Entries are passed into the fzf pipe directly I don’t think we need to escape it here, try it without and let me know?

ibhagwan:
Just tested this without shell escaping, works great without the `base64` decode, just with:
```lua
  fzf_opts    = {
    ["--preview"] = [['echo "Tab #"{1}": "{2}""']],
    ["--preview-window"] = "nohidden:up,1",
  },
```
image

Manually merged (without base64) at https://github.com/ibhagwan/fzf-lua/commit/2fca0f2ae6064aa106feef6b527a7dbefb825724

lbrayner:
@ibhagwan  This is what I meant:

Screenshot from 2024-01-15 07-42-11

It breaks if CWD constains a colon. In this case the folder name is `~/code/my-app-typescript: a (copy)`.

ibhagwan:
That because I used a colon delimiter, we can choose a different one.

> @ibhagwan ditch base64 dependency

How does this command solve your issue if the directory isn’t encoded in base64?

lbrayner:
> That because I used a colon delimiter, we can choose a different one.

Any character is allowed in a UNIX file name, that is the point. No matter what separator you choose, it can be broken, it's not a robust solution.

> How does this command solve your issue if the directory isn’t encoded in base64?

If the directory name is encoded in base64, it can be decoded without extra dependencies.

ibhagwan:
> If the directory name is encoded in base64, it can be decoded without extra dependencies.

I see what you meant now, I was confused given the current directory isn’t encoded in base 64.

> Any character is allowed in a UNIX file name, that is the point. No matter what separator you choose, it can be broken, it's not a robust solution.

Alright, I’ll make the change later when I’m next to the computer, if you wish to have that earlier and wish to submit another PR you’re welcome to :)

lbrayner:
> Alright, I’ll make the change later when I’m next to the computer, if you wish to have that earlier and wish to submit another PR you’re welcome to :)

Done!

ibhagwan:
> > Alright, I’ll make the change later when I’m next to the computer, if you wish to have that earlier and wish to submit another PR you’re welcome to :)
> 
> Done!

Merged :)

PR #1003, opened 2024-01-15
Encode tab_cwd in base64 (*NIX file names are permissive)

lbrayner:
See discussion in #1000.

Since *NIX file names allow any character, using them as is in an entry can break tabs functionality. The solution is to encode them in base64 and use a separator that is not in the base64 table, such as colon.

To decode the directory name, Neovim itself can be used:

```lua
fzf.tabs({
  fzf_opts = {
    ["--preview"] = [['echo -n "Tab #"{1}": "; nvim -u NONE -es +":verbose lua print(vim.base64.decode({2}))"']],
    ["--preview-window"] = "up,1",
  }
})
```

Or one can use `base64`:

```lua
fzf.tabs({
  fzf_opts = {
    ["--preview"] = [['echo "Tab #"{1}": $(echo {2} | base64 -d -)"']],
    ["--preview-window"] = "up,1",
  }
})
```

### DIFF
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -384,8 +384,9 @@ M.tabs = function(opts)
           function(s) return s end,
           utils.ansi_codes[opts.hls.tab_marker])
 
+        local tab_cwd_tilde_base64 = vim.base64.encode(tab_cwd_tilde)
         if not opts.current_tab_only then
-          cb(string.format("%d)%s:%s%s\t%s", t, tab_cwd_tilde, utils.nbsp,
+          cb(string.format("%d)%s:%s%s\t%s", t, tab_cwd_tilde_base64, utils.nbsp,
             fn_title_hl(title),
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @lbrayner!

ibhagwan:
@lbrayner, turns out `vim.base64` is a new addition of nvim `0.1`, I modified the code for conditional encoding.

https://github.com/ibhagwan/fzf-lua/commit/7b7e80d33468dbf98506d2ce2f185bb9dc4923f3

lbrayner:
@ibhagwan sorry about that. Maybe you should just disable it by setting `tab_cwd_tilde_base64` to empty string `""`:

```lua
       local tab_cwd_tilde_base64 = vim.base64
            and vim.base64.encode(tab_cwd_tilde)
            or ""
```

ibhagwan:
It’s alright, the current solution is acceptable, i’ll consider adding the base64 single file lib with the windows version if I find that it makes escaping easier, if UNIX escaping can be annoying windows escaping is pure hell lol

PR #1008, opened 2024-01-20
fix(changes): add defaults values for changes

mikesmithgh:
I noticed that the default values weren't set and calling `normalize_opts` was missing for the `changes` builtin. This was causing issues when I was trying to override the prompt in my configuration. I've updated the defaults, please let me know if any other changes are needed.


ibhagwan:
That was me being “cheap” on design, looks great @mikesmithgh :)

PR #1024, opened 2024-02-02
fix: only restore split layout for existing windows

### LINKED ISSUE #1022: cmdheight increases sometimes on exiting split fzf window
### Description 

Using `fzf-lua` in a split window sometimes on exiting the fzf window the cmd-line gets really big (taking up 80-90% of the nvim window height). This is the same issue described in this comment: https://github.com/ibhagwan/fzf-lua/commit/c6b2ce724417e6a74d183e8770452e0ece00593f#r138166682

The issue occurs when a floating window that is present when we open the fzf window gets closed **before** we exit the fzf window again. A common use case would be for example if you use [fidget.nvim](https://github.com/j-hui/fidget.nvim)

Since c6b2ce724417e6a74d183e8770452e0ece00593f we are saving the windowlayout using `winrestcmd` to restore it properly afterwards. If the available windows change after we saved the command it seems the restoration fails in this way. The same can happen if a split is closed before the restore command is executed, but in this case I could not always reproduce it. Since I think closing a split while using fzf is highly unlikely  (as opposed to a floating window) this is probably not really a big issue.

### Reproduction

- [x] The issue is reproducible with `mini.sh` + the following configuration
```lua
require('fzf-lua').setup({
    winopts = {
        split = 'botright new'
    }
})
```

- open a floating window and save the win_id  
```lua
:lua= vim.api.nvim_open_win(0, false, { relative = 'editor', width = 20, height = 5, row = 0, col = 0, border = 'single' })`
```
- open any fzf provider in a split
- escape the fzf teminal via `CTRL-\ CTRL-N`
- close the floating window
```lua
:lua vim.api.nvim_win_close(<winid>, false)
```
- close the fzf window (doesn't matter how)

TheBlob42:
Should fix #1022 

So far I only tested it briefly, but it should fix the issue by removing all windows from the restore command that do not exists anymore. Happy for feedback or help with testing.

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -939,7 +939,21 @@ function FzfWin:close(fzf_bufnr)
     vim.api.nvim_set_current_win(self.src_winid)
   end
   if self.winopts.split then
-    vim.cmd(self.winrestcmd)
+    -- remove all windows from the restore cmd that have been closed in the meantime
+    -- if we're not doing this the result might be all over the place
+    local winnrs = vim.tbl_map(function(win)
+      return vim.api.nvim_win_get_number(win)
+    end, vim.api.nvim_tabpage_list_wins(0))
+
+    local cmd = ""
+    for cmd_part in string.gmatch(self.winrestcmd, "[^|]+") do
+      local winnr = cmd_part:match("(.)resize")
+      if vim.tbl_contains(winnrs, winnr) then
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tysm for the investigation and fix @TheBlob42.

TheBlob42:
Unfortunately I think I did a mistake here :grimacing: 
I never added the `|` characters back into the final command, I will check this

PR #1027, opened 2024-02-04
fix(actions): toggle-ignore not preserving cwd

Bekaboo:
`toggle-ignore` action always resets cwd to the current working directory, this PR fixes the issue.


ibhagwan:
Ty @Bekaboo!

PR #1036, opened 2024-02-10
Allow setting --preview-window to false to not override FZF defaults

deathbeam:
Currently when fzf_opts["--preview-window"] is set to false, it works as long as opts.preview_offset is not set. But this is set for grep by default so it throws error. This checks if --preview-window is set before adding preview offset to it.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -502,7 +502,7 @@ M.build_fzf_cli = function(opts)
   if opts.fzf_opts["--preview-window"] == nil then
     opts.fzf_opts["--preview-window"] = M.preview_window(opts)
   end
-  if opts.preview_offset and #opts.preview_offset > 0 then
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @deathbeam, I would prefer if the condition also accepted `nil` value, I.e.:
```lua
if opts.fzf_opts["--preview-window"] ~= false …
```

deathbeam:
> Ty @deathbeam, I would prefer if the condition also accepted `nil` value, I.e.:
> 
> ```lua
> if opts.fzf_opts["--preview-window"] ~= false …
> ```

It being nil should not be possible I think because of the nil check above (where it just sets default value).

ibhagwan:
> It being nil should not be possible I think because of the nil check above (where it just sets default value).

My bad, didn’t even look at the surrounding code, merged.

PR #1057, opened 2024-02-27
fix(codeaction preview) update internal _items list with resolve results

### LINKED ISSUE #1007: Problems with code actions and jdtls language server
### Info

- Operating System: ArcoLinux
- Shell: Fish
- Terminal: wezterm
- `nvim --version`: NVIM v0.9.5
- `fzf --version`: 0.44.1 (29e67d30)

<details>

```lua
-- require('fzf-lua').setup({})
require('fzf-lua').register_ui_select()
```

My [full config](https://github.com/niveK77pur/nvim/blob/021839139446f10131c667fbbae0168facd104c1/lua/myplugins/HANDY_TOOLS_fzf_lua.lua) where I also tried disabling all configurations other than the one right above. Also tried disabling all other plugins to make sure the problem doesn't stem elsewhere.
</details>

### Description 

Code actions with jdtls are not working properly; meaning they either do not seem to trigger at all, or incorrectly modify the buffer and delete content. Simply disabling this plugin makes the code actions work as expected.

#314 seems related, but in this case it is precisely the `register_ui_select` that seems to be causing issues.

#### With fzf-lua's `register_ui_select`

Using the above configuration (only `require('fzf-lua').register_ui_select()`)

[video-Kazam_screencast_00001.webm](https://github.com/ibhagwan/fzf-lua/assets/10981161/ae779583-a4e3-47c8-bf42-a5f035055133)

#### Without fzf-lua

Completlely disabling the plugin, removing the `register_ui_select` or calling `require('fzf-lua').deregister_ui_select()` will have the same effect.

[video-Kazam_screencast_00000.webm](https://github.com/ibhagwan/fzf-lua/assets/10981161/a0fff1b7-27fa-4ce0-95a8-27a2eb4bb20c)

nabellows:
Fixes #1007 but in a non-JDTLS-specific way. In fact, by updating _items this removes unnecessary `codeAction/resolve` calls from being made if toggling the previewer, scrolling past and back to a preview result, and then within the nvim internal `on_code_action_results` implementation. This could potentially have some speedup, but more importantly, fixes the bug for JDTLS since JDTLS was returning different results between `codeAction/resolve` calls (but the first is correct).

In my testing this fixes the javaDoc/import codeActions for java (and any more I wasn't aware of with the same issue).

### DIFF
--- a/lua/fzf-lua/previewer/codeaction.lua
+++ b/lua/fzf-lua/previewer/codeaction.lua
@@ -142,7 +142,7 @@ local function diff_tuple(err, tuple, diff_opts)
 end
 
 -- https://github.com/neovim/neovim/blob/v0.9.4/runtime/lua/vim/lsp/buf.lua#L666
-local function preview_action_tuple(tuple, diff_opts, callback)
+local function preview_action_tuple(self, tuple, idx, callback)
   -- neovim changed the ui.select params with 0.10.0 (#947)
   -- { client_id, action } ==> { ctx = <LSP context>, action = <action> }
   if tuple.ctx then
@@ -157,29 +157,34 @@ local function preview_action_tuple(tuple, diff_opts, callback)
   then
     local function on_result(diff_callback, err, resolved_action)
       if err then
-        return diff_callback(err, tuple, diff_opts)
+        return diff_callback(err, tuple, self.diff_opts)
       else
-        return diff_callback(err, { tuple[1], resolved_action }, diff_opts)
+        return diff_callback(err, { tuple[1], resolved_action }, self.diff_opts)
       end
+    end
+    local function update_internal_items(resolved_action)
+      self.opts._items[idx] = { tuple[1], resolved_action }
     end
 
     if callback then
       client.request("codeAction/resolve", action, function(err, resolved_action)
+        update_internal_items(resolved_action)
         on_result(callback, err, resolved_action)
       end)
       return { string.format("Resolving action (%s)...", action.kind) }
     else
       local res = client.request_sync("codeAction/resolve", action)
       local err, resolved_action = res and res.err, res and res.result
+      update_internal_items(resolved_action)
       if type(err) == "table" or type(resolved_action) == "table" then
         return on_result(diff_tuple, err, resolved_action)
       else
         -- display the default "unsupported" message
-        return diff_tuple(nil, tuple, diff_opts)
+        return diff_tuple(nil, tuple, self.diff_opts)
       end
     end
   else
-    return diff_tuple(nil, tuple, diff_opts)
+    return diff_tuple(nil, tuple, self.diff_opts)
   end
 end
 
@@ -208,7 +213,7 @@ function M.builtin:populate_preview_buf(entry_str)
   local idx = tonumber(entry_str:match("^%d+%."))
   assert(type(idx) == "number")
   local tuple = self.opts._items[idx]
-  local lines = preview_action_tuple(tuple, self.diff_opts,
+  local lines = preview_action_tuple(self, tuple, idx,
     -- use the async version for "codeAction/resolve"
     function(err, resolved_tuple)
       if vim.api.nvim_buf_is_valid(self.tmpbuf) then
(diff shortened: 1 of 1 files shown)


nabellows:
If there are any stylistic/naming changes I'll gladly revise. I am actually pretty new to lua (debugging this issue myself was pretty painful :) )and don't know the conventions.

ibhagwan:
I see another room for improvement, if the call to resolve succeeded but the result isn't an `edit` the next time we preview the item there will be another call made to resolve, I'll fix that as well when merging.

ibhagwan:
I think this might need a different approach, overriding the `_items` isn't future-safe, neovim can change the way it resolves code actions or the callback for `ui.select` at which point our call to `on_choice` will fail.

That said, if we don't override the item tuple this might not solve #1007, I'll have to do some more testing.

nabellows:
> EDIT: was the JDTLS bug happening only on consequent calls?
Yes, in my debugging I found that the first codeAction/resolve was correct from JDTLS. The second call would reply something that seemed to try to overwrite its old document My best guess of the root cause relationship to https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/376 is that JDTLS assumes the edit has been made? Or maybe this is some other bug within JDTLS with repeated resolve calls.  

> I see another room for improvement, if the call to resolve succeeded but the result isn't an edit the next time we preview the item there will be another call made to resolve, I'll fix that as well when merging.

I'm not sure about this, since I think this is the contract of the codeAction/resolve method. Would you be expecting something else, like action.command? 

> I think this might need a different approach, overriding the `_items` isn't future-safe, neovim can change the way it resolves code actions or the callback for `ui.select` at which point our call to `on_choice` will fail.
 
 True, but I'm not sure how likely this is. I guess the best solution would be neovim providing a hook into their own `codeAction/resolve` caller and they provide this memoization for us. However,

ibhagwan:
> True, but I'm not sure how likely this is. I guess the best solution would be neovim providing a hook into their own codeAction/resolve caller and they provide this memoization for us. Ideally they should be providing a hook/function callback parameter instead of this vim.ui.select hack.

The issue is already prevalent in the code, in neovim 0.10.0 the ui.select items changed from `{ client_id, action }` to `{ ctx = ..., action = ... }`,  hence these lines ( the code already fails with neovim nightly because of this change):
https://github.com/ibhagwan/fzf-lua/blob/b697e269475b07c191afcb89fc8d4708d9b9f8a4/lua/fzf-lua/previewer/codeaction.lua#L146-L150

I have a solution for all of this including a "hack" that overwrites the resolved action only for `jdtls`, testing now and will merge soon.

ibhagwan:
Due to all the reasons mentioned above I manually merged your commit (https://github.com/ibhagwan/fzf-lua/commit/2cbe90ddbc66c4caeb8919d50867ff312c9e3847) and improved on it with https://github.com/ibhagwan/fzf-lua/commit/903a4ae4e838170e4f44fb4ce1ca647a3c9cf828.

The latest also fixes an issue where actions wouldn't be resolved on the nightly as the test for the resolving capability changed and is now:
https://github.com/ibhagwan/fzf-lua/blob/903a4ae4e838170e4f44fb4ce1ca647a3c9cf828/lua/fzf-lua/previewer/codeaction.lua#L163-L175

The jdtls "hack" was also updated to work on both 0.9 and the nightly:
https://github.com/ibhagwan/fzf-lua/blob/903a4ae4e838170e4f44fb4ce1ca647a3c9cf828/lua/fzf-lua/previewer/codeaction.lua#L188-L194

nabellows:
I like your solution! Excellent response time too!

Any reason why the jdtls hack is done specifically for jdtls? I guess to reduce blast radius to just jdtls if code_actions ui_select input changes? Otherwise I think it reduces latency of the code action when applied since you skip the LSP call.

ibhagwan:
> Any reason why the jdtls hack is done specifically for jdtls? I guess to reduce blast radius to just jdtls if code_actions ui_select input changes? Otherwise I think it reduces latency of the code action when applied since you skip the LSP call.

Exactly that, I would like to avoid modifying neovim internal structures, it doesn’t make any sense to have the items modified when running the previewer and unmodified if running with `previewer=false`, although unlikely, if this part of the runtime changes in the future I can already see the mess this kind of change would create.

As for the latency, perhaps it’s a nice to have benefit but I feel we shouldn’t try to optimize here at the expense of modifying internal LSP code.

PR #1060, opened 2024-02-28
Set proper native codeaction previewer for native profiles

deathbeam:
The codeaction issue made me remember that I wanted to PR this


ibhagwan:
Ty @deathbeam, I wanted to do this as well but it slipped my mind :)

PR #1061, opened 2024-02-28
Add missing _ to profile codeactions profile previewer

deathbeam:
Not sure why I missed it but it should be code_actions not codeactions.


ibhagwan:
Merged and squashed into one commit https://github.com/ibhagwan/fzf-lua/commit/785e7db2d5db9d173c8a5718df5c5219c9196352

PR #1067, opened 2024-03-01
fix(manpages): preview and open the correct section for the man page

johslarsen:
The man pages are grouped into different sections and it is common for identically named pages to be found in multiple sections. For instance `printf(1)` documenting the shell command and `printf(3)` documenting the C-function.

Prior to these changes the section name was stripped away from the selected entries, so it only previews and opens that page from the default section even though another section was selected. Now the apropos entry is rewritten to a `:Man` argument with the section name.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -543,9 +543,7 @@ M.help_tab = function(selected)
 end
 
 local function mantags(s)
-  return vim.tbl_map(function(x)
-    return x:match("[^[,( ]+")
-  end, s)
+  return vim.tbl_map(require'fzf-lua.providers.manpages'.manpage_vim_arg, s)
 end
 
 M.man = function(selected)
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -986,8 +986,7 @@ function Previewer.man_pages:new(o, opts, fzf_win)
 end
 
 function Previewer.man_pages:parse_entry(entry_str)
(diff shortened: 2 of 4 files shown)


ibhagwan:
Tysm @johslarsen, this makes sense.

Unfortuntaely in the past this turned out to be problematic, I will admit I didn't dive into it properly but if I recall correctly there are two different formats (also `mandb`?).

For example, on my machine (Void Linux)  this doesn't work:
image

ibhagwan:
@johslarsen, seems we've been through this before, see #25, I just can't recall ATM why I reverted but it was failing on some systems.

johslarsen:
OK, I see that it was a bit more complicated than it looked. The latest version should work better.

I have tested the latest version with neovim-0.9.5-2 and man-db 2.12.0-1 / mandoc 1.14.6-4 on archlinux. Can you check if it works better on your end now?

I changed it to commands like `man 3 printf`, which I hope are more readily supported all around. Note that for `:Man` in vim that syntax did not seem to like that syntax so for the action I kept it like `:Man printf(3)`.

ibhagwan:
Ineed that's better @johslarsen though I still see an issue when the descirption string  contains `(..)`, for example:
image

ibhagwan:
I'm assuming we just need change the  lua pattern to non-greedy inside the `()`.

johslarsen:
> I'm assuming we just need change the lua pattern to non-greedy inside the `()`.

yep, disallowing `(` in the page name helped

ibhagwan:
Merged, tysm for the contribution and the diligence in this PR @johslarsen!

Issue #1040, opened 2024-02-14
Split layout with percentage height


SlartThunder:
### Info

- Operating System: MacOS
- Shell: omz
- Terminal: iTerm
- `nvim --version`:NVIM v0.9.4
- `fzf --version`:0.42.0 (d471067)

- [x] The issue is reproducible with `mini.sh`

### Description 
I recently switched from NVIM from FZF to use FZF-LUA. 

In FZF you can set the layout to be a floating window, or stick to a position in current buffer.  
E.g: let g:fzf_layout = { 'down': '50%'}
The above command would open FZF from down the active buffer, which is very handy when you have splits all around and you want to change the buffer in that particular window.

ibhagwan:
You can use `winopts.split` to pass any command (including percentages), for example:
```lua
require'fzf-lua'.setup({
  -- "belowright new"  : split below
  -- "aboveleft new"   : split above
  -- "belowright vnew" : split right
  -- "aboveleft vnew   : split left
  winopts = { split = 'botright new' }
})
```

SlartThunder:
Thanks, I tried with 
split  = "belowright new | horizontal resize 30%"
but when having already a split, being active in the upper one, for example, when opening FZFLua it "floats" the buffer size, rather than using the 30% of the current one…
I hope I make myself understand…

ibhagwan:
> Thanks, I tried with split = "belowright new | horizontal resize 30%" but when having already a split, being active in the upper one, for example, when opening FZFLua it "floats" the buffer size, rather than using the 30% of the current one… I hope I make myself understand…

Resize doesn’t support percentage, if you want 30% you have to use a fraction of `lines`.
 
 If that’s what you want it’s better done with lua code, I can craft the 2 line function for you when I’m not AFK.

ibhagwan:
Here you go, modify this to your liking:
```lua
require("fzf-lua").setup({
  winopts_fn = function()
    -- local split = "botright new" -- use for split under **all** windows
    local split = "belowright new"  -- use for split under current window
    local height = math.floor(vim.o.lines * 0.3)
    return { split = split .. " | resize " .. tostring(height) }
  end,
})
```

SlartThunder:
Thanks for your code, though it didn't work as I expected.
Nevertheless, it gave me a hint to implement what I'm trying to achieve.
Basically, I'd like to have a different FZF setup based on status of NVIM, if it has splits or not.

I have a FZFLua custom config I call with
`
[...]

PR #1068, opened 2024-03-02
fix(preview): set flex threshold from window columns in split mode

johslarsen:
The split mode opens the fzf selection in a split from the current window, so the number of available columns can be less than the width of the full terminal, which should be taken into the account when calculating whether the width exceeds the flex threshold or not.

As discussed in #1040, fzf-tmux also have a split mode, in which case the split is relative to the full-width of the terminal window, so maybe the `fzf_bin` should also be taken into account here. Or maybe it should not because fzf-tmux only takes effect if the terminal is a tmux session (or maybe also check e.g. `vim.env.TMUX` to determine if fzf-tmux would have an effect). However, whilst testing that it turns out that having split mode set with fzf-tmux in a tmux session has a lot more serious issues. The behavior on my end (and with `mini.sh`) in that scenario is that the split pane works fine (have not tested these changes with it), but the selections gets lost. Considering this along with the default in fzf-tmux is to not actually use flex mode, so the likelihood that this breaks user configuration any further is pretty low, and it is probably safe to not take that into account for now.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -421,16 +421,10 @@ end
 ---@param o table
 ---@return string
 M.preview_window = function(o)
-  local preview_args = ("%s:%s:%s:"):format(
-    o.winopts.preview.hidden, o.winopts.preview.border, o.winopts.preview.wrap)
-  if o.winopts.preview.layout == "horizontal" or
-      o.winopts.preview.layout == "flex" and
-      vim.o.columns > o.winopts.preview.flip_columns then
-    preview_args = preview_args .. o.winopts.preview.horizontal
-  else
-    preview_args = preview_args .. o.winopts.preview.vertical
-  end
-  return preview_args
+  local hsplit = win:preview_splits_horizontally(o.winopts, 0)
+  local split = hsplit and o.winopts.preview.horizontal or o.winopts.preview.vertical
+  return ("%s:%s:%s:%s"):format(
+    o.winopts.preview.hidden, o.winopts.preview.border, o.winopts.preview.wrap, split)
 end
 
 -- Create fzf --color arguments from a table of vim highlight groups.
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -182,6 +182,21 @@ local strip_borderchars_hl = function(border)
   return borderchars
 end
 
+function FzfWin:preview_splits_horizontally(winopts, winid)
+  local columns = winopts.split and vim.api.nvim_win_get_width(winid) or vim.o.columns
+  return winopts.preview.layout == "horizontal" or
+      winopts.preview.layout == "flex" and
+      columns > winopts.preview.flip_columns
+end
+
+local function update_preview_split(winopts, winid)
+  local hsplit = FzfWin:preview_splits_horizontally(winopts, winid)
+  local preview = hsplit and winopts.preview.horizontal or winopts.preview.vertical
+  -- builtin previewer params
+  winopts.preview_pos = preview:match("[^:]+") or "right"
+  winopts.preview_size = tonumber(preview:match(":(%d+)%%")) or 50
+end
+
 local normalize_winopts = function(o)
   -- make a local copy of opts so we don't pollute the user's options
   local winopts = utils.tbl_deep_clone(o.winopts)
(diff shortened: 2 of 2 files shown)


ibhagwan:
This makes sense @johslarsen, TBH, I have a big wishlist of what to do with the window object/class, the code started as opening a single window and having fzf previews only, then progressed to neovim preview window, splits and the rest so it’s a bit messy and unsightly but since it “just works” and I’m a bit tired of refactoring I just let it be.

Optimally we should have a different class for split, tmux, etc and have the model built in such way that anyone could inherit from the base and create their own windows for each phase (main, preview, layout changes, etc), same way Telescope can work with windows created by nui.nvim.

Not too many people use splits and fzf-tmux so this isn’t high on the priority list, another example would be rotating the preview with splits or fzf-tmux, it can be supported with new fzf options.

I’ll briefly test this code and merge it, Ty!

ibhagwan:
Instead of checking for `vnew` and halving the columns, isn’t it better to check the width of the window after creation so the solution is more generic?

johslarsen:
Yes, I agree, and looked into that first. Just as a proof-of-concept I tried moving the calculation and setting of `winopts.preview_pos`/`size` to right after [where it was split](https://github.com/ibhagwan/fzf-lua/blob/ab4ce5031c52455468a1ac0606b38248aa919c7c/lua/fzf-lua/win.lua#L850), but there it had no effect. Later I found the other code in core.lua, which seems to run after the split already, so maybe my configuration ended up in that path or something. It is not obvious which of these two more or less identical handling of the flex size are used for what.

ibhagwan:
> Yes, I agree, and looked into that first. Just as a proof-of-concept I tried moving the calculation and setting of `winopts.preview_pos`/`size` to right after [where it was split](https://github.com/ibhagwan/fzf-lua/blob/ab4ce5031c52455468a1ac0606b38248aa919c7c/lua/fzf-lua/win.lua#L850), but there it had no effect. Later I found the other code in core.lua, which seems to run after the split already, so maybe my configuration ended up in that path or something. It is not obvious which of these two more or less identical handling of the flex size are used for what.

I can’t say I remember (without testing) which of these is the important one, as I tried to explain in the first comment split was added haphazardly and isn’t widely - it has lots of room for improvement.

johslarsen:
I have tested a bit more, and it seems like normalize is the one having most effect. Both with native and default presets, but now that I fixed the one in core.lua applying the config at the split seems to work nicely, so I will change to that.

Also it turns out that if you rescale the terminal whilst in a fzf selection the normalize methods gets called from within the split window, so doing any calculation there assuming that it will be split further won't work properly in all use cases

ibhagwan:
> Also it turns out that if you rescale the terminal whilst in a fzf selection the normalize methods gets called from within the split window, so doing any calculation there assuming that it will be split further won't work properly in all use cases

Perhaps we can use core.CTX for some of these

johslarsen:
I moved the calculation to the split, and that seems to work nicely now both for default and native profiles

ibhagwan:
> I moved the calculation to right after the split, and that seems to work nicely now both for default and native profiles

I’m AFK to test but it looks good, I’m thinking do we even need the code in core? It makes no sense there it should be only part of the window code.

johslarsen:
> I’m thinking do we even need the code in core?

I tried removing it. This breaks flex mode for native code. Makes sense though, since the difference between this and the other one is that it adds CLI arguments for fzf. I can refactor so both use one method to figure out what sort of split, and then the difference is only what to do for vertical/horizontal splits

johslarsen:
Something like this. That seems to work fine both for native and builtin mode

ibhagwan:
> Something like this. That seems to work fine both for native and builtin mode

Looks good, I’ll do some testing when I get a chance and merge.

ibhagwan:
@johslarsen, merged, take a look at my following commit https://github.com/ibhagwan/fzf-lua/commit/f2190ac04daf8871de300bd358e803ba42331b05

In the case we're using `fzf-tmux` (and `$TMUX` exists) we acquire the number of available columns from the tmux window using `tmux display-message -p '#{window_width}'`:
https://github.com/ibhagwan/fzf-lua/blob/f2190ac04daf8871de300bd358e803ba42331b05/lua/fzf-lua/config.lua#L507-L510

And then use it in:
https://github.com/ibhagwan/fzf-lua/blob/f2190ac04daf8871de300bd358e803ba42331b05/lua/fzf-lua/win.lua#L186-L188

johslarsen:
Yes, that should be more correct for the the fzf-tmux `-p` use case where it creates a full terminal floating window.

For the "-d/..." use case where it splits the terminal, fzf-tmux does it the same way as vim, so if vim was in a vertically split tmux pane to begin with the vim.o.columns value would be more correct than the tmux `window_width`. For `-l/-r` vertical splits it would be even less, but since it turns out that the `normalize_opts` function was called both before and after the window split any logic trying to guess the new size would be just as flawed as the one I tried originally.

ibhagwan:
> For the "-d/..." use case where it splits the terminal, fzf-tmux does it the same way as vim, so if vim was in a vertically split tmux pane to begin with the vim.o.columns value would be more correct than the tmux window_width. For -l/-r vertical splits it would be even less, but since it turns out that the normalize_opts function was called both before and after the window split any logic trying to guess the new size would be just as flawed as the one I tried originally.

Indeed, as I mentioned, this part of the code does require a refactor one day in the future.

PR #1069, opened 2024-03-02
feat(quickfix): add an only_valid option

johslarsen:
So that the quickfix/loclist providers can include less noise to begin with if users prefers that. Much like the `:clist` command that you need to add a bang to for it to include the invalid entries.


ibhagwan:
Ty @johslarsen!

PR #1070, opened 2024-03-02
fix(quickfix): handle multi-line entries more like the quicklist window

johslarsen:
That means having them displayed without new lines so that they are concatenated into on one long line. Prior to these changes if a quickfix entry included newlines those would become separate entries in the fzf selection list. Quickfix entries are formatted so they start with the file name and location, but only the first line is like that, so if one were to select one of the other lines from an entry the start of this line would be interpreted as a file name and would open a buffer "there"

Here is an example of a multiline error (not really a realistic one as it does not have a filename to begin with though), before:
quickfix_multiline_before
After:
quickfix_multiline_after

### DIFF
--- a/lua/fzf-lua/providers/quickfix.lua
+++ b/lua/fzf-lua/providers/quickfix.lua
@@ -16,6 +16,7 @@ local quickfix_run = function(opts, cfg, locations)
 
   for _, entry in ipairs(locations) do
     if entry.valid == 1 or not opts.only_valid then
+      entry.text = entry.text:gsub("\r?\n", " ")
       table.insert(results, make_entry.lcol(entry, opts))
     end
   end


ibhagwan:
Ty @johslarsen, I had no idea this was a thing.

johslarsen:
These multi-line errorformats are often used to create for instance a warning from an exception where the message is parsed from one line and the filename/... is parsed from the top of a stack trace. In those cases the message text itself is still single-lined, so I am guessing it is pretty uncommon to actually get new lines in the output.

I use it for test output. There it is very useful for summarizing assertion faults, like e.g. googletest output:
```
/home/johs/work/playground/abseil/test/strings_test.cpp:53: Failure
Value of: absl::StrContains("f", "o")
  Actual: false
Expected: true

```
Summarized in the quickfix window like:
```
test/strings_test.cpp|53 F|  Value of: absl::StrContains("f", "o") Actual: false Expected: true
```

ibhagwan:
Very nice @johslarsen

PR #1071, opened 2024-03-03
fix(fzf-tmux): get flex layout for split panes from vim width

johslarsen:
In case the tmux terminal is vertically split already, because then the pane will be split from the current pane not the full window width.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -285,9 +285,11 @@ function M.normalize_opts(opts, globals, __resume_key)
 
   -- prioritize fzf-tmux split pane flags over the
   -- popup flag `-p` from fzf-lua defaults (#865)
+  opts._is_fzf_tmux_popup = true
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @johslarsen!

Does this conclude all the split flex sizing issues?

johslarsen:
Mostly yes. Still left with whatever is causing major issues if you use fzf-tmux with split mode, but I doubt anyone actually care about that. The only reason I see for configuring both would be if you want to fall back to split instead of floating mode if you use fzf-tmux, but are not in a tmux session.

ibhagwan:
> Mostly yes. Still left with whatever is causing major issues if you use fzf-tmux with split mode, but I doubt anyone actually care about that. The only reason I see for configuring both would be if you want to fall back to split instead of floating mode if you use fzf-tmux, but are not in a tmux session.

I’m failing to imagine how these options work together at all, IMHO if someone is using fzf-tmux and tmux env exists, split should be disabled.

PR #1073, opened 2024-03-03
fix(fzf-tmux): disable split mode when using fzf-tmux in a tmux session

johslarsen:
Because for some reason the selection result does not have any effect when both are configured in and you are in a tmux session.

> I’m failing to imagine how these options work together at all, IMHO if someone is using fzf-tmux and tmux env exists, split should be disabled.

Yes, that seems to solve the problem, so that is probably better than to just leave it as is.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -509,6 +509,7 @@ function M.normalize_opts(opts, globals, __resume_key)
   if opts._is_fzf_tmux then
     local out = utils.io_system({ "tmux", "display-message", "-p", "#{window_width}" })
     opts._tmux_columns = tonumber(out:match("%d+"))
+    opts.winopts.split = nil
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @johslarsen for your thoughtful contributions.

johslarsen:
No, thank you. There are a lot of open-source project out there, and most of them are not as easy to contribute to as this one. After switching from vim to neovim this seemed like a more future-proof way to go than to stick with e.g. the original `fzf.vim`. Lua is nowhere near a perfect language, but it is a lot better than vimscript at least.

ibhagwan:
> Lua is nowhere near a perfect language, but it is a lot better than vimscript at least.

Couldn’t agree more, I was far from an expert in lua before this plugin, it’s a very easy and straight forward language.

PR #1074, opened 2024-03-07
feat(diagnostic): support 0.10 extmark signs

ofseed:
After https://github.com/neovim/neovim/pull/26193 , the internal implementation of the diagnostic signs changed to the new extmark-based sign, sign_getdefine will no longer work.

### DIFF
--- a/lua/fzf-lua/providers/diagnostic.lua
+++ b/lua/fzf-lua/providers/diagnostic.lua
@@ -68,14 +68,24 @@ M.diagnostics = function(opts)
     opts.__signs[v.severity] = {}
 
     -- from vim.diagnostic
-    local sign_def = vim.fn.sign_getdefined(v.name)
-    -- can be empty when config set to (#480):
-    -- vim.diagnostic.config({ signs = false })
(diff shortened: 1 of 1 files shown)


ibhagwan:
Looks great @ofseed, only change I would do is the call to `vin.fn.has` which I commented above.

ofseed:
I'm sorry, I didn't quite catch that, could you explain in more detail? And you can make changes on this branch directly.

ibhagwan:
> I'm a bit confused about what you just said, could you explain in more detail? And you can make changes on this branch directly.

Np, it’s a minor change, I’ll merge manually and make the change and link the line here.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/db254a901fc3ae657e3693dd78d37d5e2ae3e945, merged.

This is the line I meant (first line):
https://github.com/ibhagwan/fzf-lua/blob/db254a901fc3ae657e3693dd78d37d5e2ae3e945/lua/fzf-lua/providers/diagnostic.lua#L71-L79

PR #1077, opened 2024-03-07
fix(diagnostic): sign config may not have text

tomtomjhj:
For example, I have this in my config.
```
vim.diagnostic.config {
  signs = { severity = { min = vim.diagnostic.severity.WARN } },
}
```


ibhagwan:
Ty @tomtomjhj!

PR #1095, opened 2024-03-21
fix(diagnostic): buffer might have become invalid

tomtomjhj:
04d5a528a4159901607c11d46a398f45db2ed43f (#722) attempted to fix this issue, but it was not actually fixed due to the lack of `return`.

Unfortunately, I still don't have a stable reproducer to the original problem.


ibhagwan:
Ty  @tomtomjhj!

Btw, we can probably ditch the vim.schedule, Windows support added wrapping of the content handled in a `schdule_wrap`, I just haven’t gotten to testing it yet.
https://github.com/ibhagwan/fzf-lua/blob/730c27697510b46b0b1a1f417cafb2db6107a218/lua/fzf-lua/fzf.lua#L177

PR #1103, opened 2024-03-24
fix(fzf-vim): Guard check if fzf_colors is a table

jamestansx:
By default, `vim.g.fzf_colors` is not defined and `vim.tbl_map` requires a table.

### DIFF
--- a/lua/fzf-lua/profiles/fzf-vim.lua
+++ b/lua/fzf-lua/profiles/fzf-vim.lua
@@ -99,6 +99,10 @@ return {
     ["--layout"] = false,
   },
   fzf_colors = (function()
+    if type(vim.g.fzf_colors) ~= "table" then
+      return vim.g.fzf_colors
+    end
+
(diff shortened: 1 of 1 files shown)


ibhagwan:
Oversight on my end.

However, say someone defines this a a string this will cause errors, I’d rather fix it differently by changing this line:
https://github.com/ibhagwan/fzf-lua/blob/74906d0906293992d7b11744e40e5cec9131d9af/lua/fzf-lua/profiles/fzf-vim.lua#L108

To:
```lua
end, vim.g.fzf_colors or {})
```

If you agree with this can you modify your commit?
I can also do it myself but that would lose your authorship for the commit.

Let me know what you prefer?

ibhagwan:
@jamestansx, modified your commit with retained authorship, should fix the issue in a similar way.

https://github.com/ibhagwan/fzf-lua/commit/b607f4f6d9d738f5bbfdda387466e02d9bf262f8

PR #1106, opened 2024-03-28
fix: don't trigger autocmds when closing FZF window

toupeira:
This fixes a small bug when using Airline and showing buffers in the tabline, which listens to `BufUnload` to decide whether to show or hide the tabline.

For example when the tabline is configured to only appear with a minimum of 2 buffers, and exactly 2 buffers are open, the tabline disappears after closing an FZF window, because the code in Airline naively assumes that the count of buffers is reduced by 1 on every `BufUnload` [1].

[1] https://github.com/vim-airline/vim-airline/blob/d9f42cb46710e31962a9609939ddfeb0685dd779/autoload/airline/extensions/tabline/autoshow.vim#L31

### Notes

- Yes, I'll switch to `lualine.nvim` soon :sweat_smile: 
- There are a couple of other direct calls to `vim.api.nvim_buf_delete`, but I haven't run into problems with them yet. Maybe they should all use `util.nvim_buf_delete`?

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -942,7 +942,7 @@ function FzfWin:close(fzf_bufnr)
     pcall(vim.api.nvim_win_close, self.fzf_winid, true)
   end
   if self.fzf_bufnr and vim.api.nvim_buf_is_valid(self.fzf_bufnr) then
-    vim.api.nvim_buf_delete(self.fzf_bufnr, { force = true })
+    util.nvim_buf_delete(self.fzf_bufnr, { force = true })
   end
   -- when using `split = "belowright new"` closing the fzf
   -- window may not always return to the correct source win


ibhagwan:
Ty @toupeira, this looks good theoretically, however I’m not sure this will have no other effect, for example when closing the buf it also triggers closing of the window and probably other flows I’m not recalling at the moment.

ATM, this plug-in is very stable and I would hate to disturb that balance, every change in this plug-in is carefully tested and monitored.

> because the code in Airline naively assumes that the count of buffers is reduced by 1 on every BufUnload [1].

TBH, this fix seems to me like we’re forcing a workaround in fzf-lua that should be fixed in airline.

There are many other plugins currently interacting seamlessly with fzf-lua, dashboards and the like and I’m not sure if this will not have any other affects with other plugins for what is quite honestly **a bug in another plguin**.

toupeira:
@ibhagwan thanks, fair enough! I'll look into lualine then soon :grinning:

ibhagwan:
> @ibhagwan thanks, fair enough! I'll look into lualine then soon 😀

Probably a good time to move to an lua config :)

I usually wouldn’t mind this but I’ve been through quite a few “harmless” changes that ended up affecting other issues I didn’t think about.

Issue #1107, opened 2024-03-28
Builtin previewer shows empty buffer


toupeira:
### Info

- Operating System: Debian sid
- Shell: Bash
- Terminal: Tilix
- `nvim --version`: NVIM v0.9.5, LuaJIT 2.1.1700206165

- `fzf --version`: 0.48.1 (d579e33)

- [x] The issue is reproducible with `mini.sh`

### Description 

The builtin previewer just shows an empty buffer, using `bat` works fine. I don't see any error messages.

Screenshot from `mini.sh`:

![image](https://github.com/ibhagwan/fzf-lua/assets/6501/e163e8b5-9655-4b14-b7f9-3834cb14ca34)

ibhagwan:
- Is there any error in `:messages`?
- Is this error limited to `files` or also `grep`, `buffers`, `autocmds`, `help`, etc?
- Can you run `:FzfLua files debug=v` and post a screenshot of the output?
- Can you run `:FzfLua files debug=true _base64=false` and post the command in `:messages`?

toupeira:
@ibhagwan thanks, info is at the very end of this comment, but I realized now it's due to my `$FZF_DEFAULT_OPTS` :facepalm: 

Specifically `--preview-window 'right,50%,hidden,<60(up,70%,hidden)'`, without that the builtin previewer works :tada: 

This option should probably be stripped inside `fzf-lua`, since it has its own setup for the preview?

Another issue I have now is that scrolling the preview doesn't work, probably my default shortcuts from `$FZF_DEFAULT_OPTS` interfere here too?

My FZF config is at https://github.com/toupeira/dotfiles/blob/main/etc/profile.d/fzf.sh, the relevant bindings are:

```sh
  --bind ctrl-e:preview-down
  --bind ctrl-y:preview-up
  --bind ctrl-f:preview-half-page-down
  --bind ctrl-b:preview-half-page-up
```

And in my Vim config I have:

```lua
      keymap = {
        builtin = {
          ['<F1>']  = 'toggle-help',
          ['<C-_>'] = 'toggle-preview',
          ['<C-e>'] = 'preview-down',
          ['<C-y>'] = 'preview-up',
          ['<C-f>'] = 'preview-half-page-down',
          ['<C-b>'] = 'preview-half-page-up',
        },
        -- remove defaults, still uses $FZF_DEFAULT_OPTS
        fzf = {},
      },
```

[...]

PR #1118, opened 2024-04-03
fix(previewer): set initial layout for FZF previewer

toupeira:
### Summary

The FZF previewer now supports flex layout too and reacts to resize events, but we also need to set the initial layout according to the configuration.

### How to reproduce

1. Use the `bat` previewer and a custom layout, e.g.:

    ```lua
        preview = {
          default = 'bat',
          hidden = 'hidden',
          vertical = 'up:60%',
          horizontal = 'right:50%',
        },
    ```

1. Run `:FzfLua files`
1. The preview should be hidden, but is shown by default. When resizing the window the layout corrects itself.

This can also be seen when using the builtin previewer by default, and a provider that still uses the FZF previewer such as `git_status`.

### DIFF
--- a/lua/fzf-lua/previewer/fzf.lua
+++ b/lua/fzf-lua/previewer/fzf.lua
@@ -1,3 +1,4 @@
+local core = require "fzf-lua.core"
 local path = require "fzf-lua.path"
 local shell = require "fzf-lua.shell"
 local utils = require "fzf-lua.utils"
@@ -23,7 +24,7 @@ function Previewer.base:new(o, opts)
 end
 
 function Previewer.base:preview_window(_)
-  return nil
+  return core.preview_window(self.opts)
 end
 
 function Previewer.base:_preview_offset()


toupeira:
@ibhagwan first PR! :grinning: Please have a look.

This seems like the correct solution on the surface, but maybe there's a better way that I missed.

ibhagwan:
> This seems like the correct solution on the surface, but maybe there's a better way that I missed.

It does seem like the correct solution but I am struggling to reproduce the original issue this PR solves, why is your preview-window flags not set by:
https://github.com/ibhagwan/fzf-lua/blob/b5e08be6835e7ef81d0b90ac58c185e276a771cd/lua/fzf-lua/core.lua#L546-L548

Can you run `:FzfLua files debug=true` and inpsect the command in `:messages`? Mine shows the flag set:
```
--preview-window 'hidden:border:nowrap:down:45%:~3'
```

toupeira:
Oh snap, I was explicitly disabling `--preview-window` in my config so it wasn't added :facepalm: 

```lua
      fzf_opts = {
        ['--preview-window'] = false,
      },
```

Forgot I had that in there, I probably added it while debugging https://github.com/ibhagwan/fzf-lua/issues/1107. And I do use the builtin previewer now, so I only ran into the problem with `git_status`.

If I remove this everything works correctly, so probably no change is needed and I'll close the PR. Sorry for the noise, will make sure to be more thorough in the future! :wink:

ibhagwan:
Ty for the update @toupeira!

PR #1124, opened 2024-04-05
perf: avoid copying buffer/window variables to lua

tomtomjhj:
Context:
I noticed that `:FzfLua buffers` gets very slow and uses gigabytes of memory after I explore a directory with many files with [fern.vim](https://github.com/lambdalisue/fern.vim). After some investigation, I found out that
* the most of the time is spent converting vim value to lua value (`encode_vim_to_lua`)
* fzf-lua unintentionally loads buffer variables with `vim.fn.getbufinfo`
* fern stores the directory structure in buffer variable, which is cyclic due to reference to the parent node
* converting vim values containing cycles to lua value seems to exhibit time/space complexity that is quadratic on the degree of the graph node.

---

Problem:
Calling vim.fn.{getbufinfo,getwininfo} in lua can be slow because it has to convert potentially big and cyclic data structures stored in vim buffer/window variables into lua values.

Solution:
Remove buffer/window variables from the result before passing it to lua.

### DIFF
--- a/autoload/fzf_lua.vim
+++ b/autoload/fzf_lua.vim
@@ -1,0 +1,22 @@
+" Calling vim.fn.getbufinfo in lua can be expensive because it has to convert
+" all the buffer variables into lua values. Since fzf-lua does not access
+" buffer variables, this cost can be avoided by clearing the entry before
+" passing the info to lua.
+function! fzf_lua#getbufinfo(bufnr) abort
+  let info = getbufinfo(a:bufnr)
+  if empty(info)
+    return v:false " there is no way to return `nil` from vimscript
+  endif
+  unlet! info[0].variables
+  return info[0]
+endfunction
+
+" Similar to fzf_lua#getbufinfo, but for getwininfo.
+function! fzf_lua#getwininfo(winid) abort
+  let info = getwininfo(a:winid)
+  if empty(info)
+    return v:false
+  endif
+  unlet! info[0].variables
+  return info[0]
+endfunction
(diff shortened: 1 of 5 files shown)


ibhagwan:
Wow, interesting find and great fix @tomtomjhj!

Do you think we can recreate the vimL function as simple function in utils instead?

I’m in the middle of refactoring the `:FzfLua` auto-complete and trying to get rid of that specific vim file you modified,

ibhagwan:
I just read your comment why it’s vimL
```vim
" Calling vim.fn.getbufinfo in lua can be expensive because it has to convert
" all the buffer variables into lua values. Since fzf-lua does not access
" buffer variables, this cost can be avoided by clearing the entry before
" passing the info to lua.
```

Maybe we can define by streaming the function definition via `vim.cmd` or perhaps there is another command that loads vial similar to lua’s `loadstring`, I’ll do some digging.

ibhagwan:
Looks like we can source the vimL directly with `nvim_exec2`.

ibhagwan:
@tomtomjhj, merged, added https://github.com/ibhagwan/fzf-lua/commit/7aedfca54a24eae73f1a6ecfa8de0adcfc09c1a5 on top.

Additional changes:
- For some reason I don't trust the `autoload` folder so I made sure it's sourced on init
- Just in case, changed the implementation to use `utils.get{buf|win}info` which includes fallback to `vim.fn` if our functions aren't loaded
- Return for non-existing buffers windows changed to an empty list `[]` instead of `v:false` to prevent potential exceptions

PR #1127, opened 2024-04-08
Let user set a history folder for fzf action

zyd2001:
With fzf.vim, user can set a history folder with `g:fzf_history_dir`, and each action will have its separate history files in that folder. That's because the base fzf plugin use this variable to set hisotry file. With this change, people (like me) can transfer from fzf.vim to fzf-lua more seamlessly. I also set the filenames the same as fzf's history files. 

Currently I add the `--history` option at the last stage (in `fzf_exec`) as what [fzf plugin does](https://github.com/junegunn/fzf/blob/69b9d674a3842cc2a12d88268db1db7248cb61a6/plugin/fzf.vim#L419), since I'm not very familiar with the codebase. It can be better integrated.


ibhagwan:
@zyd2001, ty for the PR, merged manually as I had a few changes:

- Removed the custom names to match `fzf.vim`'s history files, IMHO it doesn't have to match fzf.vim's, these are scratch files that aren't that important
- If someone cares enough to match the history filename they can set `--history` specifically to the providers where the names aren't matching
- The name of the history file will now match that of `resume` map key, that makes it so all `grep_xxx` commands are stored under the same file in similar way the search/query is saved

> Currently I add the --history option at the last stage (in fzf_exec) as what [fzf plugin does](https://github.com/junegunn/fzf/blob/69b9d674a3842cc2a12d88268db1db7248cb61a6/plugin/fzf.vim#L419), since I'm not very familiar with the codebase. It can be better integrated.

- Moved to the correct places where all options are evaluated in `config.normalize_opts`
- `vim.fn.mkdir` is now wrapped in a `pcall`  to prevent failures (e.g. when `fzf_history_dir` points to a file)

Other changes:
- Removed debug prints `vim.print`

PR #1134, opened 2024-04-14
feat: Filter marks with regex string

alanfzf:
# Better mark filter
This PR, addresses two things one is a bug, where if you defined a mark inside your config file and it was not present in nvim at the moment of the search it would throw an error, the second one is the usage of lua patterns to match the marks and make it more user friendly to filter unwanted marks:

## The bug
```lua
-- the culprit is this call:
vim.fn.execute("marks %s")
```

image
image
example_bug

## Mark filtering
As of now the logic for filtering is to put a string of individual characters, to match only the marks you wan't, this works, but if you need a more complex solution your config file may end up like this:
image

When instead it could look like this:
image

And finally here's an example of the feature in action:

Screenshot 2024-04-14 161536

### DIFF
--- a/README.md
+++ b/README.md
@@ -1280,6 +1280,12 @@ require'fzf-lua'.setup {
     -- severity_limit:  keep any equal or more severe (lower)
     -- severity_bound:  keep any equal or less severe (higher)
   },
+  marks = {
+    marks = "", -- filter vim marks with a lua pattern
+    -- for example if you want to only show user defined marks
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @alanfzf, this makese sense, all of the other fzf-lua filters are lua string patterns.

The old filter was a contribution from another user and TBH I don't use marks myself so it didn't make a huge difference for me.

alanfzf:
Sure thing glad i could help @ibhagwan 😄!

PR #1143, opened 2024-04-19
fix(command): allow running the command in visual mode. (closes #1142)

### LINKED ISSUE #1142: Bug: `grep-visual` command is broken
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

arch linux

### Shell

fish 

NVIM v0.10.0-dev-2954+g13ebfafc95

0.48.1 (9ffe951f)

nil

- [X] My issue is reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

If i use visual mode to select and then run `:'<,'>FzfLua grep_visual` , nvim will report that `E481: No range allowed`. I think it's a bug in https://github.com/ibhagwan/fzf-lua/blob/52637a72ed728d7955a6a2b24afdf455ebaba623/plugin/fzf-lua.lua#L13-L17, which the `range` parament is false by default.

Same, this bug occurs when we are running 'tags_grep_visual`

XXiaoA:


ibhagwan:
Ty @XXiaoA!

PR #1150, opened 2024-04-26
feat(previewer): support all preview scroll actions from FZF

toupeira:
This adds the `preview-(up|down)`, `preview-page-(up|down)`, and `preview-(top|bottom)` actions from FZF, and renames the existing page scroll actions to `preview-half-page-(up|down)`.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -331,7 +331,7 @@ function Previewer.base:scroll(direction)
   if preview_winid < 0 or not direction then return end
   if not api.nvim_win_is_valid(preview_winid) then return end
 
-  if direction == 0 then
+  if direction == 'reset' then
     pcall(vim.api.nvim_win_call, preview_winid, function()
       -- for some reason 'nvim_win_set_cursor'
       -- only moves forward, so set to (1,0) first
@@ -342,14 +342,24 @@ function Previewer.base:scroll(direction)
       utils.zz()
     end)
   elseif not self:preview_is_terminal() then
-    -- local input = direction > 0 and [[]] or [[]]
-    -- local input = direction > 0 and [[]] or [[]]
-    -- ^D = 0x04, ^U = 0x15 ('g8' on char to display)
-    local input = ("%c"):format(utils._if(direction > 0, 0x04, 0x15))
-    pcall(vim.api.nvim_win_call, preview_winid, function()
-      vim.cmd([[norm! ]] .. input)
-      utils.zz()
-    end)
+    -- map direction to scroll commands ('g8' on char to display)
+    local input = ({
+      ['top']            = 'gg',
+      ['bottom']         = 'G',
+      ['half-page-up']   = ('%c'):format(0x15), -- [[]]
+      ['half-page-down'] = ('%c'):format(0x04), -- [[]]
+      ['page-up']        = ('%c'):format(0x02), -- [[]]
+      ['page-down']      = ('%c'):format(0x06), -- [[]]
+      ['line-up']        = 'Mgk', -- ^Y doesn't seem to work
+      ['line-down']      = 'Mgj', -- ^E doesn't seem to work
+    })[direction]
+
+    if input then
+      pcall(vim.api.nvim_win_call, preview_winid, function()
+        vim.cmd([[norm! ]] .. input)
+        utils.zz()
+      end)
+    end
   else
     -- we get here when using custom term commands using
     -- the extensions map (i.e. view term images with 'vui')
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -21,13 +21,19 @@ setmetatable(FzfWin, {
 })
 
 local _preview_keymaps = {
-  ["toggle-preview"]      = { module = "win", fnc = "toggle_preview()" },
-  ["toggle-preview-wrap"] = { module = "win", fnc = "toggle_preview_wrap()" },
-  ["toggle-preview-cw"]   = { module = "win", fnc = "toggle_preview_cw(1)" },
-  ["toggle-preview-ccw"]  = { module = "win", fnc = "toggle_preview_cw(-1)" },
-  ["preview-page-up"]     = { module = "win", fnc = "preview_scroll(-1)" },
-  ["preview-page-down"]   = { module = "win", fnc = "preview_scroll(1)" },
-  ["preview-page-reset"]  = { module = "win", fnc = "preview_scroll(0)" },
+  ["toggle-preview"]         = { module = "win", fnc = "toggle_preview()" },
+  ["toggle-preview-wrap"]    = { module = "win", fnc = "toggle_preview_wrap()" },
+  ["toggle-preview-cw"]      = { module = "win", fnc = "toggle_preview_cw(1)" },
+  ["toggle-preview-ccw"]     = { module = "win", fnc = "toggle_preview_cw(-1)" },
+  ["preview-up"]             = { module = "win", fnc = "preview_scroll('line-up')" },
+  ["preview-down"]           = { module = "win", fnc = "preview_scroll('line-down')" },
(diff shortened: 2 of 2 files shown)


lua/fzf-lua/previewer/builtin.lua
```diff
@@ -331,25 +331,25 @@ function Previewer.base:scroll(direction)
   if preview_winid < 0 or not direction then return end
   if not api.nvim_win_is_valid(preview_winid) then return end
 
-  if direction == 0 then
-    pcall(vim.api.nvim_win_call, preview_winid, function()
-      -- for some reason 'nvim_win_set_cursor'
-      -- only moves forward, so set to (1,0) first
-      api.nvim_win_set_cursor(0, { 1, 0 })
-      if self.orig_pos then
-        api.nvim_win_set_cursor(0, self.orig_pos)
```
toupeira: **note:** I'm not sure if jumping to `self.orig_pos` is still needed, or how this works exactly. Just figured using `G` would be the easiest way to implement `preview-bottom` :grinning:

lua/fzf-lua/win.lua
```diff
@@ -21,13 +21,19 @@ setmetatable(FzfWin, {
...
-  ["preview-page-down"]   = { module = "win", fnc = "preview_scroll(1)" },
-  ["preview-page-reset"]  = { module = "win", fnc = "preview_scroll(0)" },
+  ["toggle-preview"]         = { module = "win", fnc = "toggle_preview()" },
+  ["toggle-preview-wrap"]    = { module = "win", fnc = "toggle_preview_wrap()" },
+  ["toggle-preview-cw"]      = { module = "win", fnc = "toggle_preview_cw(1)" },
+  ["toggle-preview-ccw"]     = { module = "win", fnc = "toggle_preview_cw(-1)" },
+  ["preview-up"]             = { module = "win", fnc = "preview_scroll('line-up')" },
+  ["preview-down"]           = { module = "win", fnc = "preview_scroll('line-down')" },
+  ["preview-page-up"]        = { module = "win", fnc = "preview_scroll('page-up')" },
+  ["preview-page-down"]      = { module = "win", fnc = "preview_scroll('page-down')" },
```
toupeira, line 31: **note:** This is a breaking change, we could keep these mapped to `half-page-(up|down)`, or change the defaults in `defaults.lua` and the profiles.

We could also change the default Shift+Up/Down shortcuts to `preview-(up|down)` now to match FZF's defaults.

toupeira (commented):
Hey @ibhagwan, finally got around to this PR :grinning: 

What do you think? Let me know if you want any changes to the docs and the defaults.

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -331,25 +331,25 @@ function Previewer.base:scroll(direction)
...
+    -- map direction to scroll commands ('g8' on char to display)
+    local input = ({
+      ['top']            = 'gg',
+      ['bottom']         = 'G',
+      ['half-page-up']   = ('%c'):format(0x15), -- [[]]
+      ['half-page-down'] = ('%c'):format(0x04), -- [[]]
+      ['page-up']        = ('%c'):format(0x02), -- [[]]
+      ['page-down']      = ('%c'):format(0x06), -- [[]]
+      ['line-up']        = 'Mgk', -- ^Y doesn't seem to work
+      ['line-down']      = 'Mgj', -- ^E doesn't seem to work
```
toupeira, line 354: **note:** I haven't figured out why ^Y/^E don't work, they get passed through but somehow don't work in the preview window. Doing `:norm! ^E` elsewhere works fine though.

Also not quite sure why using `M` works, I tried `H` and `L` at first but those scrolled too far. With `M` it seems to work correctly, but we could also switch to using `nvim_win_set_cursor`.

ibhagwan:
Ty @toupeira! I’ll review over the weekend and share my thoughts, looks like a great addition.

lua/fzf-lua/previewer/builtin.lua
ibhagwan: This was used to bring you back to the original cursor line (usually the match/symbol/etc) after scrolling away from it, by default it was mapped to `shift-left` by the action name `preview-page-reset`.

toupeira: Oh right so it's an actual reset, I somehow assumed it was just moving to the top too :facepalm: 

Added this back now with `direction = 'reset'`.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/fcd0853770a7ef2965e2f1a8c0f1e03ad9aaea62

Ty @toupeira! Rebase-merged manually as `lua_ls` was complaining about the usage of single quotes.

ibhagwan:
I really like the line up/down variants, I'm thinking perhaps we should map `shift-{left|right}` to these as a default, what do you think?

toupeira:
@ibhagwan thanks for merging! Not sure if that was accidental, just double-checking if the breaking change is okay? :grinning: 

I don't have a strong opinion on the defaults, happy to go with whatever you prefer. I saw Shift-Left is already mapped to `preview-page-reset` though.

Also note that FZF has these default mappings:

- `shift-left`: backward-word
- `shift-right`: forward-word
- `shift-up`: preview-up
- `shift-down`: preview-down

I personally use Ctrl-e/y mappings for up/down.

ibhagwan:
> Not sure if that was accidental, just double-checking if the breaking change is okay? 😀

Not accidental, as I mentioned above somewhere I don’t consider this breaking (nothing breaks for anyone) the scroll just behaves a tiny bit different :)

> I don't have a strong opinion on the defaults, happy to go with whatever you prefer. I saw Shift-Left is already mapped to preview-page-reset though.

Correct, but I think it would not be a big deal to change, we could also find an alternative default bind for this.

> shift-left: backward-word

Wasn’t aware of the backward/forward word mappings, maybe we should consider different mappings altogether for preview scrolls?

> I personally use Ctrl-e/y mappings for up/down.

`ctrl-e` would override the fzf’s readline defaults (end of line).

toupeira:
@ibhagwan ok! :+1: 

What do you think about Shift-PageUp/Down for `preview-page-(up|down)`, and Shift-Up/down for linewise `preview-(up|down)`?

ibhagwan:
> @ibhagwan ok! 👍
> 
> What do you think about Shift-PageUp/Down for `preview-page-(up|down)`, and Shift-Up/down for linewise `preview-(up|down)`?

I think it’s a great idea, I’m just concerned about laptop keyboard which do not have PageUp/PageDown keys, can we input these with a key combo?

toupeira:
@ibhagwan laptops and other smaller keyboards should usually have PageUp/Down with a `Fn` combo at least. This one does for example and I can use Shift+Fn+Up/Down for PageUp/Down, which would match nicely with Shift+Up/Down.

I'll still remap it to something more convenient (I use Ctrl-f/b for page up/down), but I think as a default it would work well :grinning:

toupeira:
Hmm what would be neat is having PageUp/Down directly mapped (no Shift) to scroll the preview only if it's visible, and mapped to scroll the FZF results otherwise.

ibhagwan:
> Hmm what would be neat is having PageUp/Down directly mapped (no Shift) to scroll the preview only if it's visible, and mapped to scroll the FZF results otherwise.

Idk, I feel too many people already use the page up/down for scrolling the results.

ibhagwan:
@toupeira, with https://github.com/ibhagwan/fzf-lua/commit/7a2d9849fe6ee87a386a8b9486ebd2e2884184fa I removed the default `shift-left` as it overides the default fzf word left/right, if someone is missing this functionality they can add it in their setup.

I also added new defaults for preview (line) up/down with `alt-shift-{up|down}`, I avoided using `shift-pageup|pagedown` as these keys aren't supported by fzf and I prefer to maintain uniformity in functaionlity where possible, the alt-shift variant is supported in both fzf and neovim.

toupeira:
@ibhagwan thanks, works for me! :grinning:

PR #1163, opened 2024-04-30
Prevent issues if cmdline_cfg is nil

MrLys:
Not sure how/when this happens, but it kept happening on a relatively
  uncustomized version of LazyVim


ibhagwan:
Ty @MrLys, merged in https://github.com/ibhagwan/fzf-lua/commit/bb421160c790e00483789651d5ce7f3431103227.

Issue #908, opened 2023-10-16
FzfLua freezes if oldfiles contains a named pipe


mikesmithgh:
### Info

- Operating System: macos
- Shell: bash
- Terminal: Kitty 
- `nvim --version`:
```
NVIM v0.10.0-dev-1364+ga4c4b39d5-dirty
Build type: RelWithDebInfo
LuaJIT 2.1.1696795921
Run "nvim -V1 -v" for more info
```
- `fzf --version`: `0.43.0 (brew)`

- [x] The issue is reproducible with `mini.sh`

Download [testshada.txt](https://github.com/ibhagwan/fzf-lua/files/12920947/testshada.txt) (this contains a file marker on /tmp/named_pipe causing it to be added to `oldfiles`)

```
git checkout 949e9f8
mkfifo /tmp/named_pipe
NVIM='nvim -i testshada.txt' scripts/mini.sh
:FzfLua oldfiles
```
![FzfLua oldfiles](https://github.com/ibhagwan/fzf-lua/assets/10135646/010bba8d-64d4-4d95-913e-dc4c11a6be94)

### Description 

Ok I have a weird one for you 😅

The change on https://github.com/ibhagwan/fzf-lua/commit/949e9f82ccc5431b3fc4e94834dfdbc2c3a1ac3e to check if a file is readable causes oldfiles to freeze if there exists a named pipe in `vim.v.oldfiles`.

I also tested older versions before the commit, and it doesn't freeze initially. But, it does freeze when Neovim attempts to close.

This is a really rare issue that is not likely to occur, I just found it while I was experimenting with named pipes.

### Workaround
Delete the old named pipe or remove the shada file.

mikesmithgh:
@ibhagwan so I think what happened is I opened a file with Neovim `/tmp/named_pipe` for example. Then, I removed it and created it as a named pipe with `mkfifo /tmp/named_pipe`. 

I was playing around trying to see if I could write to a named pipe and read from it inside Neovim. But, gave up and had the left over files hanging out 😂 Users will find a way to break things

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/416b4745dbcfbba92fd9d29cad3f30ece6a6a4f3

@mikesmithgh - should work now.

mikesmithgh:
> [416b474](https://github.com/ibhagwan/fzf-lua/commit/416b4745dbcfbba92fd9d29cad3f30ece6a6a4f3)
> 
[...]

PR #1205, opened 2024-05-20
ref(oldfiles): simpler+faster file-check alternative

JosefLitos:
`vim.loop.fs_stat` is many times slower than just trying to open a file. So I added a backward-compatible `check_file` option to `oldfiles` to do just that. But since I'm not sure if the extra `fs_stat` checks have some extra benefits, I kept it as the default.

Also I have no idea how to format the files, so I had to do everything manually, since stylua would change half of the things because the files don't follow the `.stylua.toml` rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1006,7 +1006,9 @@ require'fzf-lua'.setup {
   oldfiles = {
     prompt            = 'History❯ ',
     cwd_only          = false,
-    stat_file         = true,         -- verify files exist on disk
+    -- "stat": verify files exist on disk and are usable files
+    -- true: just check file can be opened
+    check_file        = "stat",
     include_current_session = false,  -- include bufs from current session
   },
   buffers = {
--- a/README.md
+++ b/README.md
@@ -1006,9 +1006,7 @@ require'fzf-lua'.setup {
   oldfiles = {
     prompt            = 'History❯ ',
     cwd_only          = false,
-    -- "stat": verify files exist on disk and are usable files
-    -- true: just check file can be opened
-    check_file        = "stat",
+    stat_file         = true,         -- verify files exist on disk (or fun() for custom check)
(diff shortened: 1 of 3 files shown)


ibhagwan:
Ty @JosefLitos!

A few changes:
- Instead of over complicating with backward compat, we should leave the option named as is `stat_file`, none of the users care what happens behind the scenes if it actually does stat or not as long as the entires are readable files
- Using the new method you’d still have the rare issue when opening  a named pipe (#908)
- Overall I feel that having yet 2 different ways for detecting if an old file is valid is an over complication of the issue

I have another suggestion, what if we make it so `stat_file` can also accept a `function` and then you can set it up in your config as:
```lua
      stat_file = function(file)
        local handle = io.open(file, "r")
        if handle then handle:close() end
        return handle ~= nil
      end
``` 

> Also I have no idea how to format the files, so I had to do everything manually, since stylua would change half of the things because the files don't follow the .stylua.toml rules.

Although I have `sylua.toml` I’m still using `lua_ls` to do the formatting.

JosefLitos:
Changes made. I reverted back the option name and allowed a custom function.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/a32d807bf071a7a537c485f10a2d622847a2bc70

Ty @JosefLitos, merged manually as I made some tiny changes (which I was already in the process of :-))

You can now achieve what you want with:
```lua
:lua require("fzf-lua").oldfiles({ stat_file=require("fzf-lua").utils.file_is_readable })
```

You can test disabling the filter with:
```lua
-- Both are the same result
:lua require("fzf-lua").oldfiles({ stat_file=false })
:lua require("fzf-lua").oldfiles({ stat_file=function() return true end })
```

PR #1216, opened 2024-05-25
fix: restrict the buffer on autocmds created with `FzfWin:_nvim_create_autocmd`

idanarye:


ibhagwan:
Ty @idanarye!

Total oversight in my end, I had it in my head and forgot to add it lol

PR #1217, opened 2024-05-26
fix: use proper annotations for public APIs

uga-rosa:
Hi, Thanks for a great plugin!

I've been customizing it based on [this page](https://github.com/ibhagwan/fzf-lua/wiki/Advanced), but the annotations in `fzf_exec` seem to be wrong and I want to fix it.
I get ls warnings for fields that should be omissible.

image

Also, `fzf_live` does not have annotation, so I need to add it.

However, I am concerned that opts is inherently a more complex table.
Should I do it this way? I'm willing to do that work, but where is the list of all the fields?

```lua
---@class fzf_exec_opts
---@field cwd? string
---@field prompt? string
...
```

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -116,12 +116,12 @@ local contents_from_arr = function(cont_arr)
   return contents
 end
 
----@alias content table|function|string|nil
+---@alias content string[]|fun(fzf_cb: fun(entry?: string))|string|nil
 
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tysm @uga-rosa! This is an area I could rally use some help as I never get the time for proper annotations.

uga-rosa:
The type was incorrect when contents is a function. I will PR again after some more research.

PR #1218, opened 2024-05-26
fix: annotaion for fzf_exec

uga-rosa:
If my perception is wrong, please point it out.

- An entry is string or number (just need to be able to apply `table.concat()`).
- `fzf_cb` can receive a callback function as a second argument (optional).


ibhagwan:
Ty @uga-rosa!

PR #1285, opened 2024-06-19
fix: when bind is nil

phanen:
To reproduce the bug:
```lua
require("fzf-lua").setup { keymap = { fzf = {} } }
```
then `:FzfLua builtin<cr>` cannot open the float windows.


ibhagwan:
Ty @phanen!

PR #1313, opened 2024-07-02
readme: use underscore for emphasis

Futarimiti:
`*after*` would misguide `:helptags` to pick it up as a tag e.g. `:h after`.


ibhagwan:
Ty @Futarimiti!

PR #1370, opened 2024-07-31
fix(previewer): respect zindex

xactlyblue:
This fixes the previewer not respecting the provided zindex, leading to the main window working as intended while the previewer occluded other windows that overlap it (for example, in my case at least, the ``noice.nvim`` command popup was being mostly occluded by the previewer window with only a small portion being visible).

I made these changes pretty quickly and I'm not really familiar with the code-base, so I'd appreciate if someone with more knowledge on how the project works could make sure these changes won't cause any weird edge-cases or visual bugs. I don't think it'll cause any issue but I thought it still worth mentioning.


ibhagwan:
Merged, ty @xactlyblue!

PR #1378, opened 2024-08-03
fix(help-previewer): don't error if no table is found

TheLeoP:
If no tag is found (e.g. the user typed a prompt that no longer shows any results), the following line throws an error

https://github.com/TheLeoP/fzf-lua/blob/af659d93df1b110967f5a41e86b969fbabf858a7/lua/fzf-lua/previewer/builtin.lua#L980

Instad, this PR returns an empty table early if no tag is found


ibhagwan:
Ty @TheLeoP!

PR #1381, opened 2024-08-06
feat(lsp): add `unique_line_items` options

phanen:
For example, use `require("fzf-lua").lsp_references`
with `opts.unique_line_items = true`:
image

with `opts.unique_line_items = false`:
image

> This can also be a workaround for https://github.com/LuaLS/lua-language-server/issues/2451.


ibhagwan:
Tysm @phanen, simple and clean :)

PR #1389, opened 2024-08-09
feat(commands): add `include_builtin`

phanen:
This is a lua port of `Commands` behavior in `fzf.vim`: https://github.com/junegunn/fzf.vim/blob/6f28c8c7bb551161a0315a76488522204f39c1f4/autoload/fzf/vim.vim#L1212-L1236.

Maybe this is better than `vim.fn.getcompletion('', 'command')` because it also provides brief descriptions, and keep updated with latest document shipped with neovim installation.

### DIFF
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -13,7 +13,27 @@ M.commands = function(opts)
 
   local global_commands = vim.api.nvim_get_commands {}
   local buf_commands = vim.api.nvim_buf_get_commands(0, {})
-  local commands = vim.tbl_extend("force", {}, global_commands, buf_commands)
+
+  local builtin_commands = {}
+  -- parse help doc to get builtin commands and descriptions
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tysm @phanen, merged manually (not sure why GH didn' t pick this up), added https://github.com/ibhagwan/fzf-lua/commit/c2a0044c09e42cf43e7aa77994cce36fe846b80c on top:

- Set `include_builtin=true` as default
- Match command colors with fzf.vim
- Added missing `Commands` to fzf.vim profile

ibhagwan:
Oh, perhaps it's also  time to format the commands properly, something in between fzf.vim's and Telescope?

fzf.vim:
image

Telescope:
image

I'm also thinking to add a "builtin" previewer for this so for commands that map to a lua function (e.g. `Man`), we can display the lua function similar to what is done with `:FzfLua autocmds`?

In any event I'm in no rush on this as this is falls in the "nice to have" category.

PR #1405, opened 2024-08-15
fix(oldfiles): include unlisted buffers with `include_current_session`

toupeira:
This allows reopening buffers that were first opened in the current session, and then closed again.

Steps to reproduce:

1. Open a buffer which doesn't appear in `oldfiles` yet
2. Close it with `:bd`
3. Run `:FzfLua oldfiles` (with `include_current_session` enabled)
4. Previously the closed buffer wouldn't show up, but now it does! :tada:


ibhagwan:
Ty @toupeira!

PR #1409, opened 2024-08-17
refactor: use confirm instead of input where it makes sense

tssm:
Hey @ibhagwan!

I planned to open and issue to discuss this, but it was a simple enough change that I decided to just do it. I don't know if you used `input` deliberately instead of `confirm`, but I think overall `confirm` is in the changed cases. It requires one less keystroke, and it looks nicer if you use a GUI or TUI like Noice. I kept the `Yes`/`No` combo, so there's no need to learn new keystrokes.

And while I was here, I also wrapped the `drop` action with confirm, as it's destructive (even when most of the time you can recover things using the reflog)

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -719,21 +719,25 @@ M.git_stage_unstage = function(selected, opts)
 end
 
 M.git_reset = function(selected, opts)
-  for _, s in ipairs(selected) do
-    s = utils.strip_ansi_coloring(s)
-    local is_untracked = s:sub(5, 5) == "?"
-    local cmd = is_untracked
-        and path.git_cwd({ "git", "clean", "-f" }, opts)
-        or path.git_cwd({ "git", "checkout", "HEAD", "--" }, opts)
-    git_exec({ s }, opts, cmd)
-    -- trigger autoread or warn the users buffer(s) was changed
-    vim.cmd("checktime")
+  if vim.fn.confim("Reset " .. #selected .. " file(s)?", "&Yes\n&No") == 1 then
+    for _, s in ipairs(selected) do
+      s = utils.strip_ansi_coloring(s)
+      local is_untracked = s:sub(5, 5) == "?"
+      local cmd = is_untracked
+          and path.git_cwd({ "git", "clean", "-f" }, opts)
+          or path.git_cwd({ "git", "checkout", "HEAD", "--" }, opts)
+      git_exec({ s }, opts, cmd)
+      -- trigger autoread or warn the users buffer(s) was changed
+      vim.cmd("checktime")
+    end
   end
 end
 
 M.git_stash_drop = function(selected, opts)
-  local cmd = path.git_cwd({ "git", "stash", "drop" }, opts)
-  git_exec(selected, opts, cmd)
+  if vim.fn.confirm("Drop " .. #selected .. " stash(es)?", "&Yes\n&No") == 1 then
+    local cmd = path.git_cwd({ "git", "stash", "drop" }, opts)
+    git_exec(selected, opts, cmd)
+  end
 end
 
 M.git_stash_pop = function(selected, opts)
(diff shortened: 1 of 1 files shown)


ibhagwan:
> I planned to open and issue to discuss this, but it was a simple enough change that I decided to just do it. I don't know if you used input deliberately instead of confirm, but I think overall confirm is in the changed cases. It requires one less keystroke, and it looks nicer if you use a GUI or TUI like Noice. I kept the Yes/No combo, so there's no need to learn new keystrokes.

TBH, I wasn’t aware of confirm at the time but the intention is to use the new API `vim.ui.input` so it can be “dressed” by other plugins (such as dressing, etc), there was some issue with it so it’s currently only taking the `vin.fn.input` instead, this probably needs a revisit to see if it can be re-enabled:
https://github.com/ibhagwan/fzf-lua/blob/40904abf2bdd784aa789515ce3cd0bc267e5eff4/lua/fzf-lua/utils.lua#L1120-L1130

As for `vim.fn.confirm`, not sure but I think the same behavior (one key confirm) can be achieved with `vim.ui.input`.

Another thing I noticed in your commit is your replaced `utils.input` with `utils.confirm` (non-existent).

> And while I was here, I also wrapped the drop action with confirm, as it's destructive (even when most of the time you can recover things using the reflog)

I agree, this is a welcome change, I’m thinking this also applies to `git_reset` action in `git_status`?

tssm:
> the intention is to use the new API vim.ui.input so it can be “dressed” by other plugins (such as dressing, etc)

`confirm` gets dressed by Noice, so I guess the authors of dressing and others just forgot it exist. I don't think `input` is mean to replace `confirm`, they serve different purposes. For example, you can `set confirm` to be always shown when you may discard changes with `:q` and such.

> Another thing I noticed in your commit is your replaced utils.input with utils.confirm (non-existent).

This was overconfidence on my side, I didn't test all the changes 🙈

> I’m thinking this also applies to git_reset action in git_status?

Yeah, I can wrap those too. Is it OK to force-push?

ibhagwan:
> Yeah, I can wrap those too. Is it OK to force-push?

sure

ibhagwan:
Looks great @tssm, I’ll test this tomorrow when I’m no longer AFK.

tssm:
Cool! I was preparing a branch just with wrapping the actions with `input`, but I would like for you to have a look at it using `confirms` as I think it's the right function for this case.

But yeah, I already have a branch only adding `utils.input` to the actions that didn't have it, and I can force push it here

ibhagwan:
> Cool! I was preparing a branch just with wrapping the actions with input, but I would like for you to have a look at it using confirms as I think it's the right function for this case.

It sure does look like the right tool for the job, as long as if also works seamlessly with dressing this should be merged.

> But yeah, I already have a branch only adding utils.input to the actions that didn't have it, and I can force push it here

If you wish for these changes to be part of this PR no problem.

tssm:
In that case, I think you will ask me to revert the first commit, and modify the second, as it doesn't seem like dressing does anything to `confirm`. However, [this PR](https://github.com/neovim/neovim/pull/29310), which handles a specific confirm prompt, suggest to me that all the others (including the `confirm` function) can be managed by `vim.ui_attach`, which is what Noice does to make them floating windows. I wonder if dressing should add support instead?

Anyway, if you prefer to keep using `input`, maybe [this function](https://github.com/ibhagwan/fzf-lua/blob/40904abf2bdd784aa789515ce3cd0bc267e5eff4/lua/fzf-lua/utils.lua#L897) should be updated for consistency

ibhagwan:
> In that case, I think you will ask me to revert the first commit, and modify the second, as it doesn't seem like dressing does anything to confirm. However, https://github.com/neovim/neovim/pull/29310, which handles a specific confirm prompt, suggest to me that all the others (including the confirm function) can be managed by vim.ui_attach, which is what Noice does to make them floating windows. I wonder if dressing should add support instead?

Confirm does feel more natural in this situation and since we’re not using `vim.ui.input` I think we can safely change it to `vim.fn.confirm`.

> Anyway, if you prefer to keep using input, maybe [this function](https://github.com/ibhagwan/fzf-lua/blob/40904abf2bdd784aa789515ce3cd0bc267e5eff4/lua/fzf-lua/utils.lua#L897) should be updated for consistency

The reason this was done like that is to replicate the built in save dialog when `vim.o.confirm = true`.

PR #1415, opened 2024-08-25
fix: file_switch_or_edit with nill

kennethjang34:
Hi, I noticed when I have a buffer list open with a single buffer, calling file_switch_or_edit buffer results in some error due to the passed arg doesn't have any selected buffer.
I added a line to prevent file_switch function from producing error and instead to simply return false

Thank you,


ibhagwan:
Ty @kennethjang34

PR #1420, opened 2024-08-28
fix(codeactions): handle preview when no match buffer

phanen:


ibhagwan:
Unless used with the wrong picker, how can this happen?

nenahp:
To reproduce, after lanching fzf's term popup, type something to filter the selections into empty.
`:FzfLua lsp_code_actions previewer='codeaction'`

```lua
Error executing vim.schedule lua callback: ...e/nvim/lazy/fzf-lua/lua/fzf-lua/previewer/codeaction.lua:249: assertion failed!
stack traceback:
        [C]: in function 'assert'
        ...e/nvim/lazy/fzf-lua/lua/fzf-lua/previewer/codeaction.lua:249: in function 'populate_preview_buf'
        ...hare/nvim/lazy/fzf-lua/lua/fzf-lua/previewer/builtin.lua:263: in function 'populate_preview_buf'
        ...hare/nvim/lazy/fzf-lua/lua/fzf-lua/previewer/builtin.lua:280: in function ''
        vim/_editor.lua: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

But this not happend with:
`:FzfLua lsp_code_actions previewer='codeaction_native'`

phanen:
It seems that the `entry_str` is from `items[1]` here.
But when no entry is selected, it would be a empty string.
https://github.com/ibhagwan/fzf-lua/blob/a5283dfbf11d78a84796692a378a3701d7f9797d/lua/fzf-lua/previewer/builtin.lua#L288-L295.

So maybe it would be better to be fixed here.

ibhagwan:
> It seems that the entry_str is from items[1] here.
> But when no entry is selected, it would be an empty string.

I actually prefer the previous fix, what if someone wants to implement a callback for when there is no entry? IMHO if should be handled in the previewer code.

phanen:
Make senses, maybe better to fix in specific picker.

ibhagwan:
Tysm @phanen!

PR #1427, opened 2024-09-04
fix(jumps): when no match items

phanen:
Fix https://github.com/ibhagwan/fzf-lua/pull/1419#issuecomment-2328157264


ibhagwan:
Ty @phanen!

Issue #1443, opened 2024-09-12
Bug: Files with % aren't displayed in the preview pane when using grep-like commands


lgrosz:
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [X] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Ubuntu Linux

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.11.0-dev-131+gf55767afe

### Fzf version (`fzf --version`)

0.42.0 (d471067)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=bg+:#414559,bg:#303446,spinner:#f2d5cf,hl:#e78284 --color=fg:#c6d0f5,header:#e78284,info:#ca9ee6,pointer:#f2d5cf --color=marker:#f2d5cf,fg+:#c6d0f5,prompt:#ca9ee6,hl+:#e78284

### Is the problem reproducible with `mini.sh`?

- [X] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({ })
```

## Reproduction steps

1. Create a file, `my_%_file` with some searchable content
2. Use `:FzfLua live_grep` to search for the content

## Behavior

The preview window says "cannot stat file..." with the `%` removed from the filename

## Expected behavior

The preview window should display as if the filename didn't have a `%` in it

## Additional Information

- This only happens with grep-like command, not commands like `:FzfLua files`
- I am using `rg`

hagnerd:
This does work as expected with `FzfLua live_grep_native` so I assume the `live_grep` implementation needs to do something to make sure that any `%` is escaped properly.

More info on [escape patterns](https://www.lua.org/pil/20.2.html) in Lua .

ibhagwan:
You’re correct it’s an easy fix I just haven’t gotten to it yet.

ibhagwan:
Very confusing, this doesn't seem to reproduce, I added a debug print right after the previewer `fs.stat` and it seems to work without additional escaping:
![image](https://github.com/user-attachments/assets/394ec823-edf8-405c-800b-f46342ef319a)


[...]

PR #1445, opened 2024-09-14
fix: escape "%" in gsub target

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1443.

### DIFF
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -400,7 +400,8 @@ M.file = function(x, opts)
   else
     ret[#ret + 1] = file_is_ansi > 0
         -- filename is ansi escape colored, replace the inner string (#819)
-        and file_part:gsub(utils.lua_regex_escape(stripped_filepath), filepath)
+        -- escape `%` in path, since `string.gsub` also use it in target (#1443)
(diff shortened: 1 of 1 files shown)


ibhagwan:
If doing this change we need to make sure we’re covering all lua escape chars, we can use `utils.lua_regex_escape`:
https://github.com/ibhagwan/fzf-lua/blob/985e1002b5dc70f8196ea0605bbfffc8c8407e31/lua/fzf-lua/utils.lua#L231-L240

ibhagwan:
Just realized this happens due to the replacement string escaping and not the input string, the `%`  is probably all that’s needed.

ibhagwan:
I knew I encountered this before :)
https://github.com/ibhagwan/fzf-lua/blob/985e1002b5dc70f8196ea0605bbfffc8c8407e31/lua/fzf-lua/core.lua#L1254-L1255

Ty @phanen!

ibhagwan:
Looking at the code I just posted shouldn’t the replacement string gsub be 4x`%`?

ibhagwan:
Fixed it with a force push https://github.com/ibhagwan/fzf-lua/commit/f513524561060f2b9e3bd6d36ff046bfa03ca114.

nenahp:
Should be `4%`, I got into the second trap.

PR #1446, opened 2024-09-14
Update fzf minimum requirement

CowKeyMan:
git_commits and git_bcommits does not work with fzf v0.29. Not sure exactly which version will be best, but I set it to latest to be safe


ibhagwan:
Ty @CowKeyMan but I don’t think we need to update the min requirements for 2 pickers which may be fixed if we really wanted to.

PR #1448, opened 2024-09-20
feat(file_ignore_globs): maybe a replacement of `file_ignore_patterns` for `rg`/`fd`

phanen:
* `file_ignore_patterns` use lua patterns and need to be transformed entry by entry, especially when all other transform options(e.g. `file_icons/git_icons`) turn off.
* `file_ignore_globs` will be applied as `rg -iglob "!glob1" -iglob "!glob2" ...` and `fd --exclude "glob1" --exclude "glob2" ...`.
* Both `file_ignore_{patterns,globs}` can be a function now.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -267,14 +267,12 @@ function M.normalize_opts(opts, globals, __resume_key)
 
   -- Merge arrays from globals|defaults, can't use 'vim.tbl_xxx'
   -- for these as they only work for maps, ie. '{ key = value }'
-  for _, k in ipairs({ "file_ignore_patterns" }) do
+  for _, k in ipairs({ "file_ignore_patterns", "file_ignore_globs" }) do
     for _, m in ipairs({ globals, M.globals }) do
-      if m[k] then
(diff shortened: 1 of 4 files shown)


ibhagwan:
Ty @phanen for the work but I wouldn’t like to merge this, although this might be more convenient in your use case I find this counterintuitive and confusing to use:

- Hides the fact this adds globs to the `rg` / `fd` commands (confusing)
- Will not work if rg or fd aren’t installed (files defaults to `find` on Unix and `dir` on Windows)
- Will not work on all other pickers supported by `file_ignore_patterns` (old files, LSP, etc)
-  Patterns uses lua style regex, globs will use rg globs when using rg and fd globs when using fd and none when no command is used - good luck explaining that to users :)

phanen:
Didn't think about it that much before, thank you for pointing it out.

BTW, just found this to be a pseudo-requirement, as I could easily use .rgignore/.fdignore to achieve that. (:

ibhagwan:
> BTW, just found this to be a pseudo-requirement, as I could easily use .rgignore/.fdignore to achieve that. (:

You’re right I didn’t think about that either.

PR #1449, opened 2024-09-20
perf(escape): elimite callback

phanen:
No callback escape.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -228,15 +228,13 @@ function M.lua_escape(str)
   end)
 end
 
+---@see vim.pesc
 function M.lua_regex_escape(str)
   -- escape all lua special chars
   -- ( ) % . + - * [ ? ^ $
   if not str then return nil end
   -- gsub returns a tuple, return the string only or unexpected happens (#1257)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @phanen.

This function isn’t used in a performance critical chain (like for every entry in `make_entry`), is there a real impact on performance here using `%1` instead of a function? If this is material this wouldn’t be the only location in the code where this is happening.

I also see your eliminated the ret var and wrapped in `()` instead, I’m assuming this returns just the first value of the pair returned from `gsub` (returning the full pair can be an issue if using the function directly inside a `string.format`).

phanen:
Didn't benchmark it in detail, just saw the code and thought it may be better to don't use callback. 

And use `()` will return only the first args, I believed it should be documented in lua. (the code is just copy-paste from vim.pesc https://github.com/neovim/neovim/blob/052e048db676ef3e68efc497c02902e3d43e6255/runtime/lua/vim/shared.lua#L753-L753)

ibhagwan:
> And use () will return only the first args, I believed it should be documented in lua manual.

The syntax does make sense.

> `vim.pesc` is added in neovim 0.5.0.

I only found out about it recently

PR #1452, opened 2024-09-21
feat(keymaps): add the `show_details` option

loyd:
Hi, thanks for this package!

I'm migrating from the telescope (because of slowness on large codebases), and I noticed that the keymaps picker is too restrictive, which is not very useful for those who provide detailed descriptions for all bindings:

This PR introduces the `show_details` option, which (if `false`) disables the `detail` column completely and shows this info only if `description` is empty.

By default, `show_details = true`, thus it doesn't affect existing users.

Result:

### DIFF
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -320,7 +320,6 @@ M.keymaps = function(opts)
   opts = config.normalize_opts(opts, "keymaps")
   if not opts then return end
 
-  local formatter = "%s │ %-14s │ %-33s │ %s"
   local key_modes = opts.modes or { "n", "i", "c", "v", "t" }
   local modes = {
     n = "blue",
@@ -330,9 +329,25 @@ M.keymaps = function(opts)
     t = "green"
   }
   local keymaps = {}
+  local format = nil
 
+  if opts.show_details then
+    local formatter = "%s │ %-14s │ %-33s │ %s"
 
-  local add_keymap = function(keymap)
+    format = function(mode, lhs, desc, rhs)
+      -- we don't trim `lhs` here because it's better to violate the structure
+      -- than to cut off the most useful information
+      return string.format(formatter, mode, lhs, string.sub(desc or "", 1, 33), rhs)
+    end
+  else
+    local formatter = "%s │ %-14s │ %s"
+
+    format = function(mode, lhs, desc, rhs)
+      return string.format(formatter, mode, lhs, desc or rhs)
+    end
+  end
+
+  local function add_keymap(keymap)
     -- ignore dummy mappings
     if type(keymap.rhs) == "string" and #keymap.rhs == 0 then
       return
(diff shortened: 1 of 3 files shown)


lua/fzf-lua/providers/nvim.lua
```diff
@@ -330,9 +329,25 @@ M.keymaps = function(opts)
...
+  local format = nil
 
+  if opts.show_details then
+    local formatter = "%s │ %-14s │ %-33s │ %s"
 
-  local add_keymap = function(keymap)
+    format = function(mode, lhs, desc, rhs)
+      -- we don't trim `lhs` here because it's better to violate the structure
+      -- than to cut off the most useful information
+      return string.format(formatter, mode, lhs, string.sub(desc or "", 1, 33), rhs)
```
loyd, line 340: It's `1, 30` currently, so we're losing several chars here.

Is it intentional or simply some legacy?

ibhagwan: > Is it intentional or simply some legacy?

If I recall correctly this was part of a PR submitted (#876), don’t think it’s intentional.

doc/fzf-lua.txt
```diff
@@ -1277,6 +1277,9 @@ open an issue and I'll be more than happy to help.**
         -- by default, we ignore <Plug> and <SNR> mappings
         -- set `ignore_patterns = false` to disable filtering
         ignore_patterns   = { "^<SNR>", "^<Plug>" },
+        -- by default, both description and details are shown
+        -- `false` shows details only if the description is missing
+        show_details      = true,
```
ibhagwan: You don’t need to make any changes to `doc/fzf-lua.txt`, there’s a CI step that auto generates that directly from the README.

ibhagwan:
Ty for the PR @loyd, I answered your Q in the review and a minor comment.

Did you check the preview of the lua function callback works without show_detsil? I’m thinking perhaps we can have a more flexible approach to the issue reordering the entries with `--with-nth`?

We can set `--delimiter=|`, place the description last (with no substring limitation) and then use `--with-nth=1,2,4,3` for a “show_detail” entry and `--with-nth=1,2,4` without (the 3rd field being the details).

doc/fzf-lua.txt
loyd: Nicely)

loyd: Removed

loyd:
> Ty for the PR @loyd, I answered your Q in the review and a minor comment.

Thanks for the fast review. Fixed.

> Did you check the preview of the lua function callback works without show_detsil?

Yep, of course, it looks as previously, but without an extra three spaces in the `description` column.

> We can set --delimiter=|, place the description last (with no substring limitation) and then use --with-nth=1,2,4,3 for a “show_detail” entry and --with-nth=1,2,4 without (the 3rd field being the details).

That would be nice. However, it means columns cannot include `|` (unlikely, but possible).

ibhagwan:
> That would be nice. However, it means columns cannot include | (unlikely, but possible).

you’re right, perhaps we can add `utils.nbsp .. |` as separator and use utils.nbsp as delimiter?

loyd:
> use --with-nth=1,2,4,3 for a “show_detail”
> and --with-nth=1,2,4 without

The column shouldn't be limited, but I doubt limiting `detail` is a good idea for users without good descriptions.

Thus, details should be limited (and with `%-33s`) if `show_details=true` and shouldn't be otherwise. That means it's not only about ordering, and `--with-nth` is not enough. So, I'm not sure it's significant improvements in code (I've tried it).

Maybe calculate the real width of columns?

ibhagwan:
> > use --with-nth=1,2,4,3 for a “show_detail”
> > and --with-nth=1,2,4 without
> 
> The column shouldn't be limited, but I doubt limiting `detail` is a good idea for users without good descriptions.
> 
> Thus, details should be limited (and with `%-33s`) if `show_details=true` and shouldn't be otherwise. That means it's not only about ordering, and `--with-nth` is not enough. So, I'm not sure it's significant improvements in code (I've tried it).
> 
> Maybe calculate the real width of columns?

Maybe I’m overthinking this, `show_detail` option the way you did it should be perfect :)

ibhagwan:
And ty so much for the contribution @loyd :)

PR #1457, opened 2024-09-28
feat(filetype): devicons support

phanen:


lua/fzf-lua/providers/nvim.lua
```diff
@@ -423,6 +423,22 @@ M.filetypes = function(opts)
   local entries = vim.fn.getcompletion("", "filetype")
   if utils.tbl_isempty(entries) then return end
 
+  local get_icon
+  if pcall(require, "mini.icons") then
```
ibhagwan, line 427: Most people have `mini.nvim` installed but not necessarily use `mini.icons` and this call will lazy load the module, the proper way to check for mini.icons (as per the docs) is to check for the existence of `_G.MiniIcons`.

lua/fzf-lua/providers/nvim.lua
```diff
@@ -423,6 +423,22 @@ M.filetypes = function(opts)
...
+    get_icon = function(ft) return require("mini.icons").get("filetype", ft) end
```
ibhagwan, line 428: This seems to prioritize mini over devicons which is different to any other instance where icons are used in the code.

I think the best way to add this is to add another function to the DevIcons object (icon_by_ ft) in `devicons.lua` and  call the function directly from `require(“fzf-lua.devicons”).icon_by_ ft`, this way we keep the same auto loading logic and don’t need any lazy loading/pcall’s here, the module should take of it instead.

ibhagwan (commented):
This is the module: https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/devicons.lua

This way we can also control this by the same already known setting `file_icons`, keeping the same consistent behavior:
- false: no icons
- true: auto load prioritizing devicons
- “mini”: use mini.icons

Try it with `:FzfLua files file_icons=mini`

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/b87e37dd472ed09b017e1654e8c59752d27c964c

Sorry @phanen for robbing you of the PR, I had a few minutes and decided to implement it the way I described above.

Since devicons `get_icon_by_filetype` is very lacking (only 200 out of almost 800) I set the default to false.

>`:FzfLua filetypes file_icons=true`
image

>`:FzfLua filetypes file_icons=mini`
image

phanen:
Never mind and glad to know that. it seems that mini.icons has the first class support for this usecase.

PR #1464, opened 2024-09-30
feat: adopt upstream treesitter breaking change

RayGuo-ergou:
Due to a recent treesitter breaking change on neovim.
`vim.treesitter.language.add` returns `true` or `nil, error` now. thus can use it directly without `pcall`.

see https://github.com/neovim/neovim/pull/30379

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -755,7 +755,7 @@ end
 -- Attach ts highlighter, neovim >= v0.9
 local ts_attach = function(bufnr, ft)
   local lang = vim.treesitter.language.get_lang(ft)
-  local loaded = pcall(vim.treesitter.language.add, lang)
+  local loaded = utils.has_ts_parser(lang)
   if lang and loaded then
     local ok, err = pcall(vim.treesitter.start, bufnr, lang)
     if not ok then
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @RayGuo-ergou, but what are we really gaining from this? Seems like the code is simpler and supports older versions and the nightly with just one pcall which doesn’t have much downside.

RayGuo-ergou:
Hi @ibhagwan the motivation for me to do this change is because I noticed when preview files that do not have treesitter install, a warning will emit. e.g.
>[Fzf-lua] unable to attach treesitter highlighter for filetype 'ps1': /usr/local/share/nvim/runtime/lua/vim/treesitter.lua:426: Parser could not be created for buffer 25 and language "ps1"

I believe this due to in this upstream PR, `error` call was removed from `vim.treesitter.language.add` thus the first return value of `pcall` will always be true.

e.g.
https://github.com/neovim/neovim/pull/30379/files#diff-2a708a89ff80aafd5d7c5ab74bde2e6e6a5ac501ecbb11893f98634fce62a5a8L101
https://github.com/neovim/neovim/pull/30379/files#diff-2a708a89ff80aafd5d7c5ab74bde2e6e6a5ac501ecbb11893f98634fce62a5a8L107

ibhagwan:
> I believe this due to in this upstream PR, error call was removed from vim.treesitter.language.add thus the first return value of pcall will always be true.

I see.

PR #1467, opened 2024-10-02
Jump to single result in lsp for telescope profile

lala74:
Telescope jump to the result if number of result is 1. Add option for lsp for telescope profile to match this behavior


ibhagwan:
Ty @lala74! merged.

PR #1492, opened 2024-10-13
Fix duplicate `grep_quickfix` in README

stsewd:
The command is mentioned twice, but the second mention should be `grep_loclist`


ibhagwan:
Ty @stsewd!

Issue #781, opened 2023-06-11
Question about live_grep with rg's --fixed-strings option


deponian:
Hi again!

I've read documentation and wiki after my yesterday's [PR](https://github.com/ibhagwan/fzf-lua/pull/777) but I still have a question.

I want to perform all my searches using `live_grep` with no regex processing, only literal strings. So I use `--fixed-strings` option for `rg` and disable escaping with `no_esc = true`. But I still can't search option-like strings (`-info`, `--help`, etc.). This fails because whole query is passed as argument for `nvim --headless ...` command for `mt` version of `live_grep`.

This can be fixed by using `--` (shell's options/arguments separator) here: https://github.com/ibhagwan/fzf-lua/blob/267098387dd7a3c53f8a4cfb17285ba05d841353/lua/fzf-lua/providers/grep.lua#L301-L302

Like this:
```lua
command = command:gsub(core.fzf_query_placeholder, "{argvz}")
        .. " -- " .. core.fzf_query_placeholder
```

Am I overcomplicating things again and there is some existing option for my kind of case?

ibhagwan:
> Am I overcomplicating things again and there is some existing option for my kind of case?

You're not overcomplicating **your use case** but it does overcomplicate everything else :-)

The purposed solution solves your use case alone and only for `live_grep` **with glob support enabled**, this doesn't support using `--fixed-strings` with `grep`, for example try running `:FzfLua grep debug=true rg_glob=false no_esc=true search=--no`, this will result in:
![image](https://github.com/ibhagwan/fzf-lua/assets/59988195/aa90920f-3966-49ba-928c-b09c71eec769)

A more generic solution would be to add the `--` separator to the grep command builder at:
https://github.com/ibhagwan/fzf-lua/blob/267098387dd7a3c53f8a4cfb17285ba05d841353/lua/fzf-lua/providers/grep.lua#L103

By changing the code to:
```lua
command = ("%s -- %s %s"):format(command, search_query, search_path) 
```

Which will no result in:
![image](https://github.com/ibhagwan/fzf-lua/assets/59988195/36632b2f-5db0-48d4-bc5f-948b40108091)

[...]

PR #1502, opened 2024-10-25
Feature: Add action toggle_fixed_strings

hernancerm:
### Problem

- I could not find a simple way to toggle between regex and a literal search when using grep.

### Solution

- Implement the action `toggle_fixed_strings` to toggle the flag `--fixed-strings`. This flag is supported by [grep](https://www.gnu.org/software/grep/manual/grep.html#grep-Programs-1), [ripgrep](https://manpages.debian.org/testing/ripgrep/rg.1.en.html#SEARCH_OPTIONS) and [ag](https://man.archlinux.org/man/extra/the_silver_searcher/ag.1.en#OPTIONS). I coded this action following the implementation of `toggle_ignore`; so it was mostly an imitation game. Now toggling between regex and literal search is as simple as adding one line in the actions config of the provider:

```
··
  grep = {
    actions = {
      ["ctrl-g"] = { actions.grep_lgrep },
      ["ctrl-x"] = { actions.toggle_fixed_strings } -- Toggle between regex/literal search with <C-x>.
    },
  }
··
```

---

### Other solutions I considered

- There has already been a discussion about how to use the `--fixed-strings` flag in the grep provider. For the simple use case of just toggling between regex and literal search, I found the [code](https://github.com/ibhagwan/fzf-lua/issues/781#issuecomment-1595831081) provided by [deponian](https://github.com/deponian) to be too complicated. If possible, I would appreciate if fzf-lua provides built-in support for this flag so the user-side configuration is easy.
  - #781  

Thanks.

### DIFF
--- a/doc/fzf-lua.txt
+++ b/doc/fzf-lua.txt
@@ -1128,6 +1128,8 @@ open an issue and I'll be more than happy to help.**
           ["ctrl-g"]      = { actions.grep_lgrep }
           -- uncomment to enable '.gitignore' toggle for grep
           -- ["ctrl-r"]   = { actions.toggle_ignore }
+          -- uncomment to enable '--fixed-strings' (exact match) toggle for grep
+          -- ["ctrl-x"]   = { actions.toggle_fixed_strings }
         },
         no_header             = false,    -- hide grep|cwd header?
         no_header_i           = false,    -- hide interactive header?
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -850,6 +850,11 @@ M.toggle_ignore = function(_, opts)
   M.toggle_flag(_, vim.tbl_extend("force", opts, { toggle_flag = flag }))
 end
 
+M.toggle_fixed_strings = function(_, opts)
+  local flag = opts.toggle_ignore_flag or "--fixed-strings"
+  M.toggle_flag(_, vim.tbl_extend("force", opts, { toggle_flag = flag }))
+end
+
 M.toggle_hidden = function(_, opts)
   local flag = opts.toggle_hidden_flag or "--hidden"
   M.toggle_flag(_, vim.tbl_extend("force", opts, { toggle_flag = flag }))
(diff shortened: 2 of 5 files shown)


ibhagwan:
Ty @hernancerm but I don’t wish to add actions for every single flag possible, you can achieve this customization easily for any flag as below:
```lua

require("fzf-lua").setup({
  grep = {
    actions = {
      ["alt-g"] = {
        fn = function(_, opts)
          require("fzf-lua").actions.toggle_flag(_, vim.tbl_extend("force", opts, {
            toggle_flag = "--ignore-case"
          }))
        end,
        desc = "toggle-ignore-case",
        header = "Toggle Ignore Case",
      },
    },
  },
})
```

hernancerm:
I figured that might be a concern, I understand. Thank you for the snippet! The only thing it's missing is alternating the header between "Respect special regex chars" and "Disable special regex chars" depending on the presence of the flag. I'll try to tweak what you provided to get this behavior.

ibhagwan:
> I figured that might be a concern, I understand. Thank you for the snippet! The only thing it's missing is alternating the header between "Respect special regex chars" and "Disable special regex chars" depending on the presence of the flag. I'll try to tweak what you provided to get this behavior.

Change the header to a function and you can have that too:
```lua
    header = function(o)
      local flag = o.toggle_hidden_flag or "--hidden"
      if o.cmd and o.cmd:match(utils.lua_regex_escape(flag)) then
        return "Exclude hidden files"
      else
        return "Include hidden files"
      end
    end,
```

hernancerm:
That works like a charm. Thanks for the follow up @ibhagwan!

For future reference, here is the full action mapping I'm using to toggle `--fixed-strings` on `grep`:

```lua
["ctrl-x"] = {
  fn = function(_, opts)
    actions.toggle_flag(_, vim.tbl_extend("force", opts, {
      toggle_flag = "--fixed-strings"
    }))
  end,
  desc = "toggle-fixed-strings",
  header = function(o)
    local flag = "--fixed-strings"
    if o.cmd and o.cmd:match(utils.lua_regex_escape(flag)) then
      return "Respect regex chars"
    else
      return "Disable regex chars"
    end
  end,
}
```

ibhagwan:
Ty for the update and snippet @hernancerm!

PR #1503, opened 2024-10-26
fix(lsp): vim.lsp.util.jump_to_location is now deprecated

MariaSolOs:
`vim.lsp.util.jump_to_location` was deprecated in https://github.com/neovim/neovim/pull/30877.

This PR replaces its uses by `vim.lsp.show_document` (which `jump_to_location` uses in its implementation).

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -129,7 +129,7 @@ M.vimcmd_entry = function(_vimcmd, selected, opts, pcall_vimcmd)
       -- Adjust "<auto>" edits based on entry being buffer or filename
       local vimcmd = _vimcmd:gsub("<auto>", entry.bufnr and entry.bufname and "b" or "e")
       -- Do not execute "edit" commands if we already have the same buffer/file open
-      -- or if we are dealing with a URI as it's open with `vim.lsp.util.jump_to_location`
+      -- or if we are dealing with a URI as it's open with `vim.lsp.util.show_document`
       -- opts.__CTX isn't guaranteed by API users (#1414)
       local CTX = opts.__CTX or utils.CTX()
       if vimcmd == "e" and (entry.uri or path.equals(fullpath, CTX.bname))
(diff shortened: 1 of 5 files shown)


ibhagwan:
> `vim.lsp.util.jump_to_location` was deprecated in [neovim/neovim#30877](https://github.com/neovim/neovim/pull/30877).
> 
> This PR replaces its uses by `vim.lsp.show_document` (which `jump_to_location` uses in its implementation).

Ty @MariaSolOs! Looking at your code it seems that replacing the function name is enough, is the implementation exactly the same? Will this work similarly with Java jar files prefixed with `jdtls://`?

Do we need to maintain backward compatibility with 0.8/0.9/etc? While I don’t think we need to maintain backward compatibility forever I’d like to at least make sure this change works with the earliest supported version by lspconfig (currently at 0.8 if I’m not mistaken).

MariaSolOs:
Given that `vim.lsp.util.jump_to_location` [simply calls `show_document`](https://github.com/neovim/neovim/blob/25b53b593ef6f229fbec5b3dc205a7539579d13a/runtime/lua/vim/lsp/util.lua#L965) (even before the deprecation), I believe that this replacement should be safe in all versions that this plugin supports.

ibhagwan:
> Given that `vim.lsp.util.jump_to_location` [simply calls `show_document`](https://github.com/neovim/neovim/blob/25b53b593ef6f229fbec5b3dc205a7539579d13a/runtime/lua/vim/lsp/util.lua#L965) (even before the deprecation), I believe that this replacement should be safe in all versions that this plugin supports since the implementation behind the scenes remains identical :)

TIL :)

Ty @MariaSolOs!

phanen:
`show_document` is added after neovim 0.9... maybe compat is still needed (but do we really need maintain such old version, even debian stable have 0.7+ now... https://repology.org/project/neovim/versions)

MariaSolOs:
@phanen You're right, my bad. I'll make this backwards compatible ASAP.

I'll let @ibhagwan decide on what the minimum version support should be :)

ibhagwan:
> @phanen You're right, my bad. I'll make this backwards compatible ASAP.
> 
> I'll let @ibhagwan decide on what the minimum version support should be :)

Ty @phanen for pointing this out, I agree we don’t need to maintain backward compatibility forever, for LSP matching the minimum version requirement with nvim-lspconfig is reasonable, it’s currently at 0.8.

PR #1504, opened 2024-10-27
fix(lsp): preserve backwards compatibility with jump_to_location deprecation

MariaSolOs:
Follow up of #1503 to preserve backwards compatibility with versions 5-9.


MariaSolOs:
Thanks @phanen for catching this!

ibhagwan:
Ty @phanen & @MariaSolOs!

Issue #755, opened 2023-05-12
[Feature] Allow using a function to `register_ui_select`


akinsho:
### Description 
This relates somewhat to https://github.com/ibhagwan/fzf-lua/issues/717 where in that issue the OP was trying to get access to the options passed to the `ui.select` function so that they could change the behaviour of the previewer. In my case I'd like to do something similar but rather than change a previewer, I'd like instead to use the prompt that was passed in as the title of the window.

Rather than just solve that specific use case though I thought it would be quite helpful to allow a more flexible pattern for future variations of this request could also be solved e.g.

```lua
require('fzf-lua').register_ui_select(function(the_opts)
	local width = the_opts.kind == "codeaction" and small_width or medium_width
	local position = the_opts.kind == "codeaction" and "cursor" or "editor"
	return {
		previewer = CustomPreviewer(the_opts),
		winopts = {
 			title = the_opts.prompt,
			width = width height = 0.33,
			row = 0.5,
			relative = positions
		 },
	}
end)
```

I think this pattern would allow completing customising how the `ui_select` looks based on the current case.

ibhagwan:
Hi @akinsho, this makes sense, I’m currently traveling so might take me a couple of days to get to it though.

akinsho:
Awesome thanks 🙌🏿

PR #1512, opened 2024-11-04
Feature: Add an empty space at the end of the ui_select prompt to a better formatting when writing the input

estacioneto:
It's a small feature I think would be nice.

When using [kulala.nvim](https://github.com/mistweaverco/kulala.nvim), I've noticed the search prompt was not ideal since it had no space between the text and the input. For the default prompt, though, I've changed just a bit the result from `"Select one of> "` to `"Select one of > "`.

Screenshot 2024-11-04 at 11 38 56
Screenshot 2024-11-04 at 11 39 12

Please let me know if this is not something you'd want the plugin to deal with or if I need to make any changes (but feel free to do them as well) :)

### DIFF
--- a/lua/fzf-lua/providers/ui_select.lua
+++ b/lua/fzf-lua/providers/ui_select.lua
@@ -95,7 +95,14 @@ M.ui_select = function(items, ui_opts, on_choice)
   })
 
   -- Force override prompt or it stays cached (#786)
-  local prompt = ui_opts.prompt or "Select one of:"
+  local prompt = ui_opts.prompt
+  if prompt then
+    -- Add a space at the end of the prompt (if not already present) since the prompt and the input are inline
+    prompt = prompt:gsub("%s*$", " ", 1)
+  else
+    prompt = "Select one of :"
+  end
+
   opts.fzf_opts["--prompt"] = prompt:gsub(":%s?$", "> ")
 
   -- save items so we can access them from the action


ibhagwan:
Ty @estacioneto.

The reason the prompt doesn’t have space is because it’s controlled by the caller, the extra space should be set by kuala.

As for the default prompt it should remain matching to neovim’s default:
IMG_2639

Finally, this change doesn’t need a PR, use `register_ui_select` with a function fix the prompt and return a spaced out one if you wish, see https://github.com/ibhagwan/fzf-lua/issues/755 or  https://github.com/ibhagwan/fzf-lua/wiki#automatic-sizing-of-heightwidth-of-vimuiselect.

estacioneto:
Hi, @ibhagwan!

Thanks for the response! I've thought you'd like it to be solved here since this is the actual plugin that puts the prompt and the input on the same line.

I could achieve it with this:

```lua
  fzf.register_ui_select(function(opts)
    return { prompt = opts.prompt:gsub('%s*$', ' ', 1) }
  end)
```

Thanks again :)

ibhagwan:
> Thanks for the response! I've thought you'd like it to be solved here since this is the actual plugin that puts the prompt and the input on the same line.

I’m still conflicted whether this is a problem fzf-lua should solve by code, the solution is quite simple anyways.

PR #1523, opened 2024-11-13
New "FzfLua tmux_files" command

cgimenes:
The idea is to move this plugin to fzf-lua: https://github.com/cgimenes/tmux-files.nvim

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -1135,6 +1135,10 @@ DAP active session jump to frame
 
 Tmux paste buffers
 
+#### tmux_files
+
+Filenames (and line location) found on other Tmux panes
+
 ---
 
 ### Completion Functions
--- a/README.md
+++ b/README.md
@@ -385,6 +385,7 @@ to see detailed usage notes and a comprehensive list of all available options.**
 | Command              | List                                       |
 | -------------------- | ------------------------------------------ |
 | `tmux_buffers`         | list tmux paste buffers                    |
+| `tmux_files`           | list files that appear on other tmux panes |
 
 ### Completion Functions
 | Command              | List                                       |
--- a/doc/fzf-lua-opts.txt
+++ b/doc/fzf-lua-opts.txt
@@ -1574,6 +1574,11 @@ tmux_buffers                                       *fzf-lua-opts-tmux_buffers*
 
 Tmux paste buffers
 
+                                                                              
+tmux_files                                       *fzf-lua-opts-tmux_files*
+
+Filenames (and line location) found on other Tmux panes
+
 
                                                                               
 COMPLETION FUNCTIONS                       *fzf-lua-opts-completion-functions*
--- a/doc/fzf-lua.txt
+++ b/doc/fzf-lua.txt
@@ -462,6 +462,7 @@ TMUX                                                            *fzf-lua-tmux*
 | Command              | List                                       |
 | -------------------- | ------------------------------------------ |
 | `tmux_buffers`         | list tmux paste buffers                    |
+| `tmux_files`           | list files that appear on other tmux panes |
 
 
                                                                               
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -1092,13 +1092,14 @@ M.defaults.tmux                 = {
     fzf_opts = { ["--no-multi"] = true, ["--delimiter"] = "[:]" }
   },
   files = {
-    prompt      = "Tmux Files> ",
-    file_icons  = true and M._has_devicons,
-    color_icons = true,
-    git_icons   = false,
-    previewer   = M._default_previewer_fn,
-    _actions    = function() return M.globals.actions.files end,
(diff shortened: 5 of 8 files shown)


ibhagwan:
@cgimenes, I’m still not sure I understand the purpose of this, is your final plan close to https://github.com/laktak/extrakto?

If you are planning on extracting file names from pane history why the need for project files? Isn’t this a duplication of `:FzfLua files`?

cgimenes:
Hi @ibhagwan!
Actually, no. BTW I'm having second thoughts if adding this feature to fzf-lua makes sense, but we can discuss this after I explain my idea below.
I use vim-test together with `preservim/vimux` to execute unit tests while I'm developing; vim-test uses vimux to open a small tmux pane just below neovim with the test results.
When a test fails, I need to use something like extrakto (or tmux-copycat in my case) to yank the filename of the test that failed, or even another file in the stacktrace for exceptions, and use `FzfLua files` to open it, then go to the line that I want to change.
This process is a little cumbersome and I didn't want to run vim-test inside neovim, because I lose the ability to add breakpoints in my code that launch some kind of REPL. So, I decided to create `tmux-files.nvim`: a plugin that grabs texts from every pane open in the current tmux window, excluding the current one that should be neovim, and list every filename it finds using fzf-lua.
Now, for the `project_only` option, I don't really know if it's necessary. The idea is to filter files that aren't inside the current directory, because sometimes the stacktrace shows files from libraries outside the project. Maybe, a better idea would be to create a keymap to toggle this feature inside fzf-lua, something like <ctrl-g> for `FzfLua files`.

ibhagwan:
> BTW I'm having second thoughts if adding this feature to fzf-lua makes sense, because maybe this is just too specific for this kind of workflow, but we can discuss this further.

It is very specific unless you’re intending to go extrakto-style and extract all file names history included.

> Now, for the project_only option, I don't really know if it's necessary. The idea is to filter files that aren't inside the current directory, because sometimes the stacktrace shows files from libraries outside the project. Maybe, a better idea would be to create a keymap to toggle this feature inside fzf-lua, something like <ctrl-g> for FzfLua files.

I still don’t understand why this requires running fd/find/dir/etc, if all you wish is to exclude files outside cwd why not test the path against cwd or git root? You can use `path.is_relative_to` for that.

cgimenes:
> It is very specific unless you’re intending to go extrakto-style and extract all file names history included.

Exactly, I want history included.

> I still don’t understand why this requires running fd/find/dir/etc, if all you wish is to exclude files outside cwd why not test the path against cwd or git root? You can use `path.is_relative_to` for that.

Probably using fd/find/etc isn't the best idea. It's almost an O(n squared), because for each line found by the cmd, we run a loop to check if it exists in the fd results. But, I don't know another way of doing it. I tried using `path.is_relative_to` but it's expecting an absolute path, and most part of the cmd results are relative paths.

My next attempt will be something like a combination of `path.is_absolute`, `path.is_relative_to` and cwd concatenation.

cgimenes:
I think I've found a better solution. Could you take a look?

cgimenes:
I've found a better solution: rely on `opts.cwd_only` for absolute paths and on `utils.file_is_readable` for relative ones.

ibhagwan:
> I've found a better solution: rely on `opts.cwd_only` for absolute paths and on `utils.file_is_readable` for relative ones.

Good option if you’re using make_entry.file already, that way you also get the benefit of file_ignore_patterns, etc.

Be careful of is readable as it’s not very performant, there is a logic in old files with the `stat_file` option to skip the test, might be smart to do the same here.

cgimenes:
Done

ibhagwan:
> The idea is to move this plugin to fzf-lua: https://github.com/cgimenes/tmux-files.nvim

I like this idea, you can optimize it indefinitely without the need for PRs, and I don’t have to maintain it :)

> Done

Given the tmux command can take a long time to parse with huge history and also the fact you’re testing files against the file system I would attempt to make this “multiprocess”, that would require ditching the `fn_transform` and using imported lua code for parsing the output in a headless neovim instance.

To do that you’d need to explore the code a bit more and understand how `libuv.wrap_spawn_stdio` works and how to execute your code in the headless process (and what you can and cannot do there).

This is the “secret” of fzf-lua’s responsiveness, without this, you’re most likely to encounter a laggy experience at some point.

cgimenes:
> Given the tmux command can take a long time to parse with huge history and also the fact you’re testing files against the file system I would attempt to make this “multiprocess”, that would require ditching the `fn_transform` and using imported lua code for parsing the output in a headless neovim instance.
> 
> To do that you’d need to explore the code a bit more and understand how `libuv.wrap_spawn_stdio` works and how to execute your code in the headless process (and what you can and cannot do there).
> 
> This is the “secret” of fzf-lua’s responsiveness, without this, you’re most likely to encounter a laggy experience at some point.

I'll take a look at this later as it's very responsive at the moment. The changes I've made while working on this PR made it much more responsive than what I had in my own repo.

> I like this idea, you can optimize it indefinitely without the need for PRs, and I don’t have to maintain it :)

You're totally right. I'll close this PR now and work on my repo. Thank you so much for all your comments!

PR #1541, opened 2024-11-27
Display Selected Highlight Group In Colorscheme Provider

barrett-ruth:
Currently, running `:FzfLua highlights` doesn't actually do *anything* when an entry is selected.

I think it is logical to display the highlighted group in the command line as if `:hi <group>` was run.

### DIFF
--- a/lua/fzf-lua/providers/colorschemes.lua
+++ b/lua/fzf-lua/providers/colorschemes.lua
@@ -144,6 +144,11 @@ M.highlights = function(opts)
     end
   end
 
+  opts.fn_selected = function(selected)
+    vim.cmd('hi ' .. selected[2])
+    vim.api.nvim_exec2('hi ' .. selected[2], {})
+  end
+
   core.fzf_exec(contents, opts)
 end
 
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @barrett-ruth!

barrett-ruth:
@ibhagwan can you uhhh... look at my review please! I should've marked the PR as draft - there's *definitely* two duplicate lines of code when it comes to displaying the hl group.

ibhagwan:
> @ibhagwan can you uhhh... look at my review please! I should've marked the PR as draft - there's _definitely_ two duplicate lines of code when it comes to displaying the hl group.

Whoops, I looked at the code figured what it was supposed to do and tested it quickly and saw it working

**And** missed the duplicate lines lol

Let me know if you’d like to fix it with another PR or I ca remove one of these.

ibhagwan:
I’m confused, removing any line prevents the hl from being displayed in `:messages`… this only works when both lines exist, I must leave soon so I’ll have to take care of this later, if you have any insight feel free to submit a fix PR.

barrett-ruth:
OK. I found that was the case too - not sure why lol. I was also just wondering if there was any more elegant way to write the code, but I found it all to be functional myself. Thanks for the quick review.

PR #1545, opened 2024-12-04
fix: `attempt to index field 'render_markdown' (a nil value)`

idanarye:


ibhagwan:
Ty @idanarye, I just fixed this but reverted so you can have this merged :)

Issue #1547, opened 2024-12-04
Feature: Line numbers in tags file (ctags)


dvquentincarrel:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

Hello, I believe the "tags" subcommand can better exploit the content of a tag file to provide meaningful results to the user.  
This is the first issue I open on an open-source project, so please do not hesitate to call out any faux-pas I may have made.

### Context
By default, Universal ctags (and I believe, other ctags variants too) use a regular expression to location a symbol in a code file.   
This can lead to issues if two symbols have the same signature in a single file, where the location of each of the two symbol will be defined by the exact same regular expression, thus preventing from distinguishing them.  
I work in a codebase where this is often the case.  
Ctags allow the usage of a line number instead of a regular expression (universal ctags allows both at once), to indicate the location of a symbol in the file. 

### Issue
I have noticed the handling of line numbers as what is called the `{tagaddress}` in `:h tagsrch.txt` is a bit peculiar in fzf-lua.  
When only the line number are used (`ctags --excmd=number`), fzf-lua always shows the start of the file in the preview, and redirects the user to the first line of the file upon the selection of the tag.  
When both the regex and the line number are included in the tags file (`ctags --excmd=combined`), fzf-lua still seem to ignore the line number.

### Request
I believe fzf-lua would benefit from being able to properly use tags file which use line numbers instead of a regex.  
I have looked and tinkered around, and it seems the implementation of the feature has already been attempted.

For instance, when using both line numbers and regexes in the tags file, replacing the code in the function M.vimcmd_entry_function :
https://github.com/ibhagwan/fzf-lua/blob/d9db3b684384bcfd1393914b76f89f0205f183a3/lua/fzf-lua/actions.lua#L212-L216
by something like:
```lua
elseif entry.ctag then
  if entry.line then
      vim.api.nvim_win_set_cursor(0, { entry.line, 0 })
  else
      vim.api.nvim_win_set_cursor(0, { 1, 0 })
      vim.fn.search(entry.ctag, "W")
  end
```
correctly uses the line number to redirect the user in the file.

However, this does not work if only line numbers are used in the tags file, as the `entry` table would not contain a `ctag` element.  
The pattern extracting the line of a tag (`local line, tag = text:match("(%d-);?(/.*/)")` in make_entry.lua, function `M.tag`) expects a line number to always be followed by a regex. Maybe falling back to attempting to use a line number would be good ?

[...]

Issue #355, opened 2022-03-03
FzfLua Tags very slow with bat/cat previewer


mikesart:
### Info

- Operating System: Debian Testing
- Shell: Bash 5.1.16
- Terminal: xfce4-terminal 0.8.10 (Xfce 4.16)
- `nvim --version`: NVIM v0.6.1
- `fzf --version`: 0.29.0 (dc975e8)

When I do this:

```
tags = {
    previewer         = "builtin",
    prompt            = 'Tags❯ ',
    ctags_file        = "tags",
    git_icons         = true,       -- show git icons?
    file_icons        = true,       -- show file icons?
    color_icons       = true,       -- colorize file|git icons
  },
```

Opening up the tags window is very fast. When I switch previewer to "bat", it's incredibly slow as if it's running bat on every file as it loads the tags file?

And actually, on a project like UnrealEditor with a *lot* of tags (13,175,744) even the builtin previewer is unusable. The fzf ":Tags" command takes a few seconds, but ":FzfLua tags" takes forever and then shows nothing. This repros if I comment out my tags section above as well.

Below is my previewers section. Is this by design or am I doing something wrong here? Thank you!

```
  previewers = {
    bat = {
      cmd             = "batcat",
      args            = "--style=numbers,changes,header --color always",
      theme           = 'Dracula',  -- bat preview theme (bat --list-themes)
      config          = nil,        -- nil uses $BAT_CONFIG_PATH
    },
    builtin = {
      syntax          = true,       -- preview syntax highlight?
      syntax_limit_l  = 50,         -- syntax limit (lines), 0=nolimit
      syntax_limit_b  = 1024*1024,  -- syntax limit (bytes), 0=nolimit
    },
  },
```

ibhagwan:
>Opening up the tags window is very fast. When I switch previewer to "bat", it's incredibly slow as if it's running bat on every file as it loads the tags file?

Unfortunately that's "by design", when using a previewer that isn't `builtin` fzf-lua needs to supply line number of the tag to the fzf entry (line), without it `fzf` won't know what line offset to display in the preview window, to do that it has to `rg` every single tag entry to find the right line and that's why it's so slow.

Althought I can find the line number and supply it to `bat` "on demand" I couldn't find a way provide the preview offset (i.e. "we loaded a 300 line preview and the match is at line 250") on demand, it needs to exist in the entry and provided using the fzf `FIELD INDEX EXPRESSION`.

[...]

PR #1550, opened 2024-12-05
fix(ctag): better handle line number for tag entries

dvquentincarrel:
Attempts to fix the problems highlighted in [this issue](https://github.com/ibhagwan/fzf-lua/issues/1547), notably:
- the line number of tag entries with no regex was not properly extracted, and thus were unusable
- tag entries with both a regex and a line number did not take advantage of the line number, thus were susceptible to collisions on the regex
-   the default previewer for tags could not make use of the line number of an entry

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -210,7 +210,7 @@ M.vimcmd_entry = function(_vimcmd, selected, opts, pcall_vimcmd)
         else
           utils.jump_to_location(entry, "utf-16")
         end
-      elseif entry.ctag then
+      elseif entry.ctag and not entry.line then
         vim.api.nvim_win_set_cursor(0, { 1, 0 })
         vim.fn.search(entry.ctag, "W")
       elseif not opts.no_action_set_cursor and entry.line > 0 or entry.col > 0 then
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -439,6 +439,11 @@ M.tag = function(x, opts)
   -- from rg/grep output when using `tags_grep_xxx`
   local align = utils.has_ansi_coloring(name) and 47 or 30
   local line, tag = text:match("(%d-);?(/.*/)")
+  if not tag then
+    -- lines with a tag located solely by line number contain nothing but the
+    -- number at this point
+    line = text:match("%d+")
+  end
   line = line and #line > 0 and tonumber(line)
   return string.format("%-" .. tostring(align) .. "s%s%s%s: %s",
     name,
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -441,7 +441,7 @@ M.tag = function(x, opts)
   local line, tag = text:match("(%d-);?(/.*/)")
   if not tag then
     -- lines with a tag located solely by line number contain nothing but the
-    -- number at this point
+    -- number at this point (e.g. using "ctags -R --excmd=number")
     line = text:match("%d+")
   end
   line = line and #line > 0 and tonumber(line)
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1131,11 +1131,17 @@ function Previewer.tags:set_cursor_hl(entry)
   pcall(api.nvim_win_call, self.win.preview_winid, function()
     -- start searching at line 1 in case we
     -- didn't reload the buffer (same file)
-    api.nvim_win_set_cursor(0, { 1, 0 })
-    fn.clearmatches()
-    fn.search(entry.ctag, "W")
-    if self.win.hls.search then
-      fn.matchadd(self.win.hls.search, entry.ctag)
+    if not entry.line or entry.line == 0 then
+      api.nvim_win_set_cursor(0, { 1, 0 })
+      fn.clearmatches()
+      fn.search(entry.ctag, "W")
+      if self.win.hls.search then
+        fn.matchadd(self.win.hls.search, entry.ctag)
+      end
+    else
+      -- the entry's line is assumed to be correct,
+      -- allowing the search to be skipped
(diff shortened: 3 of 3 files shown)


lua/fzf-lua/actions.lua
```diff
@@ -211,8 +211,12 @@ M.vimcmd_entry = function(_vimcmd, selected, opts, pcall_vimcmd)
           utils.jump_to_location(entry, "utf-16")
         end
       elseif entry.ctag then
-        vim.api.nvim_win_set_cursor(0, { 1, 0 })
-        vim.fn.search(entry.ctag, "W")
+        if entry.line ~= 0 then
```
ibhagwan: Could this entire code section not be avoided by changing the condition above (L213) to:
> Note there's already a call to `nvim_win_set_cursor` with line/col at the next `elseif`
```lua
 elseif entry.ctag and not entry.line then
```

lua/fzf-lua/make_entry.lua
```diff
@@ -439,6 +439,11 @@ M.tag = function(x, opts)
   -- from rg/grep output when using `tags_grep_xxx`
   local align = utils.has_ansi_coloring(name) and 47 or 30
   local line, tag = text:match("(%d-);?(/.*/)")
+  if not tag then
```
ibhagwan, line 442: Not sure I understand this part, in what format is there no tag but the line still matches?

Can you share the command you use to generate the tags so I can test it myself?

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -1131,11 +1131,17 @@ function Previewer.tags:set_cursor_hl(entry)
   pcall(api.nvim_win_call, self.win.preview_winid, function()
     -- start searching at line 1 in case we
     -- didn't reload the buffer (same file)
-    api.nvim_win_set_cursor(0, { 1, 0 })
-    fn.clearmatches()
-    fn.search(entry.ctag, "W")
-    if self.win.hls.search then
-      fn.matchadd(self.win.hls.search, entry.ctag)
+    if not entry.line or entry.line == 0 then
```
ibhagwan: Perhaps this section can also be  avoided and have this condition instead at the top?
```lua
if tonumber(entry.line) and entry.line > 0 then
  Previewer.buffer_or_file.set_cursor_hl(self, entry)
  end
end
```

ibhagwan (commented):
Unless I'm terribly missing something here I think some of the logic can be simplified.

ibhagwan:
Left a review, btw, I recall you mentioned you're using the `bat` previewer yet you're fixing the "builtin" previewer, is the `bat` previewer working as expected?

lua/fzf-lua/actions.lua
dvquentincarrel: It can indeed. I avoided relying on the next elseif because I didn't fully understand it, but I've tested it the way you suggested and the behavior seems to match.

ibhagwan: > It can indeed. I avoided relying on the next elseif because I didn't fully understand it, but I've tested it the way you suggested and the behavior seems to match.

It’s a simple clause that should apply when a line exists (with or without column).

lua/fzf-lua/make_entry.lua
dvquentincarrel: `ctags -R --excmd=number` produces tag lines which "contain" no tag after the processing at line 441. There, the "tag" is actually the regex used to find the tag's location in the file

lua/fzf-lua/previewer/builtin.lua
dvquentincarrel: I'm having trouble following you here.  
You mean, like that ?
```lua
function Previewer.tags:set_cursor_hl(entry)
  -- pcall(vim.fn.clearmatches, self.win.preview_winid)
  pcall(api.nvim_win_call, self.win.preview_winid, function()
    if tonumber(entry.line) and entry.line > 0 then
      Previewer.buffer_or_file.set_cursor_hl(self, entry)
    else
      -- start searching at line 1 in case we
      -- didn't reload the buffer (same file)
      api.nvim_win_set_cursor(0, { 1, 0 })
      fn.clearmatches()
      fn.search(entry.ctag, "W")
      if self.win.hls.search then
        fn.matchadd(self.win.hls.search, entry.ctag)
      end
      self.orig_pos = api.nvim_win_get_cursor(0)
      utils.zz()
    end
  end)
end
```

dvquentincarrel:
I'm sorry, I was wrong when I answered about the previewer some hours ago, it was not the `bat` one , but the default tags previewer. I edited my answer in hopes you wouldn't have seen it yet, but a bit too late it seems.

- the `bat` previewer works as expected with lines.
- the `builtin` previewer works as expected with lines.
- out of those three, only the default tags previewer didn't handle lines as expected

lua/fzf-lua/previewer/builtin.lua
ibhagwan: > I'm having trouble following you here.

Added the change in my followup commit

ibhagwan:
Merged, ty @dvquentincarrel for your conribution.

I took the liberty of fixing the `fixup!` I assume you wanted :-)

dvquentincarrel:
Thank you for your time @ibhagwan!

ibhagwan:
@dvquentincarrel just recalled the old issue where the ctags line number support was introduced: https://github.com/ibhagwan/fzf-lua/issues/355#issuecomment-1058830787.

Tested with the below `ctags` command, works as expected:
```sh
ctags -R --c++-kinds=+p --fields=+iaS --extras=+q --excmd=combine
```

dvquentincarrel:
You tested with or without the changes added by this PR?

ibhagwan:
> You mean that it works without the changes added by this PR?

I mean that it doesnt regress due to this PR.

In any event I just improved the support for `--excmd=combine` by adding tag highlighting: https://github.com/ibhagwan/fzf-lua/commit/9958d6a25d50f42b920664589a90704652169e5e.

PR #1558, opened 2024-12-08
fix(actions): file_switch with formatter

MeanderingProgrammer:
When a custom formatter is used the `file_switch` action along with any other that depend on it are never able to resolve to an existing buffer.

This is because the call that resolves a selected entry to a file does not provide options so the entry resolution does not know a formatter was used.


ibhagwan:
Good find @MeanderingProgrammer, Ty for the contribution!

PR #1578, opened 2024-12-17
fix(previewer): treesitter-context not closed in `live_grep`

phanen:
## Reproduce
1. Launch live_grep in `fzf-lua` repo, paste/type `vim.api.nvim_win_call, self.border_winid`.
2. Then type: `x`.
Nothing matched, but treesitter context window remain open.
image


ibhagwan:
Ty @phanen!

PR #1600, opened 2024-12-19
fix: scroll report error on fzf match nothing

phanen:
Scroll means `preview-page-down`/`preview-page-up` here.


ibhagwan:
I simplified this by preventing the entire scroll function to happen when the preview buffer is cleared.

Issue #1517, opened 2024-11-07
Feature: Filter specific symbols in lsp_document_symbols


wustho:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

Currently seems like lsp_document_symbols shows all symbols in buffer (classes, functions, variables). It would be nice to have a feature so that we can filter out some symbols like 'variables'. There was this feature in telescope some kinda like
```
    --                 lsp_document_symbols = {
    --                     ignore_symbols = {
    --                         'variable',
    --                         'property',
    --                     },
    --                 },
```

ibhagwan:
https://github.com/ibhagwan/fzf-lua/issues/1209

xixiaofinland:
@ibhagwan I know this is closed already, but I couldn't figure out how to exclude two strings.
Such as "Variable" and "Field", 
I attempt "Field|Variable" as below, but it seems to use the exact match rather than as "A or B" pattern

```
function() fzf.lsp_document_symbols({ regex_filter = { "Field|Variable", exclude = true } }) end
```

ibhagwan:
> @ibhagwan I know this is closed already, but I couldn't figure out how to exclude two strings. Such as "Variable" and "Field", I attempt "Field|Variable" as below, but it seems to use the exact match rather than as "A or B" pattern
> 
> ```
> function() fzf.lsp_document_symbols({ regex_filter = { "Field|Variable", exclude = true } }) end
> ```

The string is matched with lua’s `string.match` when used with `exclude` that would turn into `not string.match(…, “<pattern>”)`.

I don’t think lua has an OR operation in its patterns, use a function and instead and have your own filtering (you can also use `vim.regex` this way):
```lua
regex_filter = function(str)
  if str:match('a') or str:match('b') then return false end
end
```

There’s an example on how to use regex with this somewhere in the issues.

xixiaofinland:
i get it working like so just in case somebody needs it in the future.

```
                        regex_filter = function(item, _)
                            if item.kind:match('Struct') or item.kind:match('Enum') or item.kind:match('Method') then
                                return true
                            else
                                return false
                            end
                            -- if item.kind:match('Variable') or item.kind:match('Field')
                            -- then
                            --     return false
                            -- else
                            --     return true
                            -- end
                        end
```

ibhagwan:
Ty @xixiaofinland, although you can probably simplify this by using vim.regex.

Issue #441, opened 2022-06-11
[Feature Request] Filtering LSP Symbols


barrett-ruth:
In large files, `:FzfLua lsp_workspace_symbols` results in a rather large list of workspace symbols, most of which are of useless symbol types (to me at least).

Given that the list displays the type of the symbol according to the LSP (e.g. function), could this list be filtered according to symbol types?

A desired behavior would be similar to that of aerial.nvim's in which one can specify a `filter_kind` (table or string of symbol types) and pass such to the picker.

ibhagwan:
Have you tried `lsp_live_workspace_symbols`? Similar to `live_grep` it generates a new LSP query with your search term.

As for your request it’s arguably not going to help much, the issue here is having to wait for the LSP response (feeding fzf and filtering shouldn’t be an issue even for millions of entries), for that you can either use `async_or_timeout = true` option so that the UI doesn’t hang until you get a result or use `query = <term>` which will limit the LSP result to the query.

barrett-ruth:
I totally skipped over `lsp_live_workspace_symbols` and the query key. Thanks.

ibhagwan:
> I totally skipped over `lsp_live_workspace_symbols` and the query key. Thanks.

Not your fault, documentation is hard :)

barrett-ruth:
Actually, `lsp_live_workspace_symbols` isn't quite what I want. With lua_language_server, for example, the list is full of anonymous functions and useless variables. 

`lsp_document_symbols` however does exactly what I want. I am able to type "function," for example, and it matches all entries (correctly listed for the buffer I am in) that are actually of type "function." `lsp_live_workspace_symbols` does not.

So if `lsp_document_symbols` took a query parameter, that would be perfect. This would allow me to call it like so:
```lua
require('fzf-lua').lsp_document_symbols({ query = 'Function' })
```
effectively filtering by symbol type "function."

But... it doesn't take that parameter. I could (try to) submit a PR to do this if you think it is a worthy feature.

ibhagwan:
Take a look at #80, the issue here is that the LSP document symbols request doesn't support a query argument.

It was solved by adding the `current_buffer_only` flag, (exmaples from #80):
```lua
:lua require('fzf-lua').lsp_workspace_symbols({ current_buffer_only = true, query = 'edit' })
:lua require('fzf-lua').lsp_references({ current_buffer_only = true })
```

ibhagwan:
[...]

Issue #80, opened 2021-09-01
[ Request ] - lsp_document_symbols filter argument


ldelossa:
Hey there, 

I've never been a big fan out "outline" views in Vim as I can't get them to ever work well with split windows.

I've been experimentin with using `lsp_document_symbols` command to work similarily to a document outline, but obviously a bit flatter. 

What would be interesting is if I was able to do something similar to `FzfLua lsp_document_symbols METHODS` and output all the methods of the file. Repeat for any LSP kind.

ibhagwan:
It’s already possible with `workspace_symbols` by supplying `query`argument:
```lua
:FzfLua lsp_workspace_symbols query=METHODS
```
I will add this functionality to document symbols too, is this what you’re after?

ldelossa:
Would be great!

ldelossa:
> It’s already possible with `workspace_symbols` by supplying `query`argument:
> 
> ```lua
> :FzfLua lsp_workspace_symbols query=METHODS
> ```
> 
> ~I will add this functionality to document symbols too, is this what you’re after?~ I actually cannot add query to `document_symbol` as this is not supported by the LSP implementation [only the workspace_symbol function supports `{query}`](https://neovim.io/doc/user/lsp.html).

Ahh I see.

ibhagwan:
> > It’s already possible with `workspace_symbols` by supplying `query`argument:
> > ```lua
> > :FzfLua lsp_workspace_symbols query=METHODS
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ~I will add this functionality to document symbols too, is this what you’re after?~ I actually cannot add query to `document_symbol` as this is not supported by the LSP implementation [only the workspace_symbol function supports `{query}`](https://neovim.io/doc/user/lsp.html).
> 
> Ahh I see.

I can still easily do it, but the filtering would have to be done inside fzf-lua instead, try out the `lsp_live_workspace_symbols` provider which sends your queries to the LSP server as you type (similar to live grep).

If these results are what you’re after I can add a `current_buffer_only` flag to `lsp_workspace_symbols` you can use that instead of `lsp_document_symbols`.

ldelossa:
To emulate a replacement to an "outline" plugin, all symbols in the
document should be present first (not have to be searched for).

ibhagwan:
[...]

PR #1601, opened 2024-12-19
refactoring of README.md

gennaro-tedesco:
I have seen [this comment](https://www.reddit.com/r/neovim/comments/1hgvt7c/comment/m2nvwxx/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) of yours on reddit and have decided to come up with a proposal to help you clear up the docs. Let me try to explain the rationale of this PR before you get a heart attack by looking at the diff :D 

I have been a long-term Fzf-lua user and more or less know my way around this project, in the sense that I know where to find information I need; however, I sometimes do struggle navigating through the README, the Wiki and the issues just because of course there are so many options and contributions and it is hard to have a unified glance. For this reason I have proposed a strong refactoring of the README based on the below considerations:

1. the vast majority of users, the vast majority of times (say 90% of users 90% of times) exclusively look for configuration options, hence such must be the impact focus of the README, easily reachable, easily browsable. I have moved sections around to focus on Options first and foremost.
2. Installation and configuration instructions nowadays are done in lua only, I have removed all the bits that contain vimscript instructions (they are relevant for a minority of users only)
3. I have folder most if not all sections into `<details>` expandables: the README is very long and it cannot be browsed otherwise
4. I have separated the configuration options into different (foldable) sections according to the type: at the moment there is one long list that is difficult to search through
5. I have removed the table of contents because GitHub provides it already by itself (top right of the README rendering), so that we can save some "visual space"

As I said, this is a _strong_ refactoring of the README and it serves as proposal only, I did not mean to intentionally remove instructions or else, only instead to provide a basis for thoughts in case you would like to offer a leaner documentation (especially for first time users).

README diffs are hard to read, to see what it looks like directly in the browser hop on [the fork](https://github.com/gennaro-tedesco/fzf-lua/tree/readme_proposal?tab=readme-ov-file) and check it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,108 +4,38 @@
 
 ![Neovim version](https://img.shields.io/badge/Neovim-0.5-57A143?style=flat-square&logo=neovim)
 
-[Quickstart](#quickstart) • [Installation](#installation) • [Usage](#usage) • [Commands](#commands) • [Customization](#customization) • [Wiki](https://github.com/ibhagwan/fzf-lua/wiki)
+[Installation](#installation) • [Quickstart](#quickstart) • [Commands](#commands) • [Customization](#options) • [Wiki](https://github.com/ibhagwan/fzf-lua/wiki)
 
 ![Demo](https://raw.githubusercontent.com/wiki/ibhagwan/fzf-lua/demo.gif)
 
-[fzf](https://github.com/junegunn/fzf) changed my command life, it can change
-yours too, if you allow it.
+> “because you can and you love `fzf`” ([@junegunn](https://github.com/junegunn))
+
+> [fzf](https://github.com/junegunn/fzf) is the single most impactful tool for my command line workflow, it changed my life and it can change yours too, if you allow it.
 
 </div>
 
-# Contents
+## Installation
 
-- [Quickstart](#quickstart)
-- [Rationale](#rationale)
-- [Why Fzf-lua](#why-fzf-lua)
-- [Dependencies](#dependencies)
-  + [Optional Dependencies](#optional-dependencies)
-  + [Windows Notes](#windows-notes)
-- [Installation](#installation)
-- [Usage](#usage)
-  + [Resume](#resume)
-  + [Options](#options)
-- [Commands](#commands)
-  + [Buffers and Files](#buffers-and-files)
-  + [Search](#search)
-  + [Tags](#tags)
-  + [Git](#git)
-  + [LSP | Diagnostics](#lspdiagnostics)
-  + [Misc](#misc)
-  + [Neovim API](#neovim-api)
-  + [`nvim-dap`](#nvim-dap)
-  + [`tmux`](#tmux)
-  + [Completion Functions](#completion-functions)
-- [Customization](#customization)
-  + [Profiles](#profiles)
-- [Insert-mode Completion](#insert-mode-completion)
-  + [Custom Completion](#custom-completion)
-- [Default Options](#default-options)
-- [Highlights](#highlights)
-  + [Fzf Colors](#fzf-colors)
-- [Credits](#credits)
+[![LuaRocks](https://img.shields.io/luarocks/v/ibhagwan/fzf-lua?logo=lua&color=purple)](https://luarocks.org/modules/ibhagwan/fzf-lua)
 
+Using [lazy.nvim](https://github.com/folke/lazy.nvim)
 
-## Quickstart
-
-To quickly test this plugin without changing your configuration run (will run in it's own sandbox
-with the default keybinds below):
-> [!NOTE]
-> it's good practice to first
-> [read the script](https://github.com/ibhagwan/fzf-lua/blob/main/scripts/mini.sh)
-> before running `sh -c` directly from the web
-```sh
-sh -c "$(curl -s https://raw.githubusercontent.com/ibhagwan/fzf-lua/main/scripts/mini.sh)"
+```lua
+{
+  "ibhagwan/fzf-lua",
+  -- optional for icon support
+  dependencies = { "nvim-tree/nvim-web-devicons" },
+  config = function()
+    -- calling `setup` is optional for customization
+    require("fzf-lua").setup({})
+  end
+}
 ```
 
-| Key       | Command           | Key       | Command           |
-| ----------| ------------------| ----------| ------------------|
-| `<C-\>`     | buffers           | `<C-p>`     | files             |
-| `<C-g>`     | grep              | `<C-l>`     | live_grep         |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -338,250 +194,142 @@ to see detailed usage notes and a comprehensive list of all available options.**
...
-```
-
-### Default Options
-
-**Below is a list of most (still, not all default settings), please also
-consult the issues if there's something you need and you can't find as there
-have been many obscure requests which have been fulfilled and are yet to be
-documented. If you're still having issues and/or questions do not hesitate to open an
-issue and I'll be more than happy to help.**
+where each of the above sections can be fully configured as below:
```
vunhatchuong: I think this should explicitly mention that these are the default options since that is one of the most important aspects of a plugin.

Maybe something like: "Each of the above sections can be fully configured, with the following settings serving as the default options:" so that people can also search for the keyword `default`.

ibhagwan:
> before you get a heart attack by looking at the diff :D

Disclaimer noted, although at first glance I really like where this is going and this has been **long overdue**.

> I have been a long-term Fzf-lua user and more or less know my way around this project

We've conversed enough times before that I wouldn't forget you this quickly :-)

> As I said, this is a strong refactoring of the README and it serves as proposal only, I did not mean to intentionally remove instructions or else, only instead to provide a basis for thoughts in case you would like to offer a leaner documentation (especially for first time users).

I agree with all of your suggessions, we can work on this together until this gets to a point where everyone is happy, much appreciate the effort and thought you put into this.

> README diffs are hard to read, to see what it looks like directly in the browser hop on [the fork](https://github.com/gennaro-tedesco/fzf-lua/tree/readme_proposal?tab=readme-ov-file) and check it out.

I have a custom css that renders the markdown very closely to GitHub so I can also view this offline in my browser:
> https://github.com/ibhagwan/nvim-lua/blob/af41cafd37d962984b3f54fbf0d4fe58eb065f94/lua/plugins/init.lua#L36-L62

image


But since I added render-markdown.nvim (also very cool), I use it less:
image

milanglacier:
The current documentation does not clearly state the behavior when user has passed his custom keybindings. The behavior is that when user has passed the custom keybindings, then all the default keybindings will be removed. To keep the default the keybindings, the user must set the first key to true.

Currently the documentation states that:

```lua
require('fzf-lua').setup {
    actions = {
        -- Below are the default actions, setting any value in these tables will override
        -- the defaults, to inherit from the defaults change [1] from `false` to `true`
        files = {
            false, -- do not inherit from defaults
        },
    },

}
```

However, this statement is confusing. What does "Inherit from the defaults" mean? "Default" things are what should be inherited. "Default" and "inheritance" are synonymous in my mind. 

So in order to state it clearly, I personally think if we state the behavior in an opposite direction, that should be clearer, aka something like this: `If you pass any custom keybindings here, all the default keybindings will be cleared. If you want to keep using the default keybindings with your custom keybindings,  you should set the first key to true`

See the complete discussion at here: 

https://github.com/ibhagwan/fzf-lua/discussions/1612#discussioncomment-11626072

ibhagwan:
FYI, rebased to main so fork is up-to-date on the latest commits.

ibhagwan:
Added my 2 cents on this refactor, restored the "quickstart" to the top (I think it's important) and made other modifications, categories for commands, etc, see the end result in @gennaro-tedesco's [fork](https://github.com/gennaro-tedesco/fzf-lua/tree/readme_proposal?tab=readme-ov-file)

Comments ans suggesions welcome

gennaro-tedesco:
Another section that I believe we should improve is the `lsp` pickers options, in particular I have been browsing myself a few issues ([here](https://github.com/ibhagwan/fzf-lua/issues/441) [here](https://github.com/ibhagwan/fzf-lua/issues/1517) and [here](https://github.com/ibhagwan/fzf-lua/issues/80)) to understand how to filter by symbol; likewise with the options `query` and `lsp_query` (many questions about them in the discussion section).

I am not sure myself yet how to use such options, hence I would not want to add some wrong documentation. My understanding so far is that `lsp_query` is the query that is passed to nvim API `vim.lsp.buf.workspace_symbol(query=<lsp_query>)` and whose results are independent of fzf-lua (what the LSP returns or not in such cases depends exclusively on the LSP and how it indexes the project symbols). On the other hand, `query` is the fuzzy search query which one may want to pass as argument to somehow already open the picker and filter for some fuzzy terms. Is this understanding correct?

ibhagwan:
> I am not sure myself yet how to use such options, hence I would not want to add some wrong documentation. My understanding so far is that lsp_query is the query that is passed to nvim API vim.lsp.buf.workspace_symbol(query=<lsp_query>) and whose results are independent of fzf-lua (what the LSP returns or not in such cases depends exclusively on the LSP and how it indexes the project symbols). On the other hand, query is the fuzzy search query which one may want to pass as argument to somehow already open the picker and filter for some fuzzy terms. Is this understanding correct?

Correct, query always the fuzzy march query used by fzf, lsp_query is similar to the search/regex in grep as every new input changes the underlying LSP API query (like live grep).

gennaro-tedesco:
I have pushed a little commit adding the above explanation to the `lsp` options :)

ibhagwan:
> I have pushed a little commit adding the above explanation to the `lsp` options :)

I'm not sure about this change, first the query/lsp_query should not appear as these are part of the defaults, second, the default value is `nil` and not an empty string and I still don't find the explanaiton clear, I'll think of something.. left it commented for now.

ibhagwan:
Alright, I'm pretty happy with this now, gonna give this a couple of days before merging for comments and suggesions.

I think's there's a decent balance now between hidden/expandable sections and the "visible by default" ones, suggesions welcome.

gennaro-tedesco:
I will go through the latest changes this weekend to make sure we did not forget anything. I was thinking, moreover, to browse a little the latest discussion/issues to see if there are some settings/opts that users are frequently asking for, and that might mistakenly be missing in the current README.

ibhagwan:
> I will go through the latest changes this weekend to make sure we did not forget anything. I was thinking, moreover, to browse a little the latest discussion/issues to see if there are some settings/opts that users are frequently asking for, and that might mistakenly be missing in the current README.

There are a lot of things not in the README such as the distinction between grep’s search and queue and LSP query and lsp_query and all kinds of esoteric options, the thing is nit everything has to be in the README, there’s also the wiki and OPTIONS.md

ibhagwan:
I reviewer this throughly and made some cosmetic changes, merging.

If anyone has an other sugessions/clarifications feel free to continue the discussion here or make a new PR with your suggestions.

gennaro-tedesco:
Thank you for taking the time to work on this PR!

PR #1603, opened 2024-12-19
feat(commands): Auto run command if `nargs == "0"`

msaher:


ibhagwan:
Ty for the contribution @msaher but I don't think this should be added, sounds like a somewhat user-specific customization (run if 0, not otherwise), all for preventing a single `<CR>` press.

I believe the  place for this should be in the users' config not as yet another action, you can also reuse the existing code in your custom action:

```lua
require("fzf-lua").setup({
  command_history = {
    actions = {
        ["enter"] = function(selected, opts)
          local cmd = selected[1]
          local cmd_opts = vim.api.nvim_get_commands({})[cmd]
          if cmd_opts ~= nil and cmd_opts.nargs == "0" then
              actions.ex_run_cr(selected, opts)
          else
              actions.ex_run(selected, opts)
          end
        end

    }
  }
})
```

msaher:
No worries. It's worth mentioning that this is the default behavior of telescope, so maybe it's worth adding that snippet to the wiki or docs for users migrating from that plugin.

ibhagwan:
> No worries. It's worth mentioning that this is the default behavior of telescope, so maybe it's worth adding that snippet to the wiki or docs for users migrating from that plugin.

Wasn't aware of that, I personally find it confusing to have mixed functionality (either it should run or not), I'm also not keen on running commands without confirmation from the user.

> No worries. It's worth mentioning that this is the default behavior of telescope, so maybe it's worth adding that snippet to the wiki or docs for users migrating from that plugin.

Sure that's possible, but TBH I don't think it would be that big of a deal breaker, especially since my guess is most users don't even notice this.

PR #1614, opened 2024-12-20
introduce option to filter registers types

gennaro-tedesco:
Would you be open to consider the introduction of an option to filter register types?

This PR proposes a parameter:
```lua
registers = {
  filter = "..." -- "alpha" | "named" | ""
}
```
to only show a subset of the register types.

The rationale is that some users only need registers to paste specific content they just copied into 2-3 specific registers, and reducing the list may help glancing at what you need at once.

P. S. Unrelated to the PR (for future references): do you have pre-commit hooks or CI in place to format the code with `stylua`? I have run it locally but it produced a bigger diff than expected, not sure if `.stylua.toml` settings are not respected by my formatter or else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1353,6 +1353,12 @@ require'fzf-lua'.setup {
     -- you would set this option as %a this would match characters from [A-Za-z]
     -- or if you want to show only numbers you would set the pattern to %d (0-9).
   },
+  registers = {
+    filter = "", -- only show certain registers
+    -- "alpha" -> [a-z] registers
+    -- "named" -> [0-9] registers
+    -- leave empty for all
+  },
   complete_path = {
     cmd          = nil, -- default: auto detect fd|rg|find
     complete     = { ["enter"] = actions.complete },
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -269,14 +269,26 @@ M.registers = function(opts)
   opts = config.normalize_opts(opts, "registers")
   if not opts then return end
 
-  local registers = { [["]], "_", "#", "=", "_", "/", "*", "+", ":", ".", "%" }
+  local filter = opts.filter and opts.filter or ""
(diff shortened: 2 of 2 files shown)


ibhagwan:
> Would you be open to consider the introduction of an option to filter register types?

Why not, going to sleep but I’ll take a look tomorrow.

> P. S. Unrelated to the PR (for future references): do you have pre-commit hooks or CI in place to format the code with stylua? I have run it locally but it produced a bigger diff than expected, not sure if .stylua.toml settings are not respected by my formatter or else.

I use the LSP formatter that comes with lua_ls, it’s condig is in `.editorconfig`.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/6d7c08ed7871696764084350ab1803cb9ee9454e

I apologize @gennaro-tedesco, I found the approach to specific, I added this feature albeit differently and inline with other filters we have such as the  LSP symbols one, IMHO this provides the maximum flexibility as you can come up with any filter you want using lua string patterns or code:

From the commit msg:

Filter can be a function or a string if filter is a string will be
used to filter the table using `string.match`, if filter is a
function will be used to filter the registers table using
`vim.tbl_filter`.

Examples:
1. Only alphanumeric registers: `:FzfLua registers filter=%a`
2. Only non-alphanumeric: `:FzfLua registers filter=[^%a]`
3. Only "/*+", using a filter function:
```lua
:lua require("fzf-lua").registers({filter=function(r) return r:match("[/%*%+]") ~= nil end})
```

gennaro-tedesco:
You are right, using the filter is much better!

gennaro-tedesco:
By the way, there is a `print("r", r)` in the above commit before the function return: is it intentional or was it for debugging purposes?

ibhagwan:
> By the way, there is a `print("r", r)` in the above commit before the function return: is it intentional or was it for debugging purposes?

Debugging leftover lol

PR #1622, opened 2024-12-21
prototype for preview buffer for git_branches picker

gennaro-tedesco:
This PR is a prototype for a builtin preview for the `git_branches` picker in case you would be open to consider it.

1. The user can specify what type of `log_cmd` to display, in the demo below you can see different commands in action.
2. I have modified the original `cmd` for branches by adding `--format='%(refname:short)'`: it should help with the parsing of the branches names (removing `*`, spaces and all other symbols)

I might need some guidance for the following:

- can one use the treesitter parser to syntax highlight the git logs? I saw in the codebase that there is such option but I could not make it work
- if the user specifies a `log_cmd` that isn't a "git command", then the preview _may_ fail with some obscure errors: how can one ensure that for "bad commands" we catch the exception first and default to the default `self.opts.log_cmd`? I tried to `pcall` the `vim.system` but I do not what type of exception we want to capture.

Examples:

With the default `log_cmd` we have:
demo1

With 
```lua
git = {
  branches = {
    log_cmd = "git diff --stat main",
..
}
```
we have 
demo2

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -439,8 +439,9 @@ M.defaults.git                  = {
   },
   branches = {
     prompt     = "Branches> ",
-    cmd        = "git branch --all --color",
-    preview    = "git log --graph --pretty=oneline --abbrev-commit --color {1}",
+    cmd        = "git branch --all --color --format='%(refname:short)'",
+    log_cmd    = "git log --date=relative --date-order --pretty=format:%h%x09%an%x09%ad%x09%s",
+    previewer  = { _ctor = previewers.builtin.branches },
     actions    = {
       ["enter"]  = actions.git_switch,
       ["ctrl-x"] = { fn = actions.git_branch_del, reload = true },
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1503,4 +1503,36 @@ function Previewer.keymaps:populate_preview_buf(entry_str)
   Previewer.autocmds.super.populate_preview_buf(self, entry_str)
 end
 
+Previewer.branches = Previewer.buffer_or_file:extend()
+
+function Previewer.branches:keymaps(o, opts, fzf_win)
+  Previewer.autocmds.super.new(self, o, opts, fzf_win)
+  return self
+end
+
+function Previewer.branches:parse_log_cmd(entry_str, opts)
+  local log_cmd_table = {}
+  for w in opts.log_cmd:gmatch("%S+") do
+    table.insert(log_cmd_table, w)
+  end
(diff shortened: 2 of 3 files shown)


ibhagwan:
Ty @gennaro-tedesco!

> This PR is a prototype for a builtin preview for the git_branches picker in case you would be open to consider it.

That would be welcome, I’m sure some users will like the consistency in UI vs the fzf native previewer.

> can one use the treesitter parser to syntax highlight the git logs? I saw in the codebase that there is such option but I could not make it work

You can set the file type to `git` which is what fugitive uses for basic highlighting, otherwise there are all sorts of treesitter tricks as also now used in `blines|git_blame` (try it out, really cool) you can set treesitter for any part of the text:

https://github.com/ibhagwan/fzf-lua/blob/6d7c08ed7871696764084350ab1803cb9ee9454e/lua/fzf-lua/win.lua#L901-L951

> if the user specifies a log_cmd that isn't a "git command", then the preview may fail with some obscure errors: how can one ensure that for "bad commands" we catch the exception first and default to the default self.opts.log_cmd? I tried to pcall the vim.system but I do not what type of exception we want to capture.

I’d rename it to `cmd` to be consistent with all other commands / previewer settings, as for thr error I’d rather avoid the exception and display the detailed error message directly in the buffer and set it to plain text, similar to how inaccessible files or binary file preview info msg is displayed in the file/buffer previewer.

ibhagwan:
Another thing I’d prefer changing is avoid usage if `vim.system` as its synchronous and the experience will  be bad with a long running commands / big repo, I’d prefer using a terminal buffer and running the command that way which is async, there’s already framework laid out for such cases which is used for previewing images using chafa, etc:
https://github.com/ibhagwan/fzf-lua/blob/6d7c08ed7871696764084350ab1803cb9ee9454e/lua/fzf-lua/previewer/builtin.lua#L630

ibhagwan:
TBH, although I like this, unless we’re adding more add-value features to this I don’t see a real benefit other than having the same builtin previewer layout, while nice, all the other git previewers are native, git commits, status, blame, log, all of these use a native previewer (shell command inside fzf, which also comes with the benefit of piping into git-delta).

What is the real reasoning you wish to have this one different than all the other git previewers? Also aside from the git hash which can be highlighted there’s no treesitter benefit to the log message itself as it’s not code.

gennaro-tedesco:
The reasoning was to slowly provide builtin previewers for all providers, given that they are neovim buffers they would offer more configurability than the fzf previewer.

You're right of course that there is no added value besides a little more configurability and aesthetic, we can close the PR if there is no other idea on top.

ibhagwan:
If there’s no real problem why fix it :)

2+ years nobody complained it’s not a neovim buffer.

PR #1624, opened 2024-12-22
fix: type of `fzf_live` param

ernstwi:
I think this is right, matching the type of `fn_reload` in `fzf_exec`.


ibhagwan:
Ty @ernstwi

PR #1629, opened 2024-12-23
fix: `search_history` and `command_history` not showing custom actions

Bekaboo:
The headers of `search_history` and `command_history` are always set to `:: <CR> to execute, <Ctrl-e> to edit` despite custom configs, for example, users can disable `ctrl-e` for its original purpose (move cursor to the end of prompt). This PR dynamically set the header using `core.set_header()`.


ibhagwan:
Ty @Bekaboo!

PR #1642, opened 2024-12-26
fix: multi-line command desc not captured properly

rockyzhang24:
**Problem:**
If the builtin command description spans multiple lines, only the first line is captured and showed and the remaining lines won't be matched. I believe the issue lies in mistakenly writing `+` as `%+` in the regex.

**Solution:**
Correct the regex in `string.match()` and properly prepend a whitespace when concatenating multiple lines.


ibhagwan:
Ty @rockyzhang24!

PR #1660, opened 2025-01-01
fix: prompt should be concatenated with cwd_prompt

ANGkeith:
not sure if whether to classify this as `fix` or `feat`

hoping to achieve something like that in the prompt via 

```lua
files = {
      prompt = "❯ ",
      cwd_prompt = true,
}
```
image

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -889,7 +889,7 @@ M.set_header = function(opts, hdr_tbl)
 
   if not opts then opts = {} end
   if opts.cwd_prompt then
-    opts.prompt = normalize_cwd(opts.cwd or uv.cwd())
+    opts.prompt = opts.prompt .. normalize_cwd(opts.cwd or uv.cwd())
     if tonumber(opts.cwd_prompt_shorten_len) and
         #opts.prompt >= tonumber(opts.cwd_prompt_shorten_len) then
       opts.prompt = path.shorten(opts.prompt, tonumber(opts.cwd_prompt_shorten_val) or 1)


ibhagwan:
> not sure if whether to classify this as `fix` or `feat`

Ty for the contribution but I’d say neither, having both the prompt and cwd isn’t what I’d like to see when using the cwd as prompt, neither does this work that way in fzf.vim which is the original plugin I borrowed the idea from.

If you’d like this functionality for yourself you can easily concur both using `files({ prompt = … }).

ANGkeith:
cool thanks for the quick response and the plugin! 

> having both the prompt and cwd isn’t what I’d like to see when using the cwd as prompt, neither does this work that way in fzf.vim which is the original plugin I borrowed the idea from.

but would'nt that behavior be retained via
```lua
files = {
      prompt = "",
      cwd_prompt = true,
}
```
---
> If you’d like this functionality for yourself you can easily concat both using files({ prompt = … }).

you meant something like that right ? only minor issue is that, i'll be losing the built-in prompt shortening functionality
```lua
files = {
      prompt = "❯ " .. vim.uv.cwd(), 
}
```

ibhagwan:
> you meant something like that right ? only minor issue is that, i'll be losing the built-in cwd_prompt shortening functionality


You better do that in the call directly as a function otherwise you’re setting the prompt for all files calls globally,   query the cwd at runtime in case you change the cwd.

Set  your keymap to the below, use neovim’s `pathshorten` to shorten instead:
```lua
require("fzf-lua").files(function()
  return { cwd_prompt = false, prompt = "❯ " .. vim.fn.pathshorten(vim.uv.cwd()) }
end)

PR #1664, opened 2025-01-02
fix(lsp): fix invalid access in `fzf_lsp_locations`

abonnaudet-ledger:
`fzf_lsp_locations` was making an access to a nil value, for example when calling `lsp_incoming_calls` on a element that does not have any incoming calls.


ibhagwan:
Ty @abonnaudet-ledger, modified slightly as `core.__CTX` also needed to be reset here.

Issue #1696, opened 2025-01-10
Feature: Close dirty buffers


mononoken:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

Currently, you can delete buffers in a picker using the default keybind <C-x> which utilizes `buf_del`. I use this feature sometimes when using the Buffers picker to clean up my current session.

However, the current `buf_del` implementation only allows you to close clean buffers. For the majority of situations, this is fine. However, sometimes when I am using [vim-dadbod-ui](https://github.com/kristijanhusak/vim-dadbod-ui), I end up having lots of buffers with unsaved changes that I want to delete.

I would like fzf-lua to support closing these buffers too. Ideally, the default Neovim confirmation window can confirm we want to close the unsaved buffer.

I have a working fork branch that requires minimal changes to support this.

ibhagwan:
What if we use `utils.save_dialog` for this in similar fashion to the dirty buffer actions?

ibhagwan:
Close via #1697

PR #1697, opened 2025-01-10
Feat: enable closing dirty buffer

mononoken:
## Issue

https://github.com/ibhagwan/fzf-lua/issues/1696

## Description

This adds support to closing buffers using `buf_del` (`<C-x>`) for unsaved buffers as well. The current version only supports closing "clean" buffers.

It uses `utils.save_dialog` to confirm closing an unsaved buffer.

Let me know if there is anything that I missed or can be improved. Thank you!

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -405,8 +405,15 @@ M.buf_switch_or_edit = M.file_switch_or_edit
 M.buf_del = function(selected, opts)
   for _, sel in ipairs(selected) do
     local entry = path.entry_to_file(sel, opts)
-    if entry.bufnr and not utils.buffer_is_dirty(entry.bufnr, true, false) then
-      vim.api.nvim_buf_delete(entry.bufnr, { force = true })
+    if entry.bufnr then
+      if not utils.buffer_is_dirty(entry.bufnr, true, false) then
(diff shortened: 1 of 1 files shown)


ibhagwan:
Would we lose data this way? Take a look at the actions code, we can call `utils.save_dialog` to prompt the user to save/cancel, similar to the behavior with `:set confirm`.

mononoken:
@ibhagwan thank you for the quick reply! That sounds good to me. I took a look at `utils.save_dialog` and utilized it. Let me know if I missed anything else.

ibhagwan:
> @ibhagwan thank you for the quick reply! That sounds good to me. I took a look at `utils.save_dialog` and utilized it. Let me know if I missed anything else.

Looks good @mononoken, I’ll test it when not AFK and happily merge this, Ty for the co my contribution!

mononoken:
@ibhagwan awesome! Thank you for your work on this project!

ibhagwan:
@mononoken, it didn’t work, I had to call the safe dialog in the context of the buffer using `nvim_buf_call` :)

Ty for your contribution!

ibhagwan:
FYI, reduce the condition to one statement

https://github.com/ibhagwan/fzf-lua/commit/c238e015c9e3b1516ff6c1872131c8d050cc2c94

PR #1701, opened 2025-01-10
feat(config): allow setup via global variable

horriblename:
currently, fzf-lua takes 0.075ms on nvim startup if you don't use the setup function, but jumps up to 9.8ms if you use the setup function. It would be nice to keep the short startup time without needing an external lazy-loader (lazy.nvim).

still needs adding documentation, but I'd like to know what you think of the idea first.

prior art: [rustaceanvim](https://github.com/mrcjkb/rustaceanvim) uses `vim.g.rustaceanvim` for configuration (they also accept a function so you can do more advanced stuff but I don't think fzf-lua needs that)

### DIFF
--- a/lua/fzf-lua/cmd.lua
+++ b/lua/fzf-lua/cmd.lua
@@ -6,6 +6,10 @@ local serpent = require "fzf-lua.lib.serpent"
 
 local M = {}
 
+if _G.fzf_lua ~= nil then
+  builtin.setup(_G.fzf_lua)
+end
+
 function M.run_command(cmd, ...)
   local args = { ... }
   cmd = cmd or "builtin"


horriblename:
apparently nvim can't convert mixed table/list values (e.g. `{"max_perf", an_option = 3}`) into vim dictionaries so I used `_G.fzf_lua` instead of `vim.g.fzf_lua`. the only benefit of using `vim.g` is that vimL users can still use this plugin, but they couldn't do it before anyway so... should be fine?

ibhagwan:
I’m not a big fan of using globals, feels somewhat “pollutionary” to me.

Not sure I understand the use case, this is not an issue with a plugin manager, if you’re already using your own lazy loading why not call setup before using fzf-lua for the first time instead of relying on this global.

Something as simple as wrapping all your fzf lua calls with:
```lua
If not package.loaded["fzf-lua"] then
 -- call setup
end
-- call a picker
```

horriblename:
I think it's a better user experience to get good startup times out of the box. Plus, if most people are going to lazy load the plugin anyways, why not just do it in the plugin so nobody else has to? 

On the point of global variables, I do think that having two ways to configure might not be all that beneficial. I could instead try to move the heavy lifting out of init.lua and the setup function so the setup actually finishes quickly, would that be better?

ibhagwan:
> I think it's a better user experience to get good startup times out of the box. Plus, if most people are going to lazy load the plugin anyways.

But there is good user experience out of the box, setup won’t be called automatically and the icon package won’t be loaded unless you call setup with icons.

> why not just do it in the plugin so nobody else has to?

Because this adds yet just another issue/point of failure/esoteric option nobody uses and nobody knows about, the plug-in exists almost 3 years and this hasn’t come up, 99% or the users don’t know nor care about this (or use a plug-in manager).


> On the point of global variables, I do think that having two ways to configure might not be all that beneficial. I could instead try to move the heavy lifting out of init.lua and the setup function so the setup actually finishes quickly, would that be better?

I don’t wish to add this, if someone decides to not use a plug-in manager and still lazy load they can always the 2 lines code I wrote in the other comment to “lazy setup” as well.

horriblename:
understandable, thanks for the reply!

Issue #1306, opened 2024-06-28
Feature:  Support transparent gutter via highlight group


scottmckendry:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

I'm currently in the process of adding support for Fzf-lua to my colorscheme and noticed that setting the `bg` property of `FzfLuaFzfGutter` to `"NONE"` does not make it transparent. Setting it to a solid color works fine, however.

Is this something that could be supported in the future?

ibhagwan:
Fzf runs in a neovim terminal, the gutter color needs to be translated into an ansi escape sequence and `NONE` has no color, while I could translate it manually to `-1` it would be hacky.

Why don’t you set your `fzf_colors = { ['gutter’] = '-1' }`, if you wanted to do so without calling setup (e.g. in a color scheme) you can set `vim.g.fzf_colors`:
https://github.com/ibhagwan/fzf-lua?tab=readme-ov-file#fzf-colors

scottmckendry:
Thanks! I think I can make that work with the vim.g variable.

ramojus:
This doesn't seem to be working anymore. Even if I set this color via command with `:lua require'fzf-lua'.files({ hls={["gutter"] = "-1"} })`, the gutter is still not transparent:
![image](https://github.com/user-attachments/assets/3250e3db-bdda-43ba-8aaf-75af6da07c72)

ibhagwan:
[...]

PR #1709, opened 2025-01-12
feat: allow clearing a color with highlight group

ramojus:
Implements https://github.com/ibhagwan/fzf-lua/issues/1306.

I couldn't confidently determine if the value really was `"NONE"` - the method used by this plugin returns `nil` and `vim.api.nvim_get_hl(0, { name = <hlgroup> })` returns `vim.empty_dict()`. 

With this change, if the highlight links to a non-existing group, it will be transparent. I couldn't think of a way to avoid this.

EDIT: I was planning to investigate this a bit more, but accidentally submitted the PR. I think this could be a valid solution, so I'm leaving it as is for now.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -23,6 +23,29 @@ M.resume_get = function(what, opts)
   return utils.map_get(M, what)
 end
 
+M.get_default_colors = function()
+  local fzf_hls = require("fzf-lua.defaults").defaults.__HLS.fzf
+  return {
+    ["fg"]        = { "fg", fzf_hls.normal },
+    ["bg"]        = { "bg", fzf_hls.normal },
+    ["hl"]        = { "fg", fzf_hls.match },
+    ["fg+"]       = { "fg", { fzf_hls.cursorline, fzf_hls.normal } },
+    ["bg+"]       = { "bg", fzf_hls.cursorline },
+    ["hl+"]       = { "fg", fzf_hls.match },
+    ["info"]      = { "fg", fzf_hls.info },
+    ["border"]    = { "fg", fzf_hls.border },
+    ["gutter"]    = { "bg", fzf_hls.gutter },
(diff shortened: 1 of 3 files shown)


ibhagwan:
> I couldn't confidently determine if the value really was "NONE"

> EDIT: I was planning to investigate this a bit more, but accidentally submitted the PR. I think this could be a valid solution, so I'm leaving it as is for now.

Obviously the best would be if can accurately determine the hlgroup was set to `NONE`, I'm wondering if that's possible, I'll do some quick tests and also think on potential downsides to setting a color to `-1`.

ibhagwan:
I also couldn't find a way between `NONE` values and not having the value defined at all, not with lua API nor vimL.

The issue with setting `-1` in this case is that it nullifies the fzf colorscheme, say someone has a NONE def for the fg of the `Special` hlgroup, by default (with `fzf_colors=true`) it will be linked to `FzfLuaFzfMatch` which will then nullify all fuzzy matching colors  (bad, can't match anything).

I'm not sure we can allow this type of downside vs forcing the user to add `-1` to their `fzf_colors`.

ramojus:
I agree that that would not be a good behavior.
What if we checked whether the highlight definition is still the default one and only set `-1` if that's not the case? Shouldn't be too difficult to do.

ramojus:
More of a draft and needs more testing. This might still be strange, if the users definition matches the default one...
We could also mark the places where user specifically sets their own colors and setup an autocmd to check if user changes highlights (is there such event though?).

Anyways, I'll be waiting for your thoughts. Feel free to close the PR now if you think this is too much & not worth it, as I don't have anything better to suggest.

ibhagwan:
Appreciate the effort @ramojus but it feels very hacky and doesn’t cover everything, I think this change will have to wait until there is better detection for NONE in neovim, until then I can live with having to specify `-1`.

Issue #1417, opened 2024-08-27
Bug: Misaligned entries in `helptags` output using Lazy.nvim


marcinmdev:
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [X] I have read the Wiki including the Advanced section
- [X] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Windows + Linux

### Shell

zsh, powershell

### Neovim version (`nvim --version`)

NVIM v0.10.1

### Fzf version (`fzf --version`)

0.54.3 (af4917db)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--no-separator --layout=reverse --inline-info

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [X] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
{
  "ibhagwan/fzf-lua",
  config = function()
    require("fzf-lua").setup({ "fzf-native", fzf_colors = true })
  end,
,
```

### Describe the bug / steps to reproduce

Helptags output is misaligned probably due to emoji encoding messing with line width. 

Steps to reproduce:
1. Use https://github.com/folke/lazy.nvim plugin manager
2. install fzf-lua
3. execute `:FzfLua helptags`

Output:

![2024-08-27_13-08-1724759195](https://github.com/user-attachments/assets/0ec5d956-e4da-42d5-89d3-70be8f822b46)

Tested in alacritty and Neovide (Arch Linux) and Neovide and windows terminal in Windows 11. Using FantasqueSansM Nerd Font Mono, tested with Noto Sans Mono and default Neovide font (which was Fira Code Nerd Font at the time).

Aditionally scrolling the output messes it even more:

![2024-08-27_13-08-1724759766](https://github.com/user-attachments/assets/d83720e4-5b4c-4a63-a8ca-5e06631f6541)

ibhagwan:
From the screenshot I can see the native fzf previewer for helptags is  used which means the border is drawn by the fzf process itself, this will have to be fixed upstream.

ibhagwan:
Closing as upstream.

PR #1716, opened 2025-01-16
fix(helptags): handle multibyte char align

phanen:
Native lua `string.format` cannot handle multi-byte char
Related: https://github.com/ibhagwan/fzf-lua/issues/1417

Before:
image

Now:
image

### DIFF
--- a/lua/fzf-lua/providers/helptags.lua
+++ b/lua/fzf-lua/providers/helptags.lua
@@ -60,7 +60,9 @@ M.helptags = function(opts)
     end)()
 
     local add_tag = function(t, fzf_cb, co)
-      local tag = string.format("%-80s %s%s%s", hl(t.tag), t.filename, utils.nbsp, t.filepath)
+      local w = 80 + string.len(t.tag) - vim.fn.strwidth(t.tag)
+      local tag = string.format("%-" .. w .. "s %s%s%s", hl(t.tag), t.filename, utils.nbsp,
+        t.filepath)
       fzf_cb(tag, function()
         coroutine.resume(co)
       end)


ibhagwan:
That’s great @phanen!

I recently added a minimal [utf8 library](https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/lib/utf8.lua) for these type of issues, what do you think if we change the code (and simplify) using `require("fzf-lua.lib.utf8").len()`?

ibhagwan:
> The problem is that `string.format` with `%-80s` just fill with whitespace up to "80 bytes" rather than "80 real display cell" (not sure the right term).
> 
> With `string.len` we get the number of bytes, so what needed here is just the real width. (no matter how many bytes count for a mult-byte char, it may take different width). We can just use vim(v7.3+) backended `strwidth` for this case.
> 
> Here's some test:
> 
> ```lua
> local str =
>   '🚀✨⚡️Ｈｅｌｌｏ你好あいL̓̉̑̒̌̚ơ̗̌̒̄̀ŕ̈̈̎̐̕è̇̅̄̄̐m̖̟̟̅̄̚'
> print(vim.fn.strlen(str)) -- 105
> print(string.len(str)) -- 105
> print(vim.str_utfindex(str, 'utf-8')) -- 105
> print(require('fzf-lua.lib.utf8').len(str)) -- 48
> print(vim.str_utfindex(str, 'utf-16')) -- 49
> print(vim.str_utfindex(str, 'utf-32')) -- 48
> print(vim.api.nvim_strwidth(str)) -- 29
> print(vim.fn.strwidth(str)) -- 29
> ```
> 
> Also try `nvim_strwidth`, but for unknown reason this way need `schedule_wrap` on `add_tags` to work properly.

Interesting that strwidth returns the correct visible len and utf8/32 doesn’t.

PR #1717, opened 2025-01-16
fix: disable netrw to avoid piping `!wget ..` to stdout

phanen:
Sometime pasting long url cause `!wget https:/xxx` appear in fzf win.

To reproduce with `mini.sh`
* `FzfLua live_grep`
* Paste a long url `https://a.com/bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb`
image

Just disable `netrw` then it seems fixed.

### DIFF
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -769,7 +769,8 @@ M.wrap_spawn_stdio = function(opts, fn_transform, fn_preprocess, fn_postprocess)
         _is_win and [[set VIMRUNTIME=%s& ]] or "VIMRUNTIME=%s ",
         _is_win and vim.fs.normalize(vim.env.VIMRUNTIME) or M.shellescape(vim.env.VIMRUNTIME)
       )
-  local lua_cmd = ("lua %sloadfile([[%s]])().spawn_stdio(%s,%s,%s,%s)"):format(
+  local lua_cmd = ("lua %s%sloadfile([[%s]])().spawn_stdio(%s,%s,%s,%s)"):format(
+    "vim.g.loaded_netrwPlugin=1;",
     _has_nvim_010 and "vim.g.did_load_filetypes=1; " or "",
     _is_win and vim.fs.normalize(__FILE__) or __FILE__,
     opts, fn_transform, fn_preprocess, fn_postprocess
--- a/lua/fzf-lua/shell.lua
+++ b/lua/fzf-lua/shell.lua
@@ -101,7 +101,8 @@ function M.raw_async_action(fn, fzf_field_expression, debug)
   local action_cmd = ("%s%s -n --headless --clean --cmd %s -- %s"):format(
     nvim_runtime,
     libuv.shellescape(path.normalize(nvim_bin)),
-    libuv.shellescape(("lua %sloadfile([[%s]])().rpc_nvim_exec_lua({%s})"):format(
+    libuv.shellescape(("lua %s%sloadfile([[%s]])().rpc_nvim_exec_lua({%s})"):format(
+      "vim.g.loaded_netrwPlugin=1;",
       utils.__HAS_NVIM_010 and "vim.g.did_load_filetypes=1; " or "",
       path.join { vim.g.fzf_lua_directory, "shell_helper.lua" },
       call_args


ibhagwan:
Interesting, do you think this is an upstream issue as well? Why would netrw stdout the string?

phanen:
Not sure, what I know is only `netrw` will print out these stuffs.
```sh
$ nvim --headless --clean https://www.youtube.com/
:!wget 'https://www.youtube.com/' -q -O '/tmp/nvim.phan/oFFSXD/0'
^C⏎
$ nvim --headless --clean https://www.youtube.com
**error** (netrw) not a netrw-st...://[user@]hostname[:port]/[path])^C⏎
```

ibhagwan:
I can see this doesn't happen with:
```sh
nvim --headless -u NONE https://www.youtube.com
```

Wondering if there would be a downside to change all our headless instances to `-u NONE` instead of `--clean`, then we can also ditch the `vim.g.did_load_filetypes=1`?

What do you think?

ibhagwan:
Or better yet, convert the headless instances to use the new `-l` or `-ll`?

ibhagwan:
Did a preliminary test, I think this requires 0.8/0.9 (no problem), using `-l` (or better yet `-ll`) is much better than using `--cmd`, if we’re already touching a sensitive part such as the headless instance might as well upgrade to lua loop threading.

ibhagwan:
@phanen, try https://github.com/ibhagwan/fzf-lua/tree/refactor_spawn, I modified the flags to use `-u NONE -i NONE` instead of `--clean`, will need more testing as there may be uninteded consequences.

I also made the perparation to use `-l` instead, checkout "scripts/headless_fd.sh" and "spawn.lua", the only issue I wasn't able to solve yet is waiting for the `uv.spawn` to finish (it currently exists as soon as it finishes sourcing spawn.lua.

I want to see the difference in performance between the two to understand if this is worth the effort.

phanen:
I think no problem to migrate to `-u NONE`, which just disable vim's default plugins (unused in fzf-lua).
```
$ hyperfine -w 5 './scripts/headless_fd2.sh -c ~/b/linux/'
Benchmark 1: ./scripts/headless_fd2.sh -c ~/b/linux/
  Time (mean ± σ):     428.4 ms ±  18.3 ms    [User: 868.7 ms, System: 133.3 ms]
  Range (min … max):   412.8 ms … 478.6 ms    10 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to us
e the '--warmup' or '--prepare' options.
 
$ hyperfine -w 5 './scripts/headless_fd.sh -c ~/b/linux/'
Benchmark 1: ./scripts/headless_fd.sh -c ~/b/linux/
  Time (mean ± σ):     428.8 ms ±  19.7 ms    [User: 870.1 ms, System: 132.2 ms]
  Range (min … max):   415.1 ms … 481.7 ms    10 runs
```

phanen:
* `-l` should be `-u NONE -l` in future: https://github.com/neovim/neovim/issues/31878
* No idea about how to avoid `vim.wait`, and there's a `MAX_TIMEOUT` for 32-bit machine.
https://github.com/neovim/neovim/blob/ac3859a4410e50794a083f23796e4f8ae2a24b04/runtime/lua/vim/_system.lua#L90.

ibhagwan:
> I think no problem to migrate to -u NONE, which just disable vim's default plugins (unused in fzf-lua).

I’ll test this some more, I also eliminates the `require` hack by separating the code in spawn.lua.

> No idea about how to avoid vim.wait, and there's a MAX_TIMEOUT for 32-bit machine.

I know about the max timeout but that’s just a hack I used, unless I can figure out a proper solution I won’t use it.

From your quick profiling doesn’t seem there’s a real gain in performance.

ibhagwan:
Doesn't seem like the `-l` or `-ll` is worth the hassle, in terms of performance there's better ROI investing our time in optimizing `make_entry.file` processing (string manipulations, etc) which would have a much bigger impact.

Closed via #1720.

phanen:
I debug this again, even without `netrw`, if filename >= max unix path length (256 ?). Neovim will report a error:
```
nvim -u NONE --headless aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
```
<aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" [Permission Denied]
```

Although this "Permission information" looks like incorrect... yet another upstream issue. 
https://github.com/neovim/neovim/blob/365b865d6b92f3035e95ddbc1a7d111413cbe141/src/nvim/fileio.c#L503

Issue #1686, opened 2025-01-08
Feature: Improved FzfLua resume


miroshQa:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

Continuation of https://github.com/ibhagwan/fzf-lua/issues/1595

## Problem:
Currently, ":FzfLua resume" is very limited compared to ":Telescope resume" because all picker actions that select an entry close the fzf process, so ":FzfLua resume" works by rerunning the picker again. However, rerunning the picker again is inefficient, and we also lose the selected entry position.

## Solution:
All "select entry" actions should hide the picker by default instead of closing the process and performing the corresponding action (open file, etc.).  It turned out to be possible to implement according to the discussion in https://github.com/ibhagwan/fzf-lua/issues/1595.

miroshQa:
I also think it would be good to swap 'Esc' and 'A-Esc.' So, 'Esc' should hide the picker by default, and 'A-Esc' should terminate. That would provide better defaults in my opinion

ibhagwan:
> I also think it would be good to swap 'Esc' and 'A-Esc.' So, 'Esc' should hide the picker by default, and 'A-Esc' should terminate. That would provide better defaults in my opinion

I’m going to create a profile (name suggestion welcome) that if setup will change this behavior for the user.

You’ll be able to set it from `:FzfLua profiles` or:
```lua
require("fzf-lua").setup({"hide"}) -- or whatever name we come up with 
```

We might beed two profiles one for hide and another for “sticky” for what you wanted in the other issue.

miroshQa:
Is the "hide" profile about the closing process and "sticky" about hiding the window? Did I understand that correctly?

Will the "sticky" profile be the default? I just don't get why anyone would want to use the "hide" profile then if it only has downsides, like not being able to fully use FzfLua resume. Or did I miss something, and it could actually be better for someone because we free up memory right after the picker closes, for example?

ibhagwan:
> @ibhagwan
> If you select only one entry each time, they are the same.
> If you select multiple entries(--multi)

> exec_silent does nothing but hide when you hide enter
> reload also cancel all your selection.
> I think that's expected behavior for fzf: execute-silent won't do anything to tui while reload actually replace your list?

@phanen, just did some testing, `exec_silent` seems to work pretty good, I can’t reproduce what you’re describing here, the only downside is an extremely long command with a reserved function id for every bind.

[...]

PR #1721, opened 2025-01-17
feat(hide): keep scroll position

phanen:
Would be great to resume scroll postion when resume after hide?
* related: https://github.com/ibhagwan/fzf-lua/issues/1686


ibhagwan:
Ty @phanen!

PR #1722, opened 2025-01-18
fix: merge existed `fzf_opts["--preview_window"]`

phanen:
And `opts.preview_window` seems unused here.

e.g. When set `fzf_opts["--preview-window"] = "border-left"`, while `FzfLua git_status` still open fzf's previewer with the default border.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -382,7 +382,7 @@ M.fzf = function(contents, opts)
       -- (1) when using a split we use the previewer as placeholder
       -- (2) we use 'nohidden:right:0' to trigger preview function
       --     calls without displaying the native fzf previewer split
-      opts.fzf_opts["--preview-window"] = previewer:preview_window(opts.preview_window)
+      opts.fzf_opts["--preview-window"] = previewer:preview_window(opts.fzf_opts["--preview-window"])
     end
     -- provides preview offset when using native previewers
     -- (bat/cat/etc) with providers that supply line numbers
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -433,11 +433,11 @@ function Previewer.base:zero(_)
   return act
 end
 
-function Previewer.base:preview_window(_)
+function Previewer.base:preview_window(old_preview_window)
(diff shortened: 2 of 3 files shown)


ibhagwan:
> And opts.preview_window seems unused here.

That’s an old option and correctly should be removed.


> e.g. When set fzf_opts["--preview-window"] = "border-left", while FzfLua git_status still open fzf's previewer with the default border.

I don’t think we should be meeting these options, it’s just a a source of issues, that’s why I added winopts.preview.border to accept fzf and neovim win styles`:FzfLua git_status winopts.preview.border=border-left`.

https://github.com/ibhagwan/fzf-lua/commit/fc70792718d4a41101956c10ec04b6bf6a258777

phanen:
> I don’t think we should be merging these options, it’s just a a source of issues, that’s why I added winopts.preview.border to accept fzf and neovim win styles:FzfLua git_status winopts.preview.border=border-left.

This work if just set `git.status.winopts.preview.border`.

If I set a global `preview.border="border-left"`, this seems override neovim builtin previewer border config (e.g. { '┏', '━', '┓', '┃', '┛', '━', '┗', '┃' }, work for builtin previewer, but not fzf's).

There seems lack a global config to control fzf's previewer specially.

ibhagwan:
> There seems lack a global config to control fzf's previewer specially.

That’s tough because some previewers use neovim buffer preview and some native (even on the same profile), that’s why you can set `border = function() … end` and return different borders for both neovim and fzf, see how it’s done in the “border-fused” profile, you can even return different borders for different  layouts (you probably don’t want border-left if vertical):
https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/profiles/border-fused.lua

phanen:
Awesome, be simply handled by:
```lua
          border = function(_, m)
            return m.type == 'fzf' and 'border-left'
              or { '┏', '━', '┓', '┃', '┛', '━', '┗', '┃' }
          end,

```
So I think this pr can be closed...

ibhagwan:
> So I think this pr can be closed...

Great, closing.

Up to you if you want to make a new PR to remove opts.preview_window or I can do it tomorrow.

phanen:
Ok, I'll PR then.

PR #1723, opened 2025-01-18
chore: cleanup deprecated things

phanen:
* legacy `opts.preview_window` -> `fzf_opts.--preview_window`
* `vim.fn.termopen` -> `vim.fn.jobstart` + `{ term = true }`
* `vim.loop` -> `vim.uv`
* `vim.lsp.diagnostics.get_line_diagnostics` -> `vim.lsp.diagnostic.get`
* remove term workaround 9a755f225bc6 (fixed now)
* `opts.silent` when `reload` fallback to `resume`

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -382,7 +382,7 @@ M.fzf = function(contents, opts)
       -- (1) when using a split we use the previewer as placeholder
       -- (2) we use 'nohidden:right:0' to trigger preview function
       --     calls without displaying the native fzf previewer split
-      opts.fzf_opts["--preview-window"] = previewer:preview_window(opts.preview_window)
+      opts.fzf_opts["--preview-window"] = previewer:preview_window()
     end
     -- provides preview offset when using native previewers
     -- (bat/cat/etc) with providers that supply line numbers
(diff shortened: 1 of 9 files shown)


lua/fzf-lua/core.lua
```diff
@@ -1074,7 +1074,7 @@ M.convert_reload_actions = function(reload_cmd, opts)
       end
     end
   end
-  if has_reload and reload_cmd and type(reload_cmd) ~= "string" then
+  if opts.silent ~= true and has_reload and reload_cmd and type(reload_cmd) ~= "string" then
     utils.warn(
       "actions with `reload` are only supported with string commands, using resume fallback")
```
phanen, line 1079: This warn me every time when I use lsp picker seems due to I set a global `actions.files` with reload actions.

phanen: Oh, and oldfiles.

lua/fzf-lua/providers/lsp.lua
```diff
@@ -509,7 +509,8 @@ local function gen_lsp_contents(opts)
       if utils.tbl_isempty(results) then
         if not opts.fn_reload and not opts.silent then
           utils.info(string.format("No %s found", string.lower(lsp_handler.label)))
-        else
+        end
+        if opts.fn_reload then
```
phanen, line 510: Seems a bug introduced from (a81d18ee): `FzfLua lsp_xxx silent=true` will always launch fzf even when no results.

ibhagwan:
Tysm @phanen, this is very helpful.

Did tiny 2 modifications:
- we don’t need the preview_window in the called function so I removed it
- Changed the LSP fn_reload/silence bug condition slightly (to be more coherent)

PR #1737, opened 2025-01-21
fix: improve manpage defaults

phanen:
Not sure if this is a issue only happen to me...
```sh
$ man gcc | head
troff:<standard input>:127: warning: cannot select font 'CW'
troff:<standard input>:127: warning: cannot select font 'CW'
troff:<standard input>:128: warning: cannot select font 'CW'
troff:<standard input>:128: warning: cannot select font 'CW'
troff:<standard input>:128: warning: cannot select font 'CW'
troff:<standard input>:129: warning: cannot select font 'CW'
```

And `man -k .` may truncate entries
```sh
man -k . | grep --fixed-strings '... ('
```

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -25,20 +25,18 @@ function M._preview_pager_fn()
 end
 
 function M._man_cmd_fn(bat_pager)
-  local cmd = utils.is_darwin() and "man -P cat {page} | col -bx"
-      or vim.fn.executable("mandb") == 1 and "man {page} | col -bx"
-      or "man -c {page} | col -bx"
-  if bat_pager then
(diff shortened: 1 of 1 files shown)


lua/fzf-lua/defaults.lua
```diff
@@ -794,7 +792,7 @@ M.defaults.helptags             = {
 }
 
 M.defaults.manpages             = {
-  cmd       = "man -k .",
+  cmd       = "man -kl .",
```
ibhagwan: I don’t think we should change the default command, man has proven to be very problematic and works differently on many systems (manioc/mandb/aprorpos), on my Linux `man -kl .` doesn’t work.

ibhagwan:
Ty @phanen!

PR #1743, opened 2025-01-22
fix: set TMPDIR for child to avoid lots of logging

phanen:
Related https://github.com/neovim/neovim/pull/32137.

### DIFF
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -717,6 +717,9 @@ M.expand = function(s)
   return vim.fn.expand(s)
 end
 
+
+local TMPDIR = vim.fn.fnamemodify(vim.fn.tempname(), ":h:h:h")
+
 ---@param opts string
 ---@param fn_transform string?
 ---@param fn_preprocess string?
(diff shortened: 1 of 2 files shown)


ibhagwan:
I'm not sure we need this, it has never been an issue on any system, isn't it better we wait for the upstream to resolve?

Where exectly do you encounter this and what are the consequences?

phanen:
Yesterday I rm my log, and now my log have been filled with:
```
WRN 2025-01-21T12:00:57.881 nvim.143483.0 vim_mktempdir:3280: $TMPDIR is unset
```

```sh
$ wc ~/.local/state/nvim/log
  30540  227376 2654748 .local/state/nvim/log
```

I also think it would be better to fix this in upstream... @justinmk?
If this won't be a problem for others, I may consider set a `vim.env.TMPDIR` in my config.

ibhagwan:
I don’t have TMPDIR either, could this have a side effect and now instead of a harmless log message we will encounter some more sever issue?

Say TMPDIR isn’t accessible by the headless instance running under some other context (neovide, chrome, etc), what if this will now fail the process altogether?

ibhagwan:
Tested this on both Linux and Mac with no issues, hopefully doesn’t regress on Windows (don’t see why it should).

Ty @phanen, for your insights :)

lua/fzf-lua/libuv.lua
```diff
@@ -731,13 +734,18 @@ M.wrap_spawn_stdio = function(opts, fn_transform, fn_preprocess, fn_postprocess)
...
+    _is_win and [[set TMPDIR=%s& ]] or "TMPDIR=%s ",
+    _is_win and vim.fs.normalize(TMPDIR) or M.shellescape(TMPDIR))
+
   local lua_cmd = ("lua %sloadfile([[%s]])().spawn_stdio(%s,%s,%s,%s)"):format(
     _has_nvim_010 and "vim.g.did_load_filetypes=1; " or "",
     vim.fn.fnamemodify(_is_win and vim.fs.normalize(__FILE__) or __FILE__, ":h") .. "/spawn.lua",
     opts, fn_transform, fn_preprocess, fn_postprocess
   )
-  local cmd_str = ("%s%s -n --headless -u NONE -i NONE --cmd %s"):format(
+  local cmd_str = ("%s%s%s -n --headless -u NONE -i NONE --cmd %s"):format(
```
justinmk, line 746: I didn't check how cmd_str is actually invoked, but presumably you could avoid the hassle of setting TMPDIR by simply passing the environment to the job that is started. Seems like that isn't happening.

ibhagwan: > I didn't check how cmd_str is actually invoked, but presumably you could avoid the hassle of setting TMPDIR by simply passing the environment to the job that is started. Seems like that isn't happening.

Can’t do that since these commands are used as fzf callbacks, piped into fzf or used as FZF_DEFAULT_COMMAND, in addition they will be wrapped with `sh -c` (or `cmd.exe` on Windows), which makes it extra tricky for env vars, this has been the most robust way to make everything work.

justinmk: Fzf's "vim layer" doesn't accept functions instead of a string?

ibhagwan: > Fzf's "vim layer" doesn't accept functions instead of a string?

Yes, but for process separation fzf needs to call a shell string which in turn calls a lua callback which can also be a lua function, what you’re looking at is the abstraction layer of multi process separation, that’s what gives fzf-lua the responsiveness and speed.

Issue #329, opened 2022-01-27
fzf-lua files leaves dangling neovim /tmp directories when executed in a git repository


gravndal:
Steps to reproduce:
1. `$ mkdir -p fzf-lua-repro/.config/nvim`
2. `$ cd fzf-lua-repro`
3. `$ curl https://raw.githubusercontent.com/ibhagwan/fzf-lua/main/minimal_init.lua > .config/nvim/init.lua`
4. `$ git init`
5. `$ env HOME=$(pwd) nvim`
6. `:FzfLua files`
7. `<ESC>`
8. `:quit`
9. Observe dangling `/tmp/nvim*/0` socket and directory.

Neovim version: `c5ac04331`
Nixos version: `22.05.20220127.d01b2cc (Quokka)`
Git version: `git version 2.34.1`
Fzf version: `0.29.0 (0.29.0)`

ibhagwan:
That’s because of how `tempname` works, as you can see below I’m deleting the temp file at the end but the directory is probably going to stay dangling, not sure how to address this as the directory might contain other temp files unrelated to `fzf-lua`:

https://github.com/ibhagwan/fzf-lua/blob/0f4ca97bb7063c73897b8c55204456bb88cddc3b/lua/fzf-lua/fzf.lua#L24-L37

In any event, directory pollution inside `/tmp` shouldn’t be a big deal, they don’t take space and get cleared on boot.

ibhagwan:
Found the issue, when `multiprocess=true` (the default), the command runs inside a wrapped neovim "headless" instance, when the command ends the code issues `os.exit` with the external program exit code, turns out neovim will not delete the temporary directory this way.

Solved in https://github.com/ibhagwan/fzf-lua/commit/4040e33683041c942f92a395fe5147ac704d4588

Ty for the report!

PR #1744, opened 2025-01-22
fix: clean up dangling socket and tempfile for `fzf_live`

phanen:
Use the following cmd on a large monorepo, typing quickly to ensure fzf kill process before it finished.
```vim
lua require('fzf-lua').fzf_live(function(q) return 'rg ' .. q end)
```

Then notes there're some dangling sockets and tempfiles
```sh
watch -n.1 ls -lh /run/user/1000/nvim*
watch -n.1 ls -lh /tmp/nvim.$USER/
```

## Related
* https://github.com/ibhagwan/fzf-lua/issues/329
* https://github.com/neovim/neovim/issues/30123

### DIFF
--- a/lua/fzf-lua/shell_helper.lua
+++ b/lua/fzf-lua/shell_helper.lua
@@ -73,6 +73,16 @@ local function rpc_nvim_exec_lua(opts)
     vim.fn.chanclose(chan_id)
   end)
 
+  -- Avoid dangling temp dir on premature process kills (live grep)
+  -- see more complete note in spawn.lua
+  local tmpdir = vim.fn.fnamemodify(vim.fn.tempname(), ":h")
+  if tmpdir and #tmpdir > 0 then
+    vim.fn.delete(tmpdir, "rf")
+  end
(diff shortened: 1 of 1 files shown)


ibhagwan:
Is there any downside to moving this block closer to the top of the file? I don't think we need the server anywhere in the file,  this helper calls our main neovim server but doesn't need its own named pipe.

phanen:
Tried just now, socket can be cleanup, but still creating many tempfiles.

ibhagwan:
> Tried just now, socket can be cleanup, but still creating many tempfiles.

And if the code is after the RPC call it doesn’t leave any files? Maybe the RPC call is restarting the server?

ibhagwan:
Ty @phanen, works great, still concerned whether it can still leave a dangling temp dir if killed while executing the RPC call (thus never getting to the cleanup) but I couldn’t make it happen no matter how fast I typed, so I guess this is safer in case the RPC call needs the temp dir somehow.

Issue #1346, opened 2024-07-20
Not compatible with Lua 5.1


juru1234:
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

Hi, sadly this plugin isn't compatible with neovim built with Lua 5.1 (not with LuaJIT). Eg. Ubuntu 24.04 ships such a neovim version for ppc64el, riscv64 and s390x  (https://packages.ubuntu.com/noble/neovim) because LuaJIT isn't available on all architectures. Maybe you can add to the requirements, that Lua 5.2 or LuaJIT is needed for fzf-lua? I think the main problem here are the `goto`'s in the code.

juru1234:
Lua 5.1 is also the default in neovim: https://neovim.io/doc/user/lua.html

ibhagwan:
I had no idea this was a thing, I’ll look into it.

Any clues on how to build neovim without LuaJIT on my x64 would save me some research time :)

juru1234:
Thank you. Sorry, I have no idea but I can try to look it up. I tried it on s390x Ubuntu.

ibhagwan:
Btw, how do you manage your plugins? It seems `lazy.nvim` also doesn’t support Lua51:
https://github.com/folke/lazy.nvim/blob/9a374a0fb4d3ac42dac4a129d4bead7252473c77/lua/lazy/init.lua#L56
```lua
return vim.notify("lazy.nvim requires Neovim built with LuaJIT", vim.log.levels.ERROR, { title = "lazy.nvim" })
```

ibhagwan:
Also Telescope, https://github.com/nvim-telescope/telescope.nvim?tab=readme-ov-file#getting-started:
```
Getting Started

This section should guide you to run your first builtin pickers.

Neovim v0.9.0 or the latest neovim nightly commit is required for telescope.nvim to work.
The neovim version also needs to be compiled with LuaJIT, we currently do not support Lua5.1 because of some ongoing issues.
```

juru1234:
Yes, Lazy requires LuaJIT. I use VimPlug but also Packr should work. That's why I wanted to switch from Telescope to fzf-lua :D

ibhagwan:
> Yes, Lazy requires LuaJIT. I use VimPlug but also Packr should work. That's why I wanted to switch from Telescope to fzf-lua :D

[...]

PR #1755, opened 2025-01-24
fix: clean scratch buffer introduced in #1721

phanen:
After #1721, buffer cache will not be clean after `hide`.

Old scratch buffer need a way to be delete when switching to a different picker.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -1280,6 +1280,14 @@ function M.create_user_command_callback(provider, arg, altmap)
   end
 end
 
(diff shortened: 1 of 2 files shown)


ibhagwan:
Really cool @phanen, I didn’t know 5.1/LuaJIT had garbage collection, maybe we can rid of other closure code with this approach? 

I don’t think this will pose a problem but do you think there’s a chance this could affect https://github.com/ibhagwan/fzf-lua/commit/e5c14900e8abaa962f6880853ce0f749b511aa85 / #1346?

PR #1765, opened 2025-01-26
feat: new picker for `zoxide`

ibhagwan:


doc/fzf-lua-opts.txt
```diff
@@ -1651,6 +1660,16 @@ Tmux paste buffers
 
 
                                                                               
+ZOXIDE                                                   *fzf-lua-opts-zoxide*
```
fcbruce, line 1663: @ibhagwan hey, there is a duplicated tag fzf-lua-opts-zoxide

ibhagwan: Thanks for the heads up, fixed in latest.

Issue #1771, opened 2025-01-27
Bug: error at marks


Dan7h3x:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.0-dev-1414+gb51110f4a1

### Fzf version (`fzf --version`)

0.57.0 (0476a65f)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

### Describe the bug / steps to reproduce

![Image](https://github.com/user-attachments/assets/98950c42-2701-493f-be66-967b6e89f0f5)

phanen:
I can reproduce it by:
* move to an empty line -> `m'` -> `:FzfLua marks` -> select the line of mark `'`'

Maybe related: https://github.com/neovim/neovim/issues/29807

This can fix for me: 
```lua
  api.nvim_buf_call(self.win.src_bufnr, function()
    pos = vim.api.nvim_buf_get_mark(self.win.src_bufnr, mark)
  end)
```

phanen:
Or just remove `pos[1] == tonumber(lnum)` check...

PR #1772, opened 2025-01-27
fix: nvim_buf_get_mark cannot get `'` mark correctly without curwin

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1771.


ibhagwan:
Tysm @phanen!

PR #1780, opened 2025-02-01
fix: spell_apply with no value selected

nmeum:
Without this patch, spell_apply will throw an exception if no value is selected as it attempts to concatenate a nil value with a string.

Previously: 9226c5a2291ef623ef157e0e1ffea034ffc8b8de


ibhagwan:
Ty @nmeum !

PR #1781, opened 2025-02-02
Update README.md

allworldg:
grep.hidden=false means disable hidden files?


ibhagwan:
You’re correct, missed on the refactor, Ty @allworldg !

PR #1782, opened 2025-02-02
fix: don't reference `win._previewer` in callback to make gc work

phanen:
Since de199f73a54a, garbage collection (0c384eb93f13) stops work for builtin previewer.

Fix by reference `win._previewer` outside, rather than in callback.

Minimal repro:
```lua
local ref
do
  local b = require('fzf-lua.utils').setmetatable__gc(
    { a = '4 gc done' },
    { __gc = function(self) print(self.a) end }
  )
  local tmp_func = function() print(b.a) end
  ref = tmp_func
end
-- ref = nil
collectgarbage('collect')
```

gc won't run in this case unless `ref` in explicitly set to nil somewhere, but in our case it's referenced by neovim's api internally...


ibhagwan:
Ty @phanen, I didn’t think of that when writing the callback.

PR #1786, opened 2025-02-02
Add live_grep search_paths option to doc

dennypenta:
Embedded GitHub editor adds '\n' to the last line.
I can fix fit's a critical change, I didn't fork it locally in order to make the PR

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -856,6 +856,20 @@ Search for strings/regexes using `rg`, `grep` or any other compatible grep'er bi
 
 Unless `search=...` is specified will prompt for the search string.
 
+##### grep.search_paths
+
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @dennypenta,

The txt file is auto generated from the OPTIONS.md file, can you modify the PR to add it to the markdown and it will update the txt automatically, otherwise the changes will be lost on the next docgen.

dennypenta:
done

ibhagwan:
Moved under grep (as if applies to both) and regenerates docs.

ibhagwan:
FYI, search_paths is for multiple paths, to change the root folder of the search use `FzfLua grep cwd=/path/to/project`.

PR #1787, opened 2025-02-02
fix(grep): allow matching query on line when query contains @

PeterCardenas:
## Problem

Whenever my `grep` search contained a `@`, I got spammed with a bunch of warning messages on each keystroke. Upon some investigation, I found that a `vim.regex` call made to get the highlight for the current matched line was failing since `@` is a special character for `vim.regex`.

## Solution

Escape the `@` characters with a `\` before it. This removes the warning messages and adds the highlight matching the query. I tried to limit this to the `vim.regex` call since I noticed there's some custom logic for `--fixed-strings` that is for escaping regex characters to match them explicitly.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -183,7 +183,9 @@ end
 function M.regex_to_magic(str)
   -- Convert regex to "very magic" pattern, basically a regex
   -- with special meaning for "=&<>", `:help /magic`
-  return [[\v]] .. str:gsub("[=&<>]", function(x) return [[\]] .. x end)
+  return [[\v]] .. str:gsub("[=&@<>]", function(x)
+    return "\\" .. x
+  end)
 end
 
 function M.ctag_to_magic(str)


ibhagwan:
Ty @PeterCardenas!

There’s an easier fix for this, can you add `@` to the substituted chars in utils.regex_to_magic?
https://github.com/ibhagwan/fzf-lua/blob/7b573ac57293633d9f49907624a7ed6b8333efb4/lua/fzf-lua/utils.lua#L183-L187

Line 186 should read:
```lua
return [[\v]] .. str:gsub("[=&<>]", function(x) return [[\]] .. x end)
```

If you wish I can do this myself but I don’t want to rob you of this PR :)

PeterCardenas:
@ibhagwan i can do this in this file but the vimscript looks like hieroglyphics to me (which is probably why i didn't catch that we were already doing this escaping 😅). any objections to converting it to lua?

PeterCardenas:
oh oops that's not vimscript 😅 i guess i'm just bad at reading code in general 🤣  will update the PR then

ibhagwan:
Ty @PeterCardenas 🙏😊

Issue #1789, opened 2025-02-03
Bug: race condition when fzf_bin is a shell script with a shebang


austinliuigi:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.11.0-nightly+e8a6c1b

### Fzf version (`fzf --version`)

0.57.0

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
  fzf_bin = "~/fzf_wrapper.sh"
})
```

### Describe the bug / steps to reproduce

I have a very niche usecase where I wrap my fzf command using a bash script that looks like this:
```bash
#!/usr/bin/env bash

# do stuff

fzf "$@" </proc/$$/fd/0 >/proc/$$/fd/1 2>/proc/$$/fd/2

```
 I recently installed fzf-lua to try it out,  and I set the `fzf_bin` configuration option to the path my wrapper script. I noticed that there was an issue where some providers would sometimes open up as an empty terminal buffer that just looks like it's hanging:

![Image](https://github.com/user-attachments/assets/be35453e-2417-4d03-91ff-e7d7a8a65d99)

I experimented around and observed:

1. When I set `fzf_bin = "fzf"` (no wrapper script), fzf-lua works as expected.
2. When I set `fzf_bin = "<path_to_wrapper_script>"`, fzf-lua experiences the behavior I described.
3. When I set `fzf_bin = "<path_to_wrapper_script>"` and remove the shebang from the wrapper script, fzf-lua works as expected.

I am not sure if this is an issue with fzf-lua or neovim's internal `jobstart` function or something else, but any help would be greatly appreciated. Thank you for the great plugin :)

> [!NOTE]
> Simply removing the shebang from the script isn't possible in my esoteric usecase, because I generate the script using a nixos function that automatically adds the shebang.

ibhagwan:
[...]

PR #1790, opened 2025-02-03
fix: make `opts.pipe_cmd` work with more pickers

phanen:
Related: https://github.com/ibhagwan/fzf-lua/issues/1789
* Not a really fix, but maybe `FzfLua builtin pipe_cmd=true` can workaround it now.

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -92,7 +92,7 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
       -- use input redirection with skim to prevent interface opening delay
       local bin_is_sk = opts.fzf_bin and opts.fzf_bin:match("sk$")
       local fish_shell = vim.o.shell and vim.o.shell:match("fish$")
-      if not fish_shell or bin_is_sk then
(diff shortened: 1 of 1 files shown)


ibhagwan:
> Not a really fix, but maybe FzfLua builtin pipe_cmd=true can workaround it now.

I don’t know if I want to “officially” add support for pipe_cmd, I only added it as a workaround for the (yet unexplained) hang on Windows when using FZF_DEFAULT_OPTS.

This should be merged regardless though.

ibhagwan:
@phanen, not sure I understand this PR, `FzfLua builtin pipe_cmd=true` works with HEAD, why do we need to use this?

phanen:
I originally try this to fix #1789, since in this way pipe is read by external `cat` rather than libuv internally which avoid potential blocking.

ibhagwan:
> I originally try this to fix #1789, since in this way pipe is read by external `cat` rather than libuv internally which avoid potential blocking.

I think we can skip this PR then, pipe_cmd should never be used aside from Windows testing, if something doesn’t work and we have to use it we probably have another issue that needs fixing.

Closing since #1791 supersedes this one.

PR #1791, opened 2025-02-03
fix: use async callback avoid block when open fifo

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1789.

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -104,7 +104,7 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
   table.insert(cmd, ">")
   table.insert(cmd, libuv.shellescape(outputtmpname))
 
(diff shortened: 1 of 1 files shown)


ibhagwan:
Interesting, does this fix #1789?

What about pickers using the io redirect (table content, for example `:FzfLua builtin`?
Could there other side effects to this fix?

@austinliuigi, can you test if this fixes your issue?

austinliuigi:
Just tested and it seems to be working perfectly :)

ibhagwan:
> Just tested and it seems to be working perfectly :)

Fantastic news, since this is a rather critical part of the code, gimme a few days to get home and text this and I’ll merge :)

Ty  @phanen  as always!

ibhagwan:
Ty @phanen for this fantastic fix, it’s been a while since I touched this code point.

Issue #1792, opened 2025-02-03
Feature: Delete marks in marks command similar to deleting buffers in buffer command


erik-schmidt:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

I would like to have the ability to delete a mark in the mark command similar to the `ctrl-x` keymap in buffer command, where the fzf window stays open after deleting the selected entry. I used the command in telescope to manage my marks.

I have the following action defined in my setup:
```lua
      marks = {
        actions = {
          ['ctrl-x'] = function(selected)
            local mark = selected[1]
            mark = mark:match '[^ ]+'
            vim.cmd('delm ' .. mark)
          end,
        },
      },
```

This works, but the window closes after deleting a mark. Would this be possible to have it as a dedicated action?

ibhagwan:
What about this:
```lua
      marks = {
        actions = {
          ['ctrl-x'] = {
            fn = function(selected)
              local mark = selected[1]
              mark = mark:match '[^ ]+'
              vim.cmd('delm ' .. mark)
            end,
            reload = true
          }
        },
      },
```

ibhagwan:
Complimentary of @phanen in #1793 :)

erik-schmidt:
I just tried it with the new commit and it works flawlessly. Great work, thanks!

PR #1793, opened 2025-02-03
feat: add `mark_del` action

phanen:
Close https://github.com/ibhagwan/fzf-lua/issues/1792.


ibhagwan:
Ty @phanen! I forgot ww have to convert the contents cmd to a reload style action string so `reload=true` would be seamless.

PR #1801, opened 2025-02-06
fix(typo): replace `darcula` with `dracula`

joaoinez:
Was looking through the `colorschemes.json` file and noticed these instances of `darcula`. I'm assuming this is just a typo.
So I just did a simple find and replace.


ibhagwan:
Ty @joaoinez!

PR #1822, opened 2025-02-12
New picker for vim options

thesimonho:
`FzfLua options`
A picker for viewing and working with vim options:

- Lists each option along with their current value
- Search/filter works over the option names only
- Boolean values are colored for readability (user config to turn this off)
- The current value is displayed in the previewer (in case the value is long and needs to wrap)
- Help text for the option is added to the previewer
- Pressing Enter on a boolean value will toggle the value
- Pressing Enter on a number or string value will open an input box for text entry

image


ibhagwan:
@thesimonho, tysm for this wonderful contribution, haven’t yet fully tested it but try out my changes, I modified enter to be a “reload” action and also call the populate methods within the context of the calling window, so now if you change an option it will get updated (in the source window) without losing the current cursor line.

For example, relative line, before enter:
IMG_2980

After enter:
IMG_2981


As you can see from the screens, the relative number setting is modified in the calling window, the interface reloads seamlessly, values are updated and the cursor isn’t lost.

ibhagwan:
Also modified to use `nvim_set_option_value` and added `opts.nvim_set_opts` so users can decide if they want to modify opts locally or globally (defaults to win local).

ibhagwan:
The only thing I’m conflicted about is whether we should leave the picked name as `options` or rename it to `nvim_options`, I find options a bit too vague.

thesimonho:
> @thesimonho, tysm for this wonderful contribution, haven’t yet fully tested it but try out my changes, I modified enter to be a “reload” action and also call the populate methods within the context of the calling window, so now if you change an option it will get updated (in the source window) without losing the current cursor line.

this is perfect. i was actually trying to figure out a way to do this because you're right, the refresh and losing the current line is a bit jarring

thesimonho:
> Also modified to use `nvim_set_option_value` and added `opts.nvim_set_opts` so users can decide if they want to modify opts locally or globally (defaults to win local).

should there be an action that lets the user toggle between local and global?

ibhagwan:
> > Also modified to use `nvim_set_option_value` and added `opts.nvim_set_opts` so users can decide if they want to modify opts locally or globally (defaults to win local).
> 
> should there be an action that lets the user toggle between local and global?

Or maybe enter for local and shift-enter or alt-enter for global?

thesimonho:
question: how do you feel about updating utils.input to allow a default value (the current option value). some options can be quite long and it would be quite tedious to have to type the whole thing just to make 1 small edit

thesimonho:
> > should there be an action that lets the user toggle between local and global?
> 
> Or maybe enter for local and shift-enter or alt-enter for global?

if the user sets their opts to be global scope, would enter still make a local edit?

ibhagwan:
> question: how do you feel about updating utils.input to allow a default value (the current option value). some options can be quite long and it would be quite tedious to have to type the whole thing just to make 1 small edit

Good idea, I see no issue with that, the default value is optional.

> if the user sets their opts to be global scope, would enter still make a local edit?

We remove `nvim_set_opts`, it won’t be needed anymore,  enter will (attempt) a local win option edit, if this fails the user will see an error message (e.g. try to local edit `columns`) and they can press alt-enter to perform a global edit, obviously they can press alt-enter to begin with too.

We can also add the alt-enter as a hint in the first header line (before the current header) to make that clear.

ibhagwan:
Ty @thesimonho!

I modified the keymap header to be more consistent with other pickers and use the FzfLuaHeaderBind/FzfLuaHeaderText highlight groups and also split the actions to two different ones so we have the corrent label in the F1 help.

Looks good to go :-)

PR #1829, opened 2025-02-15
fix: don't send ambiguous empty string when no selected/matched

phanen:
When trigger actions with no selected entry in hide profile, `selected` always will be `{ "" }`
To reproduce this with `mini.sh`:
* `require('fzf-lua').setup { "hide" }` (or just use `exec_silent = true` actions without this)
* `:FzfLua files cwd=/tmp` 
* typing random thing (ensure no entry selected)
* press enter, then `/tmp` opened in netrw

The root cause is that when use `{q} {+}`, `{+}` is expanded to an empty string if there's no selected entry.

We can use `{n}` to determine if it's really a no-selected case, and then send `{}` to actions rather than `{""}`.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -1236,15 +1236,21 @@ M.convert_exec_silent_actions = function(opts)
       -- Use both {q} and {+} as field indexes so we can update last query when
       -- executing the action, without this we lose the last query on "hide" as
       -- the process never terminates and `--print-query` isn't being printed
-      local field_index = v.field_index == false and "" or v.field_index or "{q} {+}"
-      if not field_index:match("^{q}") then
-        field_index = "{q} " .. field_index
+      -- When no entry selected (with {q} {+}), {+} will be forced expand to ''
+      -- Use {n} to know if we really select an empty string, or there's just no selected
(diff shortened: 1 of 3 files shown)


phanen:
@ibhagwan  Do you think should we also fix this for preview actions?
Maybe empty string should always be converted to `nil` whenever `{q}` used with `{}`.
https://github.com/ibhagwan/fzf-lua/blob/976b8f29d0222a56ba38e375ba44c78e83961be6/lua/fzf-lua/previewer/builtin.lua#L413

ibhagwan:
> @ibhagwan Do you think should we also fix this for preview actions? Maybe empty string should always be converted to `nil` whenever `{q}` used with `{}`.
> 
> https://github.com/ibhagwan/fzf-lua/blob/976b8f29d0222a56ba38e375ba44c78e83961be6/lua/fzf-lua/previewer/builtin.lua#L413
> 
> .
> Or a revisit of: [#1420 (comment)](https://github.com/ibhagwan/fzf-lua/pull/1420#issuecomment-2316658230) Maybe ensure `entry_str == nil` when no entry selected.

Good idea, ty @phanen!

I’ll test this a bit and merge.

PR #1821, opened 2025-02-12
feat: align numbers in ui-select function

Bekaboo:
Align the numbers in the ui-select function for a cleaner visual effect.

Before:

image

After:

image

Test file:

```lua
require('fzf-lua.providers.ui_select').register()
vim.ui.select(
  {
    'Acai',
    'Apple',
    'Apricot',
    'Asian Pear',
    'Avocado',
    'Banana',
    'Blackberry',
    'Blood Orange',
    'Blueberry',
    'Boysenberry',
    'Breadfruit',
    'Cactus Pear',
    'Cantaloupe',
    'Carambola',
    'Cherimoya',
    'Cherry',
    'Clementine',
    'Coconut',
    'Cranberry',
    'Custard Apple',
    'Date',
    'Dragonfruit',
    'Durian',
    'Elderberry',
    'Fig',
    'Gooseberry',
    'Grapefruit',
    'Grapes',
    'Guava',
    'Honeydew',
    'Jackfruit',
    'Jujube',
    'Kiwi',
    'Kumquat',
    'Lemon',
    'Lime',
    'Lingonberry',
    'Loquat',
    'Longan',
    'Lychee',
    'Mandarin',
    'Mango',
    'Mangosteen',
    'Marionberry',
    'Mulberry',
    'Nectarine',
    'Olive',
    'Orange',
    'Papaya',
    'Passion Fruit',
    'Peach',
    'Pear',
    'Persimmon',
    'Pineapple',
    'Plantain',
    'Plum',
    'Pomegranate',
    'Pomelo',
    'Prickly Pear',
    'Prune',
    'Quince',
    'Raisin',
    'Rambutan',
    'Raspberry',
    'Red Banana',
    'Red Currant',
    'Sapodilla',
    'Sapote',
    'Soursop',
    'Star Apple',
    'Star Fruit',
    'Strawberry',
    'Sugar Apple',
    'Tamarind',
    'Tangerine',
    'Tomato',
    'Ugli Fruit',
    'Watermelon',
    'White Currant',
    'White Sapote',
    'Yellow Passion Fruit',
    'Yuzu',
    'Acerola',
    'Ackee',
    'African Cherry Orange',
    'Ambarella',
    'Araza',
    'Babaco',
    'Bael',
    'Barbados Cherry',
    'Barberry',
    'Bearberry',
    'Bilberry',
    'Black Sapote',
    'Black Mulberry',
    'Brazilian Guava',
    "Buddha's Hand",
    'Calamansi',
[...]

PR #1837, opened 2025-02-18
fix(codeaction): consider whitespace when getting index

MariaSolOs:
With `jdtls` there's a space before the code action index, causing the `assert` in the following line to fail.


ibhagwan:
This is due to #1821, I thought we got them all, Ty @MariaSolOs!

MariaSolOs:
> This is due to #1821, I thought we got them all, Ty @MariaSolOs!

Always a pleasure :)

PR #1855, opened 2025-02-25
fix(complete): fallback to user provided `opts.cwd`

phanen:
Problem: `require("fzf-lua").complete_file { cwd = vim.fs.root(0, ".git") }` don't work.

Solution: Allow fallback to user passed `opts.cwd`, (higher priority than `uv.cwd()`)


ibhagwan:
Ty @phanen!

PR #1857, opened 2025-02-27
fix(lsp): correct jump1 logic for call hierarchy

andrewbraxton:
Current logic will actually jump to every location even if there is more than one result.


ibhagwan:
Ty @andrewbraxton!

PR #1862, opened 2025-03-04
fix(history): Restore the default header

andrewbraxton:
2ab648faf fixed custom headers but broke default headers for
command|search history.


ibhagwan:
Ty @andrewbraxton!

PR #1864, opened 2025-03-04
fix: `Unable to stat file` after cwd change

phanen:
https://github.com/ibhagwan/fzf-lua/commit/b39f725cc59fbd83bfaca73ce48f4cdc50388e2d fix for action, but not for previewer.


scripts/make_cli.lua
```diff
@@ -16,7 +16,7 @@ if glob then
 else
   -- All test files
   find_files = function()
-    return vim.fn.globpath("tests", "**/*_spec.lua", true, true)
+    return vim.fn.globpath("tests", "**/ui_spec.lua", true, true)
```
ibhagwan: I think this change is by mistake?

ibhagwan:
I think you changed make_cli.lua glob by mistake? You know you can run `make test glob=ui` to run just the ui spec

ibhagwan:
Isn’t it better to leave the glob as it was matching only xxx_spec.lua files rather than all?

phanen:
I don't know why `make test glob=ui` don't do test for me locally.
So what's wrong again now...

ibhagwan:
> I don't know why `make test glob=ui` don't do test for me locally. So what's wrong again now... Ok I miss `cwd_prompt=false`.

I was wrong, the glob matches non-recursively, run `make test glob=f`.

ibhagwan:
Tysm @phanen!

PR #1865, opened 2025-03-04
fix: `ctrl-g` should not change layout when use `live_grep` with `winopts.split`

phanen:
`ctrl-g` in `live_grep` picker should not change layout:
```lua
require('fzf-lua').setup { winopts = { split = 'botright new' } }
```

Did a bisect, this happened after https://github.com/ibhagwan/fzf-lua/commit/d932025a62bc6644cb531e74cf2f1ef35606fd2b..


ibhagwan:
Ty @phanen !

PR #1869, opened 2025-03-09
fix(files): use POSIX compliant options for find_opts

folliehiyuki:
See `find(1)` man page for more details. The `!` character is escaped here to avoid it being interpreted weirdly by the shell (.e.g in bash's conditional clauses). I also remove the escaped dot character in -path argument. `find` doesn't treat `/` and `.` differently.


ibhagwan:
Ty @folliehiyuki !

PR #1871, opened 2025-03-10
fix: reload before unhide buffers/tabs pickers

phanen:
After ~https://github.com/ibhagwan/fzf-lua/commit/da9a0898541b34b95361e4223df17a7f680be070~ https://github.com/ibhagwan/fzf-lua/commit/480e29c20cb324bb9bf3d6f7d8e5505bcb49d555
Actions like `ctrl-x` may throw an error since CTX updated.

This PR
* reload all CTX include `buflist` for buffers
* force a reload by a dummy reload bind when unhide

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -299,7 +299,7 @@ M.CTX = function(opts)
   -- perhaps a min impact optimization but since only
   -- buffers/tabs use these we only include the current
   -- list of buffers when requested
-  if opts.includeBuflist and not M.__CTX.buflist then
+  if opts.force or opts.includeBuflist and not M.__CTX.buflist then
     -- also add a map for faster lookups than `utils.tbl_contains`
     -- TODO: is it really faster since we must use string keys?
     M.__CTX.bufmap = {}
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -601,29 +601,35 @@ M.defaults.buffers              = {
   _actions              = function()
     return M.globals.actions.buffers or M.globals.actions.files
   end,
-  actions               = { ["ctrl-x"] = { fn = actions.buf_del, reload = true } },
+  actions               = {
+    ["ctrl-x"] = { fn = actions.buf_del, reload = true },
+    ["f11"] = { fn = actions.dummy_abort, reload = true },
+  },
+  _reload_bind          = "<F11>",
   _cached_hls           = { "buf_nr", "buf_flag_cur", "buf_flag_alt", "path_linenr" },
 }
 
 M.defaults.tabs                 = {
-  previewer   = M._default_previewer_fn,
-  tab_title   = "Tab",
-  tab_marker  = "<<",
-  file_icons  = 1,
-  color_icons = true,
-  _actions    = function()
+  previewer    = M._default_previewer_fn,
+  tab_title    = "Tab",
+  tab_marker   = "<<",
+  file_icons   = 1,
+  color_icons  = true,
+  _actions     = function()
     return M.globals.actions.buffers or M.globals.actions.files
   end,
-  actions     = {
+  actions      = {
     ["enter"]  = actions.buf_switch,
     ["ctrl-x"] = { fn = actions.buf_del, reload = true },
+    ["f11"]    = { fn = actions.dummy_abort, reload = true },
   },
-  fzf_opts    = {
+  _reload_bind = "<F11>",
+  fzf_opts     = {
     ["--multi"]     = true,
     ["--delimiter"] = "[\\):]",
     ["--with-nth"]  = "5..",
   },
-  _cached_hls = { "buf_nr", "buf_flag_cur", "buf_flag_alt", "tab_title", "tab_marker", "path_linenr" },
+  _cached_hls  = { "buf_nr", "buf_flag_cur", "buf_flag_alt", "tab_title", "tab_marker", "path_linenr" },
 }
 
 M.defaults.lines                = {
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -797,8 +797,8 @@ function M.fzf_winobj()
   return loadstring("return require'fzf-lua'.win.__SELF()")()
 end
 
-function M.CTX()
-  return loadstring("return require'fzf-lua'.core.CTX()")()
+function M.CTX(...)
+  return loadstring("return require'fzf-lua'.core.CTX(...)")(...)
 end
 
 function M.__CTX()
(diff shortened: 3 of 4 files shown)


ibhagwan:
> After https://github.com/ibhagwan/fzf-lua/commit/da9a0898541b34b95361e4223df17a7f680be070
Actions like ctrl-x may throw an error since CTX updated.

Interesting, I don’t see the connection to the changes in the commit to reload actions.

Tysm @phanen for yet another PR :)

Can’t say I’m super excited about this as it feels a bit “hacky” having a dummy reload bind, I’m trying to think is there another approach to this with an fzf event? Maybe a `resize` bind triggers on unhide? Maybe with a `opts._run_once` which gets cleared after a CTX reload?

ibhagwan:
Btw, there is a better solution but it requires a big refactor and also to drop support for skim, I’ve been thinking for a while of using a `fzf --listen`, have a few processes of fzf running and communicate with fzf using libcurl bindings or luasocket with http requests, this gives us most flexibility and also eliminates the process spawn time on windows (slow).

ibhagwan:
@phanen, checkout https://github.com/ibhagwan/fzf-lua/compare/main...pr_1871, I still wasn't able to make it work but I feel this is the right direction, I'm trying to hijack the resize event and then using a variable as I mentioned at the previous comments to perform a "oneshot" reload.

The resize however doesn't trigger on unhide so I've been trying to trigger a resize with SIGWINCH (signal 28) but it seems to do nothing when sent to fzf (even with `kill -28 <fzf pid?`), perhaps there's another signal that can work with fzf but I didn't have time as I had to leave.

I will attempt to solve this later.

ibhagwan:
> Yes, with this would be easier as just curl a reload action and possible to avoid killing a fzf process, and also elimate tons of hack in current fzf-lua (e.g. on windows there's no SIGWINCH 28? so fzf's `resize' don't work on it

Maybe it’s not that critical but I hate to rely on additional binaries being is installed, that’s why I prefer a native lua HTTP post library but luasocket is yet another requirement so I’m also not loving it.

Windows not having sig 28 I can live with, windows already has loads of limitations and doesn’t really work well with hide, for example `git_status` plus hide on windows doesn’t work due to “command too long` (too many reload + execute silent actions).

I’m still not sure why sit 28 doesn’t work with fzf on my Linux, to be explored.

> I once wrote a demo https://github.com/phanen/fzf.lua

Interesting we had the same thought :)

> maybe fzf-lua first need some layer like this to replace https://github.com/vijaymarupudi/nvim-fzf.git.

My idea was an Fzf class/object, each new instance with their own listen address (+ auth token so “safe”) so we’re not limited to one fzf process, that the user can instantiate and attach to any window, that way you could run say `:FzfLua live_grep instance=1` and `FzfLua files instance=2` and then you have a persistent live grep/files, when you run `:FzfLua resume instance=2` you’re back on the files instance and can guarantee it won’t get overwritten by the next `:FzfLua …`, we can also have “sticky” fzf-lua windows this way.

A few concerns with the new approach:
1. Increased requirements (curl/lib/etc, unless we can find native simple lua HTTP post lib)
2. We have to drop skim
3. I’m unsure if the HTTP API is sufficient to do everything fzf-lua needs yet (junegunn keeps adding new methods so that’s a plus)
4. FzfLua has been very stable for a while now, this will be a lot of work/testing all while being in the “nice to have” category (not really user driven because we’re “missing” something)

On the plus side this will simplify the code significantly and we can create a proper API and code structure from scratch as opposed to the sometimes not so organized building blocks we currently have due to fzf-lua being developed for a long time incrementally, definitely doesn’t help that I’m also busy in life and other interesting projects so have to find the right motivation for such a big rewrite.

ibhagwan:
> I found there're still many limitations even if use HTTP API, since many flags are one-way-door (behavior like --delimiter/--header-lines cannot be changed later by curl a fzf action). So to ensure a picker work as expected we cannot choose instance=n randomly...

I feared the HTTP API isn’t there yet to accommodate all the features we need.

> the other benefit I can think of is that we can handle all keymap in neovim (rather than by --bind). But that may not worth a major refactoring currently.

Combining actions/keymap.fzf/keymap.builtin into one table with neovim style binds is also something that was on my mind for a while, it’s even possible with the current architecture just a tiny bit more complicated to design, but then again there needs to be proper motivation as the onky benefit for this is nicer/simpler configuration.

phanen:
@ibhagwan I checkout your SIGWINCH commit again know why it didn't work: 

it send signal to `/usr/bin/sh` rather than `fzf`.

ibhagwan:
> @ibhagwan I checkout your SIGWINCH commit again know why it didn't work:
it send signal to /usr/bin/sh rather than its fzf child.
> Patch on your commit that work:

Nice @phanen! I still haven’t gotten to it as I’m busy with a new open source project I will soon(ish) publish :)

Do you find the SIGWINCH approach cleaner?

phanen:
At least better than feeding termcodes...?

ibhagwan:
> At least better than feeding termcodes...?

I agree.

phanen:
Related: https://github.com/ibhagwan/fzf-lua/discussions/1901#discussioncomment-12545120.

ibhagwan:
@phanen, will you look at my latest PR branch commit https://github.com/ibhagwan/fzf-lua/compare/main...pr_1871 and let me know what you think?

Instead of [unconditionally calling reload on every resize (fake or not)](https://github.com/ibhagwan/fzf-lua/pull/1871#issuecomment-2731182052) I modified the resize event to a transform that only calls the reload once upon our "fake" resize and only does so with pickers that have `opts.__reload_cmd` (buffers, tabs, others).

The only thing I'm not sure is if I want to add a var to enable the "hide-reload" only for specific pickers or do so for all pickers that have `opts.__reload_cmd`?

phanen:
> The only thing I'm not sure is if I want to add a var to enable the "hide-reload" only for specific pickers or do so for all pickers that have `opts.__reload_cmd`?

IMO that's enough since those buffer is not performance critical? I'm not those using thousand of buffers but I test `Buffers`/`Tabs` by `argadd **/*.lua` in neovim repo it work instantly when unhide.

But there seems a bug (racing?) when I hit `unhide` shortcut then really quickly hit `resume` shortcut:
```
Error executing vim.schedule lua callback: /home/phan/b/fzf-lua/lua/fzf-lua/shell.lua|313| /home/phan/b/neovim/runtime/lua/vim/shared.lua:273: t: expected table, got nil
|| stack traceback:
|| [builtin#36]: at 0x560d3fc38020
lua/fzf-lua/shell.lua|313| in function 'fn'
lua/fzf-lua/shell.lua|77| in function </home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:76>
```

ibhagwan:
Ty @phanen for the feedback, https://github.com/ibhagwan/fzf-lua/commit/aeff8132009a7fc55c6c43bca4288d5ba26a5393  merged into main.

> But there seems a bug (racing?) after I open :Buffers then hit hide shortcut then very quickly hit unhide shortcut:

If this becomes a problem we’ll deal with it then.

PR #1885, opened 2025-03-14
feat(defaults): make delta switch dark/light

phanen:
`delta` have [osc 10/11 support] support, so it can detect light/dark in
terminal like [kitty,
neovim](https://github.com/bash/terminal-colorsaurus/blob/main/doc/terminal-survey.md),
but sadly not in fzf previewer.

But we can pass a flag to delta using neovim's `&bg` (neovim caculate a
luminance to know if the terminal is dark/light
https://github.com/neovim/neovim/blob/0d5a93be22d27dd095ecbc02775de1f3f4b8d3c9/runtime/lua/vim/_defaults.lua#L776).

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -21,7 +21,8 @@ function M._default_previewer_fn()
 end
 
 function M._preview_pager_fn()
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @phanen!

PR #1888, opened 2025-03-15
fix: `{q}` may be expanded to nothing rather than `''` on windows

phanen:
Fix https://github.com/ibhagwan/fzf-lua/discussions/1887


ibhagwan:
Ty @phanen!

PR #1898, opened 2025-03-18
feat: more table options can be function

phanen:
Fix https://github.com/ibhagwan/fzf-lua/discussions/1870#discussioncomment-12535353.
```lua
require('fzf-lua').setup {
  tabs = {
    keymap = function()
      local curtab = fn.tabpagenr()
      local pos = 1
      for i, t in ipairs(api.nvim_list_tabpages()) do
        if i < curtab then pos = pos + 1 + #api.nvim_tabpage_list_wins(t) end
      end
      return { fzf = { load = string.format("transform:echo 'pos(%d)'", pos) } }
    end,
  },
}
```

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -127,6 +127,11 @@ do
   m.globals = M.globals
 end
 
+local eval = function(v, ...)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @phanen !

Issue #1907, opened 2025-03-22
Feature: Handle 'winborder' option from nvim nightly


andrewbraxton:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

Neovim nightly has a 'winborder' option but setting it to something like 'rounded' is not handled well by fzf-lua. A border will be unexpectedly added to the entire Neovim window and the previewer scrollbar. I suspect that `border = 'none'` needs to be explicitly added to the window config for any calls to api.nvim_open_win()

andrewbraxton:
<img width="810" alt="Image" src="https://github.com/user-attachments/assets/7d71139e-1cbf-4e59-bef2-2dfc25c1e9a8" />

This commit fixes the scrollbar, but there's still an extra border placed around the entire window.

phanen:
That seems a transparent window used as backdrop (I found I've disabled it in my config `winopts.backdrop=100`...

ibhagwan:
Ty @andrewbraxton, fixed in https://github.com/ibhagwan/fzf-lua/pull/1910 thanks to @phanen.

andrewbraxton:
Thanks @phanen @ibhagwan !

PR #1908, opened 2025-03-22
fix: force `border = "none"` for scrollbar to ignore `&winborder`

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1907.


ibhagwan:
Ty @phanen!

Issue #1904, opened 2025-03-21
Bug: An action is never called again when it `return` on no-match, under hide profile `reload=true`


nyngwang:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

mac

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.0-dev-2061+gfa85543e3b

### Fzf version (`fzf --version`)

0.60.3 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [x] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup {
        'hide',
        tabs = {
          fzf_opts = { ['--with-nth'] = '3,5..' },
          formatter = 'path.filename_first',
          show_unlisted = true,
          tab_marker = { ' <<< CURRENT', 'LspInlayHint' },
          actions = {
            ['ctrl-r'] = {
              fn = function (selected, opts)
                -- print(vim.inspect(opts))
                print('debug ctrl-r, vim.inspect(selected):', vim.inspect(selected))
                print('debug ctrl-r, opts._last_query:', opts._last_query)
                local query, pos, entry = unpack(selected, 1, 3)
                if #entry == 0 then
                  return
                end
                print(query, entry)
              end,
              reload = true,
              field_index = '{q} $FZF_POS {}',
            },
          }
        },
      }
```

### Describe the bug / steps to reproduce

In my custom action, I want to detect the case where an action is called no no-match, i.e. there is no entry in the list, and skip the latter part of the function. Currently, after the following `return` is executed, the action will never get called again unless the fzf-lua window is re-open:

```lua
if #entry == 0 then
  return
end
```

1. `:e 123.lua`
2. `:FzfLua tabs`
3. type `123`, press `ctrl-r`
4. type `4` to update query to `1234`, press `ctrl-r`, should notice that the :message prompt stuck.
5. delete `4` to update query to `123`, press `ctrl-r`, this action is not triggered, proven in step 6.
6. `:message` doesn't print log for step 5.

## Demo

[...]

PR #1909, opened 2025-03-22
fix: don't use unused `{+}` to avoid `{+}` is not expand when there's no match

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1904.


ibhagwan:
Ty @phanen! Should’ve been written like this to begin with, once I started converting lua table/funcfion pickers to reload I copied the same code to all, the `{+}` is indeed never used in the reload.

PR #1913, opened 2025-03-22
feat: support snacks.image

phanen:
To make local image preview need something upstream: https://github.com/folke/snacks.nvim/pull/1618.

But I think it have been work for url image/latex.

And note your must set `winblend=0` for kitty backend, https://github.com/folke/snacks.nvim/pull/1615.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -229,6 +229,7 @@ end
 
 function Previewer.base:get_tmp_buffer()
   local tmp_buf = api.nvim_create_buf(false, true)
+  vim.bo[tmp_buf].modeline = true
   vim.bo[tmp_buf].modifiable = true
   vim.bo[tmp_buf].bufhidden = "wipe"
   return tmp_buf
@@ -895,6 +896,7 @@ function Previewer.base:update_ts_context()
   -- https://github.com/neovim/neovim/commit/45e606b1fddbfeee8fe28385b5371ca6f2fba71b
   -- For more info see #1922
   local lang = vim.treesitter.language.get_lang(ft)
+  if not utils.has_ts_parser(lang) then return end
   local parser = vim.treesitter.get_parser(self.preview_bufnr, lang)
   local context_updated
   for _, t in ipairs({ 0, 20, 50, 100 }) do
@@ -967,87 +969,93 @@ function Previewer.buffer_or_file:do_syntax(entry)
   if not entry or not entry.path then return end
   local bufnr = self.preview_bufnr
   local preview_winid = self.win.preview_winid
-  -- can be used by snacks.image to get abspath of e.g. [file.png] https://github.com/folke/snacks.nvim/pull/1618
-  vim.b[bufnr].bufpath = entry.path
-  if api.nvim_buf_is_valid(bufnr) and vim.bo[bufnr].filetype == "" then
-    if fn.bufwinid(bufnr) == preview_winid then
-      -- do not enable for large files, treesitter still has perf issues:
-      -- https://github.com/nvim-treesitter/nvim-treesitter/issues/556
-      -- https://github.com/nvim-treesitter/nvim-treesitter/issues/898
-      local lcount = api.nvim_buf_line_count(bufnr)
-      local bytes = api.nvim_buf_get_offset(bufnr, lcount)
-      local syntax_limit_reached = 0
-      if self.syntax_limit_l > 0 and lcount > self.syntax_limit_l then
-        syntax_limit_reached = 1
-      end
-      if self.syntax_limit_b > 0 and bytes > self.syntax_limit_b then
-        syntax_limit_reached = 2
-      end
-      if syntax_limit_reached > 0 and self.opts.silent == false then
-        utils.info(string.format(
-          "syntax disabled for '%s' (%s), consider increasing '%s(%d)'", entry.path,
-          syntax_limit_reached == 1 and ("%d lines"):format(lcount) or ("%db"):format(bytes),
-          syntax_limit_reached == 1 and "syntax_limit_l" or "syntax_limit_b",
-          syntax_limit_reached == 1 and self.syntax_limit_l or self.syntax_limit_b
-        ))
-      end
-      if syntax_limit_reached == 0 then
-        local fallback = (function()
-          local ft = entry.filetype or vim.filetype.match({ buf = bufnr, filename = entry.path })
-          if type(ft) ~= "string" then
-            return true
-          end
-          -- Use buf local var as setting ft might have unintended consequences
-          -- currently only being used in `update_render_markdown` but might be
-          -- of use in the future?
-          vim.b[bufnr]._ft = ft
-          local ts_enabled = (function()
-            if not self.treesitter or
-                self.treesitter.enabled == false or
-                self.treesitter.disabled == true or
-                (type(self.treesitter.enabled) == "table" and
-                  not utils.tbl_contains(self.treesitter.enabled, ft)) or
-                (type(self.treesitter.disabled) == "table" and
-                  utils.tbl_contains(self.treesitter.disabled, ft)) then
-              return false
-            end
-            return true
-          end)()
-          local ts_success = ts_enabled and ts_attach(bufnr, ft)
-          if not ts_success then
-            pcall(function() vim.bo[bufnr].syntax = ft end)
-          else
-            self:update_render_markdown()
-            self:update_ts_context()
-          end
-        end)()
-        if fallback then
-          if entry.filetype == "help" then
-            -- if entry.filetype and #entry.filetype>0 then
-            -- filetype was saved from a loaded buffer
-            -- this helps avoid losing highlights for help buffers
-            -- which are '.txt' files with 'ft=help'
-            pcall(function() vim.bo[bufnr].filetype = "help" end)
-          else
-            -- prepend the buffer number to the path and
-            -- set as buffer name, this makes sure 'filetype detect'
-            -- gets the right filetype which enables the syntax
-            local tempname = path.join({ tostring(bufnr), entry.path })
-            pcall(api.nvim_buf_set_name, bufnr, tempname)
-          end
-          -- nvim_buf_call has less side-effects than window switch
-          local ok, _ = pcall(api.nvim_buf_call, bufnr, function()
-            vim.cmd("filetype detect")
-          end)
-          if not ok then
-            utils.warn(("syntax highlighting failed for filetype '%s', ")
-              :format(entry.path and path.extension(entry.path) or "<null>") ..
-              "open the file and run ':filetype detect' for more info.")
-          end
-        end
-      end
-    end
+  if not api.nvim_buf_is_valid(bufnr)
+      or vim.bo[bufnr].filetype ~= ""
+      or fn.bufwinid(bufnr) ~= preview_winid
+  then
+    return
   end
+
+  -- assign a name for noname scratch buffer
+  -- can be used by snacks.image to abspath e.g. [file.png]
+  -- https://github.com/folke/snacks.nvim/pull/1618
+  vim.b[bufnr].bufpath = entry.path
+
+  -- do not enable for large files, treesitter still has perf issues:
+  -- https://github.com/nvim-treesitter/nvim-treesitter/issues/556
+  -- https://github.com/nvim-treesitter/nvim-treesitter/issues/898
+  local lcount = api.nvim_buf_line_count(bufnr)
+  local bytes = api.nvim_buf_get_offset(bufnr, lcount)
+  local syntax_limit_reached = 0
+  if self.syntax_limit_l > 0 and lcount > self.syntax_limit_l then
+    syntax_limit_reached = 1
+  end
+  if self.syntax_limit_b > 0 and bytes > self.syntax_limit_b then
+    syntax_limit_reached = 2
+  end
+  if syntax_limit_reached > 0 and self.opts.silent == false then
+    utils.info(string.format(
+      "syntax disabled for '%s' (%s), consider increasing '%s(%d)'", entry.path,
+      syntax_limit_reached == 1 and ("%d lines"):format(lcount) or ("%db"):format(bytes),
+      syntax_limit_reached == 1 and "syntax_limit_l" or "syntax_limit_b",
+      syntax_limit_reached == 1 and self.syntax_limit_l or self.syntax_limit_b
+    ))
+  end
+
+  if syntax_limit_reached ~= 0 then
+    return
+  end
+
+  -- filetype detect
+  ---@type string
+  local ft = (function()
+    local ft = entry.filetype or vim.filetype.match({ buf = bufnr, filename = entry.path })
+    if type(ft) == "string" then
+      return ft
+    end
+    -- prepend the buffer number to the path and
+    -- set as buffer name, this makes sure 'filetype detect'
+    -- gets the right filetype which enables the syntax
+    local tempname = path.join({ tostring(bufnr), entry.path })
+    pcall(api.nvim_buf_set_name, bufnr, tempname)
+    -- nvim_buf_call has less side-effects than window switch
+    -- doautocmd filetypedetect BufRead (vim.filetype.match + ftdetect) + do_modeline
+    local ok, _ = pcall(api.nvim_buf_call, bufnr, function()
+      utils.eventignore(function() vim.cmd("filetype detect") end, preview_winid, "FileType")
+    end)
+    if not ok then
+      utils.warn(("':filetype detect' failed for '%s'"):format(entry.path or "<null>"))
+    end
+    return vim.bo[bufnr].filetype
+  end)()
+
+  if ft == "" then return end
+
+  -- Use buf local var as setting ft might have unintended consequences
+  -- used in `update_render_markdown`, `attach_snacks_image`
+  vim.b[bufnr]._ft = ft
+
+  local ts_enabled = (function()
+    if not self.treesitter or
+        self.treesitter.enabled == false or
+        self.treesitter.disabled == true or
+        (type(self.treesitter.enabled) == "table" and
+          not utils.tbl_contains(self.treesitter.enabled, ft)) or
+        (type(self.treesitter.disabled) == "table" and
+          utils.tbl_contains(self.treesitter.disabled, ft)) then
+      return false
+    end
+    return true
+  end)()
+
+  local ts_success = ts_enabled and ts_attach(bufnr, ft)
+  if not ts_success then
+    pcall(function() vim.bo[bufnr].syntax = ft end)
+    return
+  end
+
+  self:update_render_markdown()
+  self:update_ts_context()
 end
 
 function Previewer.base:maybe_set_cursorline(win, pos)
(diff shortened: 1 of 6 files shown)


ibhagwan:
That was fast @phanen :)

And looks pretty cllean and easy.

ibhagwan:
> To make local image preview need something upstream: https://github.com/folke/snacks.nvim/pull/1618.

Haven’t explored snacks.image code yet so maybe I’m totally off here but maybe we can use the same code used by snacks.picker previewer to display the image?

phanen:
Here is all Snacks.picker code about image preview (so it support file preview only)
https://github.com/folke/snacks.nvim/blob/f0d34336dbac2909654ca05aabc472edc73c7c8a/lua/snacks/picker/preview.lua#L40-L45
https://github.com/folke/snacks.nvim/blob/f0d34336dbac2909654ca05aabc472edc73c7c8a/lua/snacks/picker/preview.lua#L102-L104

I did a bit more to support preview image inline in markdown/tex.

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -902,6 +910,28 @@ function Previewer.base:update_render_markdown()
...
+  local bufnr, preview_winid = self.preview_bufnr, self.win.preview_winid
+  if not simg
+      or not (simg.config.doc.inline and simg.terminal.env().placeholders)
+      or vim.b[bufnr].snacks_image_attached then
+    return
+  end
+
+  -- restore default winblend when on unsupport ft
+  local ft = vim.b[bufnr]._ft
+  _G._fzf_lua_snacks_langs = _G._fzf_lua_snacks_langs or simg.langs()
```
phanen, line 958: Here `simg.langs` will use `nvim_get_runtime_file` with wildcards to scan for query file (which neovim internally spawn a `vim.o.shell`...

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -902,6 +910,28 @@ function Previewer.base:update_render_markdown()
...
+  if not vim.tbl_contains(_G._fzf_lua_snacks_langs, vim.treesitter.language.get_lang(ft)) then
+    vim.wo[preview_winid].winblend = self.winblend
+    return
+  end
+
+  vim.b[bufnr].snacks_image_src = entry.path
+  vim.wo[preview_winid].winblend = 0 -- https://github.com/folke/snacks.nvim/pull/1615
```
phanen, line 964: I didn't want my preview always has a `winblend=0` only to support a inline preview, so I temporaly toggle it.

nenahp:
Maybe give a option to disable it. But snacks.image seems only render the part shown on screen?
https://github.com/folke/snacks.nvim/blob/e55ae37bebd53ab0e24a47d88ef50267207ffd91/lua/snacks/image/doc.lua#L207

I try with a huge markdown file with many inline thing: https://github.com/folke/snacks.nvim/blob/ecd77df17d74cf0a3dbb1488c61d64e6435e4dca/tests/image/big.md.

With or without this pr, if previewer enabled, the are the same slow. (treesitter should be the main cost, even async parsing in nightly don't help much).

lua/fzf-lua/previewer/builtin.lua
ibhagwan: Why not set your own config defaults `winopts.preview.winopts.winblend=0`?
https://github.com/ibhagwan/fzf-lua/blob/ac6a34ea39831ec71c14f72075facf377ea9a00d/lua/fzf-lua/defaults.lua#L75-L87

lua/fzf-lua/previewer/builtin.lua
ibhagwan: I see

lua/fzf-lua/previewer/builtin.lua
phanen: I want `winblend=0` only for buffer attached with snacks.image (that's also necessary to make preview image work..). But not for all filetype e.g. when preview `.lua`/`.c` file.

Or if I miss something, can user already config a winopts for each preview buffer?

ibhagwan: > Or if I miss something, can user already config a winopts for each preview buffer?

No no you got it right :-)

ibhagwan:
@phanen 

- Is this still dependent on https://github.com/folke/snacks.nvim/pull/1618 or can it be merged prior?
- With this change I feel it would also be nice to have our media extensions be auto detected based on binaries installed,  snacks (if kitty protocol supported) -> ueberzugpp -> chafa -> viu, what do you think?

phanen:
> Is this still dependent on

It can be merged. Only on some edge case:
e.g. `test.png` cannot be shown in noname preview buffer
https://github.com/folke/snacks.nvim/blob/95878ad32aaf310f465a004ef12e9edddf939287/tests/image/test.jsx#L6

ibhagwan:
> e.g. test.png cannot be shown in noname preview buffer

When does this happen in fzf-lua? Albeit a fake path our preview buffer should always have a name no?

ibhagwan:
> No preview buffer have no name, for all fzf-lua/snacks.picker's preview buffer `nvim_get_buf_name` always return `""`.
> 
> EDIT: if you try to set to the same name by `nvim_buf_set_name`:
> 
> ```
> E5108: Error executing lua [string ":lua"]:1: Vim:E95: Buffer with this name already exists
> ```


That’s why we use a “fake name” with the buffer number prefixed, this way the filetype detect works perfectly, if it returns an empty string on a snacks image buffer maybe we need to move the call to attach after these lines or use the same hack (will snacks find the image as the fake path is non existent…)?

https://github.com/ibhagwan/fzf-lua/blob/a09296cf7afaf22842b3bff7e5bfdec943a3c29c/lua/fzf-lua/previewer/builtin.lua#L996-L1000

phanen:
I'm a bit confusing about the difference between `vim.filetype.match` and `filetype detect`...

phanen:
What `filetype detect` does:
* first it `runtime! filetype.{lua,vim}`: https://github.com/neovim/neovim/blob/fd973c0a4ec0246708d0fcc46e66e38dd1f89a26/src/nvim/ex_docmd.c#L7635-L7637.

ibhagwan:
> ... wait, I insert a debugprint here and nothing printed, so this logic is never reached...?
> vim.filetype.match recognize it's markdown/jsx, so no fallback,

you’re correct, this is old fallback which apparently won’t be reached much since the addition of vim.filetype.match.


> I'm a bit confusing about the difference between vim.filetype.match and filetype detect...

vim.filetype.match didn’t exist when fzf-lua was first released with 0.5, filetype detect only works for current buffer and existed before 0.5

ibhagwan:
What if use the old bufnr prefix bufname hack just for snacks until your PR is merged?

ibhagwan:
> Emm.. they are essentially a bit different...

We should probably remove the filetype detect fallback, I don’t see it ever useful, if vim.filetype.match fails we have other issues.

phanen:
modeline is useful when `filetype.lua don't have some werid custom ft pattern.

ibhagwan:
> filetype detect do one thing more is do_modeline

I can live without modeline filetype custom ft.

ibhagwan:
> modeline is useful when `filetype.lua don't have some werid custom ft pattern. Maybe implement a do modeline function if needed...

We can leave the fallback for modeline, either way is fine.

ibhagwan:
> OFF-TOPIC to this: set a `b:bufpath` for noname buffer is also usable in `snacks.bigfile`. `snacks.bigfile` teach `vim.filetype.match` to match bigfile as `ft=bigfile`, then `snacks.bigfile` use `autocmd FileType` to do some "setup" (e.g. `set syntax=real_ft` if we want treesitter fallback).
> 
> But after considering for a while, I think both fzf-lua/snacks.picker don't really need this ([folke/snacks.nvim@0ed941f](https://github.com/folke/snacks.nvim/commit/0ed941f4881d29b20389a567bb8b544b49ee37bb)):
> 
> * since they all eventignore "FileType" when load new buffer.. so `snacks.bigfiles` can only detect, no syntax highlighting setup triggered
> * fzf-lua detect "bigflie_limit"/"syntax_limit" by its own..
> * snacks.picker also implement separate logic for bigfile detection (https://github.com/folke/snacks.nvim/blob/e2cb9df7d0695911f60a4510191aaf4a3d0d81ad/lua/snacks/picker/core/preview.lua#L359-L369)...

Interesting, thanks for the investigation :)


I see you just added the fake path https://github.com/ibhagwan/fzf-lua/pull/1913/commits/8ef98af81bba5f5e99c45a4945cdcc06d1c2bb57.

phanen:
> What if use the old bufnr prefix bufname hack just for snacks until your PR is merged?

Sadly snacks cannot resolve fake path.

ibhagwan:
> > What if use the old bufnr prefix bufname hack just for snacks until your PR is merged?
> 
> Sadly snacks cannot resolve fake path.

We’ll have to wait for your PR to be merged then.

phanen:
CI is stucked... And I hope macos test is irrelated. Should I force push it again??

phanen:
Hmm... another thing I learn from `snacks.bigfile`: a bigfile don't need to be big

Try this, then `FzfLua files`, type a l p h i n e , it a bit stuck
``` 
curl -ssL https://raw.githubusercontent.com/fish-shell/fish-shell/ad90ae292d72722ba111bac90bd4d2c62bfbbc05/share/tools/web_config/js/alpine.js > alphine.js
```

This should be improved..

ibhagwan:
> ~CI is stucked...~ DONE (4 minute ubuntu). And I hope macos test is irrelated. Should I force push it again??

I re-ran the failed job and it succeeded, Mac sometimes does that, probably needs a similar fix to #1914.

> Hmm... another thing I learn from snacks.bigfile: a bigfile don't need to be bit
> Try this, then FzfLua files, type a l p h i n e , it a bit stuck

Minified (one line) JavaScript libs are an issue, needs improvement for sure, at last it’s cached so should delay just once.

Note: need to revert the fallback delete from this PR, tested with a modeline I have in ghostty/config and it fails to detect properly because filetype.match doesn’t read my modeline (ft=config).

phanen:
> Note: need to revert the fallback delete from this PR, tested with a modeline I have in ghostty/config and it fails to detect properly because filetype.match doesn’t read my modeline (ft=config).

I didn't delete it: https://github.com/ibhagwan/fzf-lua/blob/89b670f1cd1b15ed116b9385d2ae6cc195eef98e/lua/fzf-lua/previewer/builtin.lua#L1030-L1035.

Maybe another issue.

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -992,12 +1028,6 @@ function Previewer.buffer_or_file:do_syntax(entry)
             -- this helps avoid losing highlights for help buffers
             -- which are '.txt' files with 'ft=help'
             pcall(function() vim.bo[bufnr].filetype = "help" end)
-          else
-            -- prepend the buffer number to the path and
-            -- set as buffer name, this makes sure 'filetype detect'
-            -- gets the right filetype which enables the syntax
```
ibhagwan, line 998: This deletion

ibhagwan:
> I didn't delete it:

added comment on the commit.

phanen:
ghostty is not detected by neovim core, but by `ftdetect.vim` packaged by linux distro:
https://github.com/ghostty-org/ghostty/blob/cfea2ea12cf1ef805659ffeae058f03b4639c788/src/build/GhosttyResources.zig#L164-L167

ibhagwan:
> Not sure why this have higher priority than do_modelines, do_modelines should always prefered if have.

You’re right.

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -928,11 +936,41 @@ function Previewer.base:update_render_markdown()
...
   local bufnr = self.preview_bufnr
   local preview_winid = self.win.preview_winid
+  -- snacks.picker/bigfile
+  vim.b[bufnr].bufpath = entry.path
+  -- a different way to determine buffer..
+  -- fakename makes sure 'filetype detect' gets the right filetype (seems not really useful now)
+  -- prepend the buffer number to the path and
+  -- set as buffer name, this makes sure 'filetype detect'
+  -- gets the right filetype which enables the syntax
+  local tempname = path.join({ tostring(bufnr), entry.path })
```
ibhagwan: I’m not sure about this change, if it doesn’t help us with snacks, this may have other adverse effects, also `entry.path` isn’t guaranteed, might be a URI.

phanen: I revert it.

phanen: Is `vim.b.xx` a anti pattern? I think it's good enough... as long as we don't assign a complex structure e.g. table to it

ibhagwan: > Is `vim.b.xx` a anti pattern? I think it's good enough... as long as we don't assign a complex structure e.g. table to it

I don’t see any issue with `vim.b`

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -983,6 +1016,7 @@ function Previewer.buffer_or_file:do_syntax(entry)
             vim.b[bufnr]._ft = ft
             self:update_render_markdown()
             self:update_ts_context()
+            self:attach_snacks_image()
```
ibhagwan: Is this the right place to call snacks attach? This ties the snacks to having successful ts attach no?

I also don’t think it should be inside `do_syntax` which won’t be called if the user disables previewer syntax highlighting, maybe more suitable within `Previewer.buffer_or_file:preview_buf_post`?

phanen: Indeed. It snacks.image require treesitter query (image.scm) but not require treesitter highlighting.

phanen: ~Hmm... but snacks.image require filetype detection, now this part is bundled together inside `do_syntax`, so https://github.com/ibhagwan/fzf-lua/pull/1913/commits/e52bc4946c059523c0a234f761cc92af084cdc3f don't work with `opts.syntax_delay`~ Fixed.

ibhagwan:
Ty @phanen, I think it makes more sense now, I had a feeling `_ft` might be of use one day :)

phanen:
I see why. `modeline` don't work for scratch buffer.

phanen:
Find a new option called `eventignorewin` (better than `eventignore`), but require `v0.11.0+`:
```lua
    api.nvim_win_call(self.win.preview_winid, function()
      local save_ei = vim.wo.eventignorewin
      local tempname = path.join({ tostring(bufnr), entry.path })
      pcall(api.nvim_buf_set_name, bufnr, tempname)
      vim.wo.eventignorewin = "FileType"
      pcall(api.nvim_buf_call, bufnr, function() vim.cmd("filetype detect") end)
      local ft = vim.bo[bufnr].ft
      if not ts_attach(bufnr, ft) then
        pcall(function() vim.bo[bufnr].syntax = vim.bo[bufnr].ft end)
      end
      vim.wo.eventignorewin = save_ei
    end)
```

ibhagwan:
> Should we set modeline=true or implement a ft-only modeline detecter? I don't think set modeline for each buffer is more dangerous than enable snacks.image for each markdown...

I don’t think it should be an issue, maybe a performance issue on very large buffers if the modeline is at the end?

We can also just say it’s a known issue and call it a day, I don’t think it’s a big deal if the preview buffer has the wrong syntax when modeline.

> Find a new option called eventignorewin (better than eventignore), but require v0.11.0+:

Nice find I didn’t know about it, if this solves this easily why not, otherwise I feel it’s too much work just for mode line.

ibhagwan:
> I split filetype detect out in do_syntax. And use early return. (fix a edgecase: if fallback to "filetype detect" then ts highlight is never considered before).

Previewer had a feature-creep over time, refactoring some of the logic is welcome.


> I still have no idea why sometime e.g. ghostty syntax don't work correctly with eventignore. I didn't found upstream issue. Maybe I'll look into it later...

Waste of your time, not critical :)

phanen:
Never mind it's a upstream issue.

https://github.com/ghostty-org/ghostty/pull/7119

ibhagwan:
> Never mind it's a upstream issue.
> 
> [ghostty-org/ghostty#7119](https://github.com/ghostty-org/ghostty/pull/7119)

Wow, what  a crazy investigation you’ve done on this one :)

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -1034,6 +1034,7 @@ function Previewer.buffer_or_file:do_syntax(entry)
   vim.b[bufnr]._ft = ft
 
   local ts_enabled = (function()
+    if (bytes / lcount) > 1000 then return false end
```
ibhagwan: Great idea, I didn’t think about this, should we alert the user TS is disabled for this file like we do for the bytes/lines limit or maybe remove the warnings altogether to make the behavior consistent?

phanen: This is a stolen of `snacks.bigfile`: https://github.com/folke/snacks.nvim/blob/8deea64dba3b9b8f57e52bb6b0133263f6ff171f/lua/snacks/bigfile.lua#L53-L54.
I don't think anyone would like to enable treesitter in a big minified file...

~Maybe a new options instead of adding warnings.~
~Maybe we can make `treesitter.enabled` accept a function..~
~No that's more confusing to me now.~

I make it a private options with no warnings. We can refactor if more user requesting for complex customization (e.g. per-ft limit maybe, then we can make treesitter.enabled a function).

ibhagwan: Agree, this doesn’t really need a warning.

ibhagwan:
Do you feel this is ready to merge @phanen?

phanen:
Still some problems:
* The scrollbar is incorrect when scroll to the bottom of a buffer with image. (https://github.com/ibhagwan/fzf-lua/pull/1913/commits/5a28850b8b0a9488ff1ff8cbcfde7c548d1de898 didn't fix it correctly)
* Image didn't update on preview win resizing.

phanen:
It seems correct now.

ibhagwan:
I'm mighty confused lol, I tried testing the snacks.image enabled property with `:FzfLua files previewer.snacks_image.enabled=false` and realized it didn't work eventhough `enabled=false` properly (the image is still rendered?), I then went ahead and actually deleted the entire `Previewer.base:attach_snacks_image` function and the calls at lines 1194/1199 and the previewer still renders images using snacks.image... tried the commit before this PR and this doesn't happen.

Still testing, but something within these commits is triggering snacks.image in the preview buffer even without the attach function :-)

Another bugs I found:
- Toggle preview off/on with F4 causes the preview buf to have line numbers
image

- After the line numbers appear, F2 fullscreen causes the image to disappear
image

ibhagwan:
Ok mystery solved, these lines trigger snacks.image regardless of the enabled property:
https://github.com/ibhagwan/fzf-lua/blob/cfa22f72636b6a062e3c4219e39032cdce68d156/lua/fzf-lua/previewer/builtin.lua#L836-L842

ibhagwan:
This is pretty cool to see, even snacks itself doesn't render it's' own inline images :-)

Fzf-lua's README with fzf-lua:
image

With snacks picker (no render-markdown or inline images):
image

PR #1914, opened 2025-03-22
ci: fix flaky test by adding wait time

midchildan:
This attempts to fix a test failure that's happening in the [Nixpkgs](https://github.com/NixOS/nixpkgs) build.

- [Build log](https://cache.nixos.org/log/62gmgc5p9n1pflivl0k5xnwdj7fb65nr-lua5.4-fzf-lua-0.0.1783-1.drv)
- [Archived build log](https://web.archive.org/web/20250322234335/https://cache.nixos.org/log/62gmgc5p9n1pflivl0k5xnwdj7fb65nr-lua5.4-fzf-lua-0.0.1783-1.drv)

I also encountered the same error on my own GitHub Actions workflow, and was able to resolve it by applying this fix.

- [Failing commit](https://github.com/midchildan/dotfiles/commit/3f82a4e82bbfd9d1471ba40863d19ed24cb949ee)
- [Fixed commit](https://github.com/midchildan/dotfiles/commit/dcf915a4af10b18d7c8972cfb97e86f833fdb1b0)

I'll make this a draft PR for now, because it's untested on the Nixpkgs build servers.

### DIFF
--- a/tests/file/ui_spec.lua
+++ b/tests/file/ui_spec.lua
@@ -162,6 +162,7 @@ T["files()"]["executable"] = new_set({ parametrize = { { "fd" }, { "rg" }, { "fi
     child.wait_until(function()
       return child.lua_get([[_G._fzf_load_called]]) == true
     end)
+    sleep(100)
     child.expect_screen_lines(screen_opts)
     child.type_keys("<c-c>")
     child.wait_until(function()


ibhagwan:
I’m not a big fan of “sleep” fixes, perhaps there’s a better way with a more distinctive condition to wait for.

I’d prefer to understand why this fails and maybe add a ‘wait for’ certain event in order to fix this properly.

midchildan:
Looking at the output, I think it's just taking too long to spawn and obtain output from the `find` process.

```
>                                      ⠴ 1/1 (0)
```

Or is this already supposed to wait until FZF has all the candidates?

```lua
    child.wait_until(function()
       return child.lua_get([[_G._fzf_load_called]]) == true
     end)
```

ibhagwan:
> Or is this already supposed to wait until FZF has all the candidates?

This waits for fzf’s load event, according to the manual:
```
       load
              Triggered when the input stream is complete and the initial
              processing of the list is complete.

              e.g.
                   # Change the prompt to "loaded" when the input stream is
              complete
                   (seq 10; sleep 1; seq 11 20) | fzf --prompt 'Loading> '
              --bind 'load:change-prompt:Loaded> '

```

phanen:
Maybe try `result`
       result
              Triggered when the filtering for the current  query  is  complete
              and the result list is ready.

ibhagwan:
> Maybe try `result`
> 
> ```
>        result
>               Triggered when the filtering for the current  query  is  complete
>               and the result list is ready.
> ```

I thought about result but I didn’t test if it triggers when no query is used.

phanen:
Ok, result is triggered even before `load`..
```
{ find; sleep 2; } | fzf --bind 'result:execute-silent:notify-send res "$(date)"' --bind 'load:execute-silent:notify-send load "$(date)"'
```

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/a56dfa157527c127100a54d52e7ccabfef5a728b

@midchildan, the above should solve your issue, lmk if otherwise.

Issue #1810, opened 2025-02-08
Bug: User mapping alt-h shadows flag binding for toggle_hidden


kaddkaka:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

bash

### Neovim version (`nvim --version`)

VIM v0.10.4

### Fzf version (`fzf --version`)

`0.44.1 (debian)`  (maybe I receive newer verion from vim-plug)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--no-mouse

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua

```

### Describe the bug / steps to reproduce

1. launch min.sh
2. `ctrl-p`
3. hold down `alt-h` for 10~15 seconds

The picker window hangs and goes empty:

![Image](https://github.com/user-attachments/assets/8239f39b-71a7-469b-935c-2d1bcf763135)

Trying to get out of this state (by pressing `ctrl-\` I received this error message:
```
Error executing vim.schedule lua callback: ...vim/site/pack/vendor/start/fzf-lua/lua/fzf-lua/shell.lua:313: vim/shared.lua:0: t: expected table, got nil
stack traceback:
        [builtin#36]: at 0x5b97d06da200
        ...vim/site/pack/vendor/start/fzf-lua/lua/fzf-lua/shell.lua:313: in function 'fn'
        ...vim/site/pack/vendor/start/fzf-lua/lua/fzf-lua/shell.lua:77: in function <...vim/site/pack/vendor/start/fzf-lua/lua/fzf-lua/shell.lua:76>
```

And ended up in the buffers picker:
![Image](https://github.com/user-attachments/assets/24c41e04-58e2-48aa-93af-649e789b8418)

Is this a case of "if it hurts don't do it"?

ibhagwan:
> Is this a case of "if it hurts don't do it"?

Still shouldn’t happen, I did  a similar test many times with live_grep ctrl-g and was never able to get it to hang, I’ll try with alt-h.

ibhagwan:
[...]

PR #1915, opened 2025-03-23
fix: close pipe when actions error

phanen:
Fix https://github.com/ibhagwan/fzf-lua/discussions/1901#discussioncomment-12588036.

If pipe is not closed, terminal won't response and refresh unless we
press a ctrl-c.

This can happened when:
```
actions["ctrl-g"].fn = function() error('xxx') end
```


phanen:
I guess this is also related: https://github.com/ibhagwan/fzf-lua/issues/1810. Keep pressing a key may trigger an error in actions.

ibhagwan:
Ty @phanen!

PR #1916, opened 2025-03-23
fix: don't ignore err in action

phanen:
Silent all error message make it hard to debug things...


nyngwang:
@phanen Is this related to #1915? I have tested that with the current HEAD bb3e837 and it seems that the freeze still happen on both:

1. Just query one time with `hello` and select a file.
2. First press `ctrl-g` action, type `hello` and select a file.

nenahp:
Weird, I think freeze should have been fixed.

ibhagwan:
Ty @phanen!

PR #1925, opened 2025-03-27
feat(lsp_code_actions): Use `apply` based on fzf `--select-1`

andrewbraxton:
This is an improvement on the default behavior of `--select-1` because it won't flash an fzf window momentarily before applying the code action when there is only one option.


ibhagwan:
Ty @andrewbraxton!

Maybe we can also add `apply = opts.jump1 or …` to the apply condition to match the LSP location pickers?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/1dcef2467af609e0688f8136b8ea720645c9c41c

Added jump1 and merged manually, Ty @andrewbraxton!

andrewbraxton:
Cool, was going to get to it today but that works too 😄

Issue #1929, opened 2025-03-27
Bug: Neovim 0.11 rendering bug


reybits:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

macOS

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.0

### Fzf version (`fzf --version`)

0.60.3 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--height 75% --layout=reverse --border --no-hscroll --info inline-right

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [x] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
            -- profile
            "hide", -- or "telescope",

            winopts = {
                width = 0.96,
                height = 0.96,
                preview = {
                    layout = "vertical",
                    vertical = "up:70%",
                },
            },
            keymap = {
                builtin = {
                    false,
                    ["<F1>"] = "toggle-help",
                    ["<M-/>"] = "toggle-help",

                    ["<M-S-f>"] = "toggle-fullscreen",

                    ["<M-p>"] = "toggle-preview",
                    ["<M-w>"] = "toggle-preview-wrap",

                    ["<M-c>"] = "toggle-preview-cw",
                    ["<M-S-c>"] = "toggle-preview-ccw",

                    ["<C-d>"] = "preview-page-down",
                    ["<C-u>"] = "preview-page-up",
                },
                fzf = {
                    false,
                    ["alt-a"] = "toggle-all",
                    ["alt-g"] = "first",
                    ["alt-G"] = "last",
                },
            },
            files = {
                hidden = false,
            },
            buffers = {
                prompt = "❯ ",
            },
            keymaps = {
                prompt = "❯ ",
                winopts = { preview = { hidden = true } },
            },
            actions = {
                files = {
                    -- true, -- if set to true, enables default settings.

                    ["enter"] = actions.file_edit_or_qf,
                    ["ctrl-q"] = actions.file_sel_to_qf,
                    ["alt-f"] = actions.toggle_follow,
                    ["alt-h"] = actions.toggle_hidden,
[...]

PR #1933, opened 2025-03-29
fix: nvim_open_win from ctx win

phanen:
See https://github.com/neovim/neovim/issues/33133#issuecomment-2763104765.

Temporally fix https://github.com/ibhagwan/fzf-lua/issues/1929.

*EDIT*: But it didn't handle float win not created by fzf...
* e.g. treesitter.context's win on ctx buffer
* e.g. lsp fidget's spinner win?

Although If you want, you can always override `api.nvim_win_open`,. I think it's usually harmless...

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -94,10 +94,12 @@ function TSContext.update(winid, bufnr, opts)
   else
     assert(context_lines)
     local function open()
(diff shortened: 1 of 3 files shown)


ibhagwan:
Ty @phanen  for your amazing work on this issue.

> Although If you want, you can always override api.nvim_win_open,. I think it's usually harmless...

I agree, but who knows what voodoo issues this might create for other plugins, I’d rather the workaround is fzf-lua local so it’s easier to isolate, if someone still experiences the issue we’ll know it’s another plug-in.

PR #1935, opened 2025-03-30
fix(previewer): scroll previewer on buffer without ft detected

phanen:
`:FzfLua files query=screenshots`, then scroll.


ibhagwan:
Ty @phanen, I didn’t realize we get there on no ft buffers.

PR #1936, opened 2025-03-30
fix(previewer): `nvim_win_call` fail when `ctx.winid` is invalid

phanen:
In `hide` profile, `:new`, `:FzfLua files`, `<esc>`, `:FzfLua resume`

### DIFF
--- a/lua/fzf-lua/test/helpers.lua
+++ b/lua/fzf-lua/test/helpers.lua
@@ -151,7 +151,13 @@ M.new_child_neovim = function()
   child.unload = function()
     -- Unload Lua module
     child.lua([[_G.FzfLua = nil]])
-    child.lua([[package.loaded["fzf-lua"] = nil]])
+    child.lua([[
+      for k, v in pairs(package.loaded) do
+        if k:match("^fzf%-lua") then
+          package.loaded[k] = nil
+        end
+      end
+    ]])
 
     -- Remove global vars
     for _, var in ipairs({ "server", "directory", "root" }) do
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -1059,7 +1059,11 @@ function M.nvim_open_win(bufnr, enter, config)
 end
 
 function M.nvim_open_win0(bufnr, enter, config)
-  return vim.api.nvim_win_call(M.CTX().winid, function()
+  local winid = M.CTX().winid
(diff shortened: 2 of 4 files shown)


phanen:
But the behavior still not feels good. `resume` didn't trigger `unhide` first in this case idk why.

ibhagwan:
>  I wonder if mini.test can check screenshot in a "wait_until" way, like how neovim test suite does:

This would be a question for @echasnovski.

echasnovski:
> > I wonder if mini.test can check screenshot in a "wait_until" way, like how neovim test suite does:
> 
> This would be a question for @echasnovski.

This should be possible with a custom helper: `vim.wait` with callback that checks necessary condition. I don't think implementing that directly into `expect.reference_screenshot` is a good idea, as I personally don't really like the pattern. Having an explicit wait time and known in advance checks seems to be more robust testing long term.

phanen:
Thanks for you reply. I personally want to avoid use of hardcode wait time (since it can be differ from different platform)... I find it's a bit tricky to add this "wait" feature outside of mini.test. Would you like to take a look at my PR https://github.com/echasnovski/mini.nvim/pull/1716. @echasnovski

echasnovski:
Mostly answered in the 'mini.nvim' PR. Especially [this comment](https://github.com/echasnovski/mini.nvim/pull/1716#issuecomment-2764600219).

> I personally want to avoid use of hardcode wait time (since it can be differ from different platform)...

Yes, that's why it is possible to set different wait time depending on platform. That is the approach I do in 'mini.nvim' and it works fine. Implementing own helper outside of 'mini.test' internal methods should be still possible with something like this:

```lua
local wait_for_screenshot = function(screenshot_name, timeout, interval)
  interval = interval or 1
  local screenshot_path = 'tests/screenshots/' .. screenshot_name
  local ok, out
  local predicate = function()
    ok, out = pcall(MiniTest.expect.reference_screenshot, child.get_screenshot(), screenshot_path)
    return ok
  end
  vim.wait(timeout, predicate, interval)
  if ok then return end
  -- Doesn't really report error right now. Might require a full
  -- `new_expectation()` call.
  error(out)
end
```

Its usage then requires creating a reference screenshot manually at a known path and using this helper to test it with "wait for proper screen state" workflow.

ibhagwan:
rebased/fix up the 3rd commit and merged.

@phanen, tysm!

Issue #1949, opened 2025-04-05
Bug: Argument list picker's actions behaves abonormally


aristidebm:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

fish

### Neovim version (`nvim --version`)

NVIM v0.11.0 Build type: RelWithDebInfo LuaJIT 2.1.1741730670

### Fzf version (`fzf --version`)

0.60.3 (0012183e)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=bg+:-1,gutter:-1,bg:#303446,spinner:#f2d5cf,hl:#e78284 --color=fg:#c6d0f5,header:#e78284,info:#ca9ee6,pointer:#f2d5cf --color= marker:#babbf1,fg+:#c6d0f5,prompt:#ca9ee6,hl+:#e78284 --color=selected-bg:#51576d --multi

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require("fzf-lua").setup({
  grep = { prompt = 'Grep❯ ' },
  args = {
    prompt            = 'Args❯ ',
    files_only        = true,
    -- actions inherit from 'actions.files' and merge
    actions = {
            ["ctrl-x"] = false,
            ["ctrl-d"] = { fn =  require("fzf-lua").actions.arg_del, reload = true }
    },
  },
  -- fzf_colors = true,
  fzf_colors = {
      -- make gutter color and current line background color match
      -- the terminal color
      ["bg+"] = { "bg", { "Normal" } },
  },
  -- show additional informations on each entry right
  -- defaults = { formatter = 'path.filename_first' },
  winopts = {
     split = 'bo new',
     backdrop = 100,
     height = 0.7,
     width = 0.9,
     border = 'single',
     preview = {
         hidden = "hidden"
     }
  },
  keymap = {
    fzf = {
        -- fzf binary commands prefix (you can check here https://github.com/junegunn/fzf/blob/d938fdc496ccecfe5d747500927b675d31e1835a/src/options.go#L1390a for more context) 
[...]

PR #1950, opened 2025-04-06
fix(actions): argdel only accept relative path

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1949.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -211,7 +211,8 @@ M.vimcmd_entry = function(_vimcmd, selected, opts, pcall_vimcmd)
           -- Force full paths when `autochdir=true` (#882)
           vimcmd = string.format("%s %s", vimcmd, (function()
             -- `:argdel|:argadd` uses only paths
-            if vimcmd:match("^arg") then return entry.path end
+            -- argdel only accepts relative path (#1949)
+            if vimcmd:match("^arg") then return path.relative_to(entry.path, uv.cwd()) end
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @phanen, I thought about that too but what if the arg was added with full path? We will have the opposite issue.

I also thought of adding `|argdedupe` to the arg_add action, I didn’t know it existed but I think would be a good addition.

ibhagwan:
> ~I didn't realize that. Another hack way is create a /tmp dir somewhere, cd to it, argsdel, cd back.~

Or we can pcall twice one with relative and one full path? Or try to get info from `vim.fn.argc`? I thought of a few ways but didn’t get a chance to test yet as AFK.

phanen:
> I thought about that too but what if the arg was added with full path

Just found if we argadd a fullpath, `args`/`fn.argv()` will show the relative path to cwd. If `relpath_to` is implemented correctly. This approach won't be wrong.

ibhagwan:
> > I thought about that too but what if the arg was added with full path
> 
> Just found if we `:argadd`/`/bin/nvim fullpath` a fullpath, `:args`/`fn.argv()` will show the relative path to cwd. If `relpath_to` is implemented correctly. This approach won't be wrong.

That depends on how you expect it to behave when the second path isn’t part of the first, see the test cases:
https://github.com/ibhagwan/fzf-lua/blob/6488ada2f376e47789391dc353b6d91a3f9de6f6/tests/path_spec.lua#L212-L270

ibhagwan:
Ty @phanen!

PR #1954, opened 2025-04-06
fix(previewer): Apply fn_transform_cmd to regex query parsing

### LINKED ISSUE #1927: Feature: Improve git grep Support for Glob Patterns
### Have you RTFM'd?

### Feature Request

First of all, thank you for this amazing plugin! It has been incredibly useful, and I really appreciate the work that has gone into it.

Currently, when using `rg` to search with a glob pattern, the command is correctly parsed as:
```
rg --iglob *.lua' -e 'test'
```
However, when performing the same search with `git grep`, it is parsed as:
```
git grep 'test -- *.lua'
```
This results in an incorrect search behavior.

While `rg_glob_fn` allows users to customize glob patterns, the `rg_glob` option is currently specific to `rg`, and there is no equivalent way to customize glob patterns for `git grep`.

Would it be possible to use glob patterns with `git grep`?
If not, could you consider adding support for this feature?

drop-stones:
This PR addresses an issue where the `fn_transform_cmd` wasn't being applied to the regex query when parsing.
Now, before parsing the regex query, the `fn_transform_cmd` is checked and applied if it exists.
This allows the query to be retrieved and transformed based on user configuration.

Please review and let me know if anything needs adjustment. Thank you!

Fixes #1927

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1040,7 +1040,10 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
   local mgrep, glob_args = require("fzf-lua.providers.grep"), nil
   local regex = self.opts.__ACT_TO == mgrep.grep and self.opts._last_query
       or self.opts.__ACT_TO == mgrep.live_grep and self.opts.search or nil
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @drop-stones this is perfect :)

Issue #950, opened 2023-12-12
live grep > E484: Can't open file /usr/local/share/nvim/syntax/syntax.vim


mikesmithgh:
### Info

- Operating System: macos
- Shell: bash GNU bash, version 5.2.21(1)-release (aarch64-apple-darwin23.0.0)
- Terminal: kitty 0.31.0 created by Kovid Goyal
- `nvim --version`:
```
NVIM v0.10.0-dev-1819+gb40170f7a-dirty
Build type: RelWithDebInfo
LuaJIT 2.1.1702233742
```
- `fzf --version`: 0.44.1 (brew)

- [x] The issue is reproducible with `mini.sh`

<details>
<summary>fzf-lua configuration</summary>

```lua
require('fzf-lua').setup({
})
```
</details>

### Description 
When I use live-grep I am getting errors about opening `syntax.nvim`.

It appears to be looking for:
```
ADDITIONAL DATA DIRECTORY (DEFAULT)  
                  *$XDG_DATA_DIRS*              Nvim: stdpath("data_dirs")
    Unix:         /usr/local/share            /usr/local/share/nvim
                  /usr/share                  /usr/share/nvim
    Windows:      Not applicable              Not applicable
```

I build neovim nightly and have the environment variable set `VIMRUNTIME=/Users/mike/gitrepos/neovim/runtime`

`/usr/local/share/nvim` and `usr/share/nvim` do not exist on my machine

### Demo

https://github.com/ibhagwan/fzf-lua/assets/10135646/6f25bf3d-e818-4c95-a446-2a597ae76978

ibhagwan:
Can you run `:FzfLua live_grep debug=true` and post the command that will appear under `[DEBUG]` as the first outpt line?

Then run `:new | term` to open a new neovim terminal and run the same command inside?

mikesmithgh:
sure, I ran it but the error doesn't seem to be coming from rg. Let me know if you need me to try anything else 👍

:FzfLua live_grep debug=true
```
*Rg> a                                                                                                        
  3/3 (0)                                                                                                     
  :: <ctrl-g> to Fuzzy Search                                                                                 
󰅂  [DEBUG]: rg --column --line-number --no-heading --color=always --smart-case --max-columns=4096 -e 'a'      
   󰈙 a.txt:1:1:a                                                                                              
   E484: Can't open file /usr/local/share/nvim/syntax/syntax.vim                                              
                                          
```
:new | term
```bash
[...]

Issue #716, opened 2023-04-04
Barebones lazy setup fails to execute `serverstart`


mattiasbonte:
### Info

- Operating System: MacOs
- Shell: Zsh
- Terminal: Wezterm
- `nvim --version`: nightly
- `fzf --version`:  0.38 (brew)

- [ ] The issue is reproducible with `mini.sh`

<details>
<summary>fzf-lua configuration</summary>

`.config/lazytest/lua/plugins/fzf.lua`
```lua
return {
	{
		"ibhagwan/fzf-lua",
		dependencies = { "junegunn/fzf" },
		cmd = { "FzfLua" },
        }
}
```
</details>

### Description 

Hello! I'm trying to set up fzf-lua in a barebones lazy.nvim installation, but fail at something vital it seems.
I did use the method described in [this video](https://www.youtube.com/watch?v=LkHjJlSgKZY) where I use t[his newly implemented feature](https://github.com/neovim/neovim/pull/22128) in the nightly version. So it's not my main nvim installation but one at a different path.

This is the error I am seeing.

```lua
E5108: Error executing lua: Vim(lua):E5108: Error executing lua Vim:Failed to start server: no such file or directory
stack traceback:
	[C]: in function 'serverstart'
	.../nvim_distros/lazytest/lazy/fzf-lua/lua/fzf-lua/init.lua:28: in main chunk
	[C]: in function 'require'
	...e/nvim_distros/lazytest/lazy/fzf-lua/lua/fzf-lua/cmd.lua:2: in main chunk
	[C]: in function 'require'
	[string ":lua"]:1: in main chunk
	[C]: at 0x0102cb0220
stack traceback:
	[C]: at 0x0102cb0220
```
When I check out the file where this error is thrown, it's apparently trying to start a RTP server? I've looked into your personal dotfiles for a clue to what I might be doing wrong but can't really see anything special with your lazy setup that could help me.

Anyway, Im a bit lost in what all this means so I thought let's ask the professional :)! 

Any idea what might be going wrong?

Thanks in advance for your time.

ibhagwan:
> It's apparently trying to start a RTP server? 

Correct, fzf-lua uses process sepration for most of the heavy lifiting, the named pipe is how the child process communicates with the main neovim instance for querying users's config and other options.

What is the exact version of your nightly?

[...]

PR #1955, opened 2025-04-06
fix(rpc): use `nvim -l` to get argv easilier (also less side effects)

phanen:
1. `nvim -l` still quit immediately even with `coroutine.yield`...
```
local co = coroutine.create(function()
  coroutine.yield()
  os.exit(0)
end)

vim.defer_fn(function()
  print('defer done')
  coroutine.resume(co)
end, 1000)

coroutine.resume(co)
```

2. `vim.wait` don't work on windows?

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -279,6 +279,8 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
       ["SHELL"] = shell_cmd[1],
       ["FZF_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
       ["SKIM_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
+      ["FZF_LUA_SERVER"] = vim.g.fzf_lua_server,
+      ["VIMRUNTIME"] = vim.env.VIMRUNTIME,
       ["FZF_DEFAULT_OPTS"] = (function()
         -- Newer style `--preview-window` options in FZF_DEFAULT_OPTS such as:
         --   --preview-window "right,50%,hidden,<60(up,70%,hidden)"
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -89,11 +89,26 @@ M.spawn = function(opts, fn_transform, fn_done)
   else
     table.insert(args, tostring(opts.cmd))
   end
+
+  ---@diagnostic disable-next-line: missing-fields
   local handle, pid = uv.spawn(shell, {
     args = args,
     stdio = { nil, output_pipe, error_pipe },
     cwd = opts.cwd,
-    env = opts.env,
+    ---@diagnostic disable-next-line: assign-type-mismatch
+    env = (function()
+      -- uv.spawn will override all env when table provided?
+      -- steal from $VIMRUNTIME/lua/vim/_system.lua
+      local env = vim.fn.environ() --- @type table<string,string>
+      env["NVIM"] = vim.v.servername
+      env["NVIM_LISTEN_ADDRESS"] = nil
+      env = vim.tbl_extend("keep", opts.env or {}, env or {})
+      local renv = {} --- @type string[]
+      for k, v in pairs(env) do
+        renv[#renv + 1] = string.format("%s=%s", k, tostring(v))
+      end
+      return renv
+    end)(),
     verbatim = _is_win,
   }, function(code, signal)
     on_exit_called = true
@@ -675,18 +690,12 @@ M.wrap_spawn_stdio = function(opts, fn_transform, fn_preprocess, fn_postprocess)
   assert(opts and type(opts) == "string")
   assert(not fn_transform or type(fn_transform) == "string")
   local nvim_bin = os.getenv("FZF_LUA_NVIM_BIN") or vim.v.progpath
-  local nvim_runtime = os.getenv("FZF_LUA_NVIM_BIN") and ""
-      or string.format(
-        _is_win and [[set VIMRUNTIME=%s& ]] or "VIMRUNTIME=%s ",
-        _is_win and vim.fs.normalize(vim.env.VIMRUNTIME) or M.shellescape(vim.env.VIMRUNTIME)
-      )
   local lua_cmd = ("lua %sloadfile([[%s]])().spawn_stdio(%s,%s,%s,%s)"):format(
     _has_nvim_010 and "vim.g.did_load_filetypes=1; " or "",
     vim.fn.fnamemodify(_is_win and vim.fs.normalize(__FILE__) or __FILE__, ":h") .. "/spawn.lua",
     opts, fn_transform, fn_preprocess, fn_postprocess
   )
-  local cmd_str = ("%s%s -n --headless -u NONE -i NONE --cmd %s"):format(
-    nvim_runtime,
+  local cmd_str = ("%s -n --headless -u NONE -i NONE --cmd %s"):format(
     M.shellescape(_is_win and vim.fs.normalize(nvim_bin) or nvim_bin),
     M.shellescape(lua_cmd)
   )
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -51,6 +51,8 @@ end
 ---@param opts {cwd: string, cmd: string|table, env: table?, cb_finish: function, cb_write: function, cb_err: function, cb_pid: function, fn_transform: function?, EOL: string?, process1: boolean?, profile: boolean?}
 ---@param fn_transform function?
 ---@param fn_done function?
+---@return uv.uv_process_t proc
+---@return integer         pid
 M.spawn = function(opts, fn_transform, fn_done)
   local EOL = opts.EOL or "\n"
   local output_pipe = uv.new_pipe(false)
@@ -233,6 +235,7 @@ M.spawn = function(opts, fn_transform, fn_done)
     output_pipe:read_start(read_cb)
     error_pipe:read_start(err_cb)
   end
+  return handle, pid
 end
 
 M.async_spawn = coroutinify(M.spawn)
@@ -690,14 +693,10 @@ M.wrap_spawn_stdio = function(opts, fn_transform, fn_preprocess, fn_postprocess)
   assert(opts and type(opts) == "string")
   assert(not fn_transform or type(fn_transform) == "string")
   local nvim_bin = os.getenv("FZF_LUA_NVIM_BIN") or vim.v.progpath
-  local lua_cmd = ("lua %sloadfile([[%s]])().spawn_stdio(%s,%s,%s,%s)"):format(
-    _has_nvim_010 and "vim.g.did_load_filetypes=1; " or "",
-    vim.fn.fnamemodify(_is_win and vim.fs.normalize(__FILE__) or __FILE__, ":h") .. "/spawn.lua",
-    opts, fn_transform, fn_preprocess, fn_postprocess
-  )
-  local cmd_str = ("%s -n --headless -u NONE -i NONE --cmd %s"):format(
+  local cmd_str = ("%s -u NONE -l %s %s"):format(
     M.shellescape(_is_win and vim.fs.normalize(nvim_bin) or nvim_bin),
-    M.shellescape(lua_cmd)
+    vim.fn.fnamemodify(_is_win and vim.fs.normalize(__FILE__) or __FILE__, ":h") .. "/spawn.lua",
+    M.shellescape(("return %s,%s,%s,%s"):format(opts, fn_transform, fn_preprocess, fn_postprocess))
   )
   return cmd_str
 end
--- a/lua/fzf-lua/shell_helper.lua
+++ b/lua/fzf-lua/shell_helper.lua
@@ -24,7 +24,6 @@ end
 
 local preview_socket, preview_socket_path = get_preview_socket()
 
-local chan_id
 local preview_receive_socket
 uv.listen(preview_socket, 1, function(_)
   preview_receive_socket = uv.new_pipe(false)
@@ -43,22 +42,22 @@ uv.listen(preview_socket, 1, function(_)
     if not data then
       uv.close(preview_receive_socket)
       uv.close(preview_socket)
-      if _is_win then -- uv.stop() quit slow on windows
-        os.exit(0)
-      end
-      uv.stop() -- exit
+      -- on windows: ci fail when use uv.stop()
+      -- on linux: zero event can freeze https://github.com/ibhagwan/fzf-lua/pull/1955#issuecomment-2785474217
+      -- uv.stop()
+      os.exit(0)
       return
     end
     io.write(data)
   end)
 end)
 
-local function rpc_nvim_exec_lua(opts, fzf_selection)
+local rpc_nvim_exec_lua = function(opts)
   local success, errmsg = pcall(function()
     -- for skim compatibility
     local preview_lines = vim.env.FZF_PREVIEW_LINES or vim.env.LINES
     local preview_cols = vim.env.FZF_PREVIEW_COLUMNS or vim.env.COLUMNS
-    chan_id = vim.fn.sockconnect("pipe", vim.env.NVIM, { rpc = true })
+    local chan_id = vim.fn.sockconnect("pipe", vim.env.FZF_LUA_SERVER or vim.env.NVIM, { rpc = true })
     vim.rpcrequest(chan_id, "nvim_exec_lua", [[
       local luaargs = {...}
       local function_id = luaargs[1]
@@ -71,10 +70,11 @@ local function rpc_nvim_exec_lua(opts, fzf_selection)
     ]], {
       opts.fnc_id,
       preview_socket_path,
-      fzf_selection,
+      opts.fzf_selection,
       tonumber(preview_lines),
       tonumber(preview_cols),
     })
+    vim.fn.chanclose(chan_id)
   end)
 
   if not success or opts.debug then
@@ -94,19 +94,15 @@ local function rpc_nvim_exec_lua(opts, fzf_selection)
     os.exit(1)
   end
 
-  uv.run() -- return when uv.stop()
-  vim.fn.chanclose(chan_id)
+  uv.run("once")
+  uv.run() -- noreturn, quit by os.exit
 end
 
-
 local args = vim.deepcopy(_G.arg)
--- remove filename
-args[0] = nil
--- remove "--"
-table.remove(args, 1)
--- remove serialized opts
+args[0] = nil -- remove filename
 local opts = {
   fnc_id = tonumber(table.remove(args, 1)),
   debug = table.remove(args, 1) == "true",
+  fzf_selection = args
 }
-rpc_nvim_exec_lua(opts, args)
+rpc_nvim_exec_lua(opts)
--- a/lua/fzf-lua/shell_helper.lua
+++ b/lua/fzf-lua/shell_helper.lua
@@ -4,6 +4,10 @@ local uv = vim.uv or vim.loop
 
 local _is_win = vim.fn.has("win32") == 1 or vim.fn.has("win64") == 1
 
+if vim.v.servername and #vim.v.servername > 0 then
+  pcall(vim.fn.serverstop, vim.v.servername)
+end
+
 ---@return string
 local function windows_pipename()
   local tmpname = vim.fn.tempname()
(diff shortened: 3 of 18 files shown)


ibhagwan:
> nvim -l still quit immediately even with coroutine.yield...

because we have to yield to the thread outside of the coroutine.create function, not sure if this is possible, maybe there is a libuv function that waits for a thread, I’ll look.

ibhagwan:
```
uv.thread_join({thread})                                      *uv.thread_join()*

                > method form `thread:join()`

                Parameters:
                - `thread`: `luv_thread_t userdata`

                Waits for the `thread` to finish executing its entry function.

                Returns: `boolean` or `fail`
```

ibhagwan:
I wonder why this fails on Windows, I no longer have access to a Windows machine so I can’t play around with it.

phanen:
Even on linux, I'm confusing about the behavior of `uv.run()`, sometimes it wait for all callback then infinitely hang? Sometimes it work like `uv.run("once")` (when put it at the last line of file).

ibhagwan:
> The linux ci passed now only because uv.run wait all callback and hang but then fzf kill it. (leaving dangling tmpfiles)

Not optimal for sure

phanen:
> The linux ci passed now only because `uv.run` wait all callback and hang but then fzf kill it. (leaving dangling tmpfiles)

This description is not correct. fzf didn't kill it. It just cannot return from `uv.run`, so we need to delete tmpfile in callback.

> Even on linux, I'm confusing about the behavior of `uv.run()`, sometimes it wait for all callback then infinitely hang? Sometimes it work like `uv.run("once")` (when put it at the last line of file).

Remove or defer `fn.chanclose`, then `uv.run` will be no return.

Now the only problem is windows: `uv.run` seems quit immediately.

ibhagwan:
> Now the only problem is windows: uv.run seems quit immediately.

I’m asking myself is all this hassle worth it for the only edge case when action argument is the pwd? Even if we make this work, what other issues might dome up from this change…

phanen:
Nvim will load all argv as buffer... so I guess this way maybe a slightly faster (but I didn't feel the change), and some other edges...
```
FzfLua live_grep query=/dev/null
```

nenahp:
* `opts.fzf_lua_server` seems unusable on windows (reuse vim.env.NVIM here... but on linux there's still some flatpak problem related to naming? 
* something seems wrong with shellescape on windows, so I just avoid serialized a table...
* `uv.stop` take much long time on windows, so just `os.exit` immediately

phanen:
* `opts.fzf_lua_server` seems unusable on windows (reuse vim.env.NVIM here... but on linux there's still some flatpak problem related to naming? 
* something seems wrong with shellescape on windows, so I just avoid serialized a table...
* `uv.stop` take much long time on windows, so just `os.exit` immediately

ibhagwan:
Looks much better now without the string serialized args :)

lua/fzf-lua/libuv.lua
```diff
@@ -89,11 +89,25 @@ M.spawn = function(opts, fn_transform, fn_done)
...
+      -- uv.spawn will override all env when table provided?
+      -- steal from $VIMRUNTIME/lua/vim/_system.lua
+      local env = vim.fn.environ() --- @type table<string,string>
+      env["NVIM"] = vim.v.servername
+      env["NVIM_LISTEN_ADDRESS"] = nil
+      env = vim.tbl_extend("keep", opts.env or {}, env or {})
+      local renv = {} --- @type string[]
+      for k, v in pairs(env) do
+        renv[#renv + 1] = string.format("%s=%s", k, tostring(v))
+      end
```
ibhagwan, line 111: How is this working without `return renv`?

phanen: I forgot it. But it still work. BTW, why we need to pass a $VIMRUNTIME in cli args before?

phanen: I see: https://github.com/ibhagwan/fzf-lua/issues/950.

phanen:
Fail to to make `nvim -l spawn.lua` pass the ci so I revert it finally (although it seems working well if run fzf-lua interactively). We're still able to use `shell_helper` to forward processed stream from fzf-lua server. (Didn't try but if I understand correctly, I guess ci can be passed if `multiprocess=false`).

I need to figure out the behavior of `vim.{api,fn,rpcrequst,...}` (required by `fn_{transform,pre/postprocess}`) in luv callback (when `nvim -l`). Neovim upstream will spawn `nvim -l/-ll` for testsuite, but only native lua is used (no 
 `vim.api`, and `vim.rpcrequest` is also never used (it just use a own lua client built from scratch via `vim.mpack` https://github.com/neovim/neovim/tree/700a25e6218e016b5adb0ddee740be4618d717a2/test/client).

lua/fzf-lua/libuv.lua
ibhagwan: > I see: #950.

I’m wondering is VIMRUNTIME still required with the `-l` change.

ibhagwan:
> Fail to to make nvim -l spawn.lua pass the ci so I revert it finally (although it seems working well if run fzf-lua interactively).

Does the CI fail on all OS’s or just Windows?

lua/fzf-lua/fzf.lua
```diff
@@ -279,6 +279,8 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
       ["SHELL"] = shell_cmd[1],
       ["FZF_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
       ["SKIM_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
+      ["FZF_LUA_SERVER"] = vim.g.fzf_lua_server or vim.v.servername,
```
ibhagwan: Not sure I understand why `vim.v.servername` is used here? Why is `vim.g.fzf_lua_server` ever `nil`?

phanen: I also use `vim.env.FZF_LUA_SERVER or vim.env.NVIM` in `shell_helper`
Just defensive programming, should I remove it? It's harmless though.

phanen:
> Does the CI fail on all OS’s or just Windows?

All os on several same test: https://github.com/ibhagwan/fzf-lua/actions/runs/14314121577?pr=1955.

> I might be wrong but if I recall the code correctly multiprocess has no effect on shell_helper? Multiprocess only wraps string commands using spawn.lua?

I means if we use `multiprocess=false`, then `spawn.lua` will not used. And `shell_helper` will be used instead? (since it can also print to stdout?

lua/fzf-lua/shell_helper.lua
```diff
@@ -20,40 +24,41 @@ end
...
-      -- while vim.v.argv provides access to the raw argument, however, vim.v.argv
-      -- contains the headless wrapper arguments so we need to index backwards
-      table.insert(fzf_selection,
-        _is_win and vim.v.argv[#vim.v.argv - nargs + 1 + i] or vim.fn.argv(i))
-    end
     -- for skim compatibility
     local preview_lines = vim.env.FZF_PREVIEW_LINES or vim.env.LINES
     local preview_cols = vim.env.FZF_PREVIEW_COLUMNS or vim.env.COLUMNS
-    local chan_id = vim.fn.sockconnect("pipe", opts.fzf_lua_server, { rpc = true })
+    chan_id = vim.fn.sockconnect("pipe", vim.env.FZF_LUA_SERVER or vim.env.NVIM, { rpc = true })
```
ibhagwan: When do we call shell helper outside of the fzf environment so  `FZF_LUA_SERVER` is empty?
If there is such a case and we connect to self (`NVIM`), wouldn’t it be better to skip the RPC call and call the function from the _registry directly (using the fnc_id)?

phanen: I try to make `shell_helper` usable if we copy debug info out then used it in cmdline. 
(before this commit it's always useable since we pass fzf-lua-server socket address in cli args). 
But just now I found this only can be used in shell from nvim's `:term`...

> If there is such a case and we connect to self (`NVIM`), wouldn’t it be better to skip the RPC call and call the function from the _registry directly (using the fnc_id)?

`$NVIM` always points to parent instance (not chlid self (`vim.v.servername`), which is never used and stoped already since we have dangling socket)

lua/fzf-lua/fzf.lua
ibhagwan: > I also use `vim.env.FZF_LUA_SERVER or vim.env.NVIM` in `shell_helper` Just defensive programming, should I remove it? It's harmless though.

No problem with being defensive but in some cases if we know with 100% certainty a certain var must exist IMHO it’s better to let this fail and we learn from the error what went wrong, otherwise we may get into a situation where things work but not quite right and we’ll have a hard time figuring out the “voodoo” issue.

Not sure about this one though, maybe this can workaround the `serverstart` issue with a bad XDG_RUNTIME_DIR (#716), or it might be better to let or fail as not having g:_fzf_lua_server` is critical enough to purposefully fail this?

lua/fzf-lua/shell_helper.lua
ibhagwan: > I try to make shell_helper usable if we copy debug info out then used it in cmdline.
(before this commit it's always useable since we pass fzf-lua-server socket address in cli args).

I see, so we can run an action without knowing/specifying fzf lua’a server address.

> But just now I found this only can be used in shell from nvim's :term...

Bummer

ibhagwan:
From the failed tests:

`-l ./scripts/make_cli.lua ;` - if I can see this (now revered) file maybe I’ll have an idea how to fix the 
 wait condition? Do we know which condition falls, load/postprocess (I’m assuming on_create succeeds as it’s not a shell action)?

ibhagwan:
If I can see the branch of the failed tests maybe I’ll have an idea how to fix it, I can see the load even failed, if we run the branch locally with a “load” event added in `setup` similar to the test helper, does it get executed?
https://github.com/ibhagwan/fzf-lua/blob/13ec251a963ed5645dc8ad49e0fde13dac42ec0f/lua/fzf-lua/test/helpers.lua#L141-L144

ibhagwan:
Looking at this https://github.com/ibhagwan/fzf-lua/commit/d99d21d238b6ada46e55c61cacb4275c2789b36a previous commit I think I know why the CI fails with spawn.

If the spawn command is run using FZF_DEFAULT_COMMAND it will inherit FZF_LUA_SERVER and there will be no issue but that’s not always the case, for example, the Windows tests use `opts.pipe_cmd` in this case the server env var won’t be inherited, to fix this the command must have the server env var prefixed (like the runtime was), so this line:
https://github.com/ibhagwan/fzf-lua/blob/13ec251a963ed5645dc8ad49e0fde13dac42ec0f/lua/fzf-lua/libuv.lua#L688-L689

Should resolve into `FZF_LUA_SERVER=… nvim -l …`

nenahp:
> Should resolve into `FZF_LUA_SERVER=… nvim -l …`

No, I `make test` locally, it don't fix...

phanen:
> Should resolve into FZF_LUA_SERVER=… nvim -l …

No, I make test locally, it don't fix...

ibhagwan:
> > Should resolve into FZF_LUA_SERVER=… nvim -l …
> 
> No, I `make test` locally, it don't fix...

Yes I realized after I wrote this was most likely not going to work.

phanen:
Ensure there's no call to `vim.schedule/vim.fn/vim.api/vim.rpcrequest` in `fn_transform/fn_xx` then it will work.

I think this is related:
```
                                                        *api-fast*
Most API functions are "deferred": they are queued on the main loop and
processed sequentially with normal input.  So if the editor is waiting for
user input in a "modal" fashion (e.g. the |hit-enter-prompt|), the request
will block.  Non-deferred (fast) functions such as |nvim_get_mode()| and
|nvim_input()| are served immediately (i.e. without waiting in the input
queue).  Lua code can use |vim.in_fast_event()| to detect a fast context.
```

ibhagwan:
```lua
require("fzf-lua").setup({
  keymap = {
    fzf = {
      true,
      load = function()
        print("called")
        _G._fzf_load_called = true
      end,
    }
  }
})
```
Strangely, the callbacks works with `FzfLua zoxide` but not with `FzfLua files`, still troubleshooting.

phanen:
`nvim -l a.lua`: `uv.run()` prevent `vim.schedule/fn` in this case:
```
vim.schedule(function() print(vim.fn.eval('&rtp')) end)
vim.uv.run()
```

I find I can make it quit by resize the screen...

ibhagwan:
Unsure yet why but the issue is with the workaround for #1914, removing the workaround makes it work:
https://github.com/ibhagwan/fzf-lua/blob/13ec251a963ed5645dc8ad49e0fde13dac42ec0f/tests/file/ui_spec.lua#L34-L38

ibhagwan:
I changed the wait condition at the end of spawn.lua to the below, it passes the load/postprocess condition but now fails randomly on clearing the create event:
```lua
-- while vim.uv.run() do end -- os.exit in spawn_stdio
while tonumber(vim.uv.os_getpriority(pid)) do
  vim.wait(100, function() return tonumber(vim.uv.os_getpriority(pid)) == nil end)
end
```

ibhagwan:
Getting some other weird issues where the cursor hangs on backspace after triggering the zero event (requires ctrl-c to unlock) - will also need to be looked at.

phanen:
Why we want to avoid `vim.wait(..)`... what's wrong with `while true do vim.wait(100) end`
`os.exit` can quit as expected

ibhagwan:
> Why we want to avoid `vim.wait(..)`... what's wrong with `while true do vim.wait(100) end` `os.exit` can quit as expected

Idk, waiting just seems like unsound practice, maybe I’m wrong.

> > Getting some other weird issues where the cursor hangs on backspace after triggering the zero event (requires ctrl-c to unlock) - will also need to be looked at.


I’m not sure, if this comes from the actions, the issue would be in shell_helper,  my guess would be another uv.run block.

ibhagwan:
> No I reproduce this without making change to shell,shell_helper.

I can’t reproduce this on main, most likely the issue comes from the changes in this PR.

lua/fzf-lua/shell_helper.lua
```diff
@@ -20,40 +24,42 @@ end
...
+
   preview_receive_socket:read_start(function(err, data)
     assert(not err)
     if not data then
       uv.close(preview_receive_socket)
       uv.close(preview_socket)
-      vim.schedule(function()
-        vim.cmd([[qall]])
-      end)
+      if _is_win then -- uv.stop() quit slow on windows
```
ibhagwan: Can probably remove this condition since we now always os.exit

lua/fzf-lua/shell_helper.lua
```diff
@@ -105,10 +91,18 @@ local function rpc_nvim_exec_lua(opts)
 
   if not success then
     io.stderr:write(("FzfLua Error: %s\n"):format(errmsg or "<null>"))
-    vim.cmd([[qall]])
+    os.exit(1)
   end
+
+  uv.run() -- return when uv.stop()
```
ibhagwan: Since we no longer call uv.stop and use os.exit should we move the channel close as well?

phanen: We cannot close since `fn.chanclose` is not allowed when `uv.run`

ibhagwan: What I’m saying is since we don’t uv.stop do we even need the uv.run?

phanen: Using `vim.wait`?

I find this https://github.com/ibhagwan/fzf-lua/pull/1955/commits/7e2873e8ba481ffbfda2b517cc23b9e4a2543b0f  can fix ci

ibhagwan: > Using `vim.wait`?
> 
> I find this [7e2873e](https://github.com/ibhagwan/fzf-lua/commit/7e2873e8ba481ffbfda2b517cc23b9e4a2543b0f) can fix ci

Nice find!

do you think it’s worth doing the same in spawn.lua for consistency?

phanen: No. This hack don't apply to spawn.lua I don't know why.

Here the first `uv.run("whatever")` will always return quickly (due to `chanclose`?). The second `uv.run` will be expected `noreturn` and finally quit by `os.exit`.

lua/fzf-lua/libuv.lua
```diff
@@ -218,6 +233,7 @@ M.spawn = function(opts, fn_transform, fn_done)
     output_pipe:read_start(read_cb)
     error_pipe:read_start(err_cb)
   end
+  return pid
```
ibhagwan: Shall we change this to `return handle, pid` so the return values match the `uv.spawn` return values?

phanen: make sense

ibhagwan:
Are you still working on this one? I feel we’re close to this being ready.

One thing we’ll have to add is up the dependency to neovim 0.8/0.9,I think `-l` was added with 0.8 I’ll have to check, we can also get rid of more backward compact code once we do that.

phanen:
`fs.normalize` is `v0.8.0`  things. But already exist in codebases..

ibhagwan:
> `fs.normalize` is `v0.8.0` things. But already exist in codebases.. It seems `v0.9.0+`... [neovim/neovim@7c94bcd](https://github.com/neovim/neovim/commit/7c94bcd2d77e2e54b8836ab8325460a367b79eae)
> 
> Latest `v0.8.3` don't have it, but here a same name option existing `v0.1.0 ~ v0.8.3`:
> 
> ```
> 							*-l*
> -l		Lisp mode.  Sets the 'lisp' and 'showmatch' options on.
> ```

I think we can require 0.9, it’s been long enough, this should simplify things and we can remove quite a few `if utils.__HAS…`, especially the annoying one for the highlight bug in 0.8 lol
https://github.com/ibhagwan/fzf-lua/blob/47b85a25c0c0b2c20b4e75199ed01bb71e7814f5/lua/fzf-lua/init.lua#L149-L161

ibhagwan:
> * https://repology.org/project/neovim/versions ubuntu 22.04 is v0.6.1 (EOL is 2027)
> * Current min support version is v0.7.0? [aaf7604](https://github.com/ibhagwan/fzf-lua/commit/aaf7604bca38890cb487b29738f8cb02586bc818).
> 
> If merge this that's would be `v0.7.0`(which work well with builtin previewer) -> `v0.9.0`... although I start using neovim since `v0.10.0`... I'm not sure how many user will complain...

I think we can tag latest version to work with 0.7 and require 0.9, anyone using neovim at this point is most likely using 0.10+, almost no plugins support 0.7/0.8, lspconfig/treesitter all require 0.9, if someone has an old config and wants to update fzf-lua they can use the `tag = …`.

phanen:
> I think we can tag latest commit to work with 0.7 and require 0.9, anyone using neovim at this point is most likely using 0.10+,

Make sense. I never see a ancient user for neovim, someone use old version but just keep compatibale config. And since neovim never backport thing to unsupported release. So I think ideally nothing will break if stick to a tag.

I found fzf-lua have always been a `0.0.xx` release package: https://github.com/folke/lazy.nvim/blob/b6eba0d02600b678959f90876bc678178b4dd4f6/manifest#L23236-L24903. We should be listed in https://0ver.org/ (:

ibhagwan:
> I found fzf-lua have always been a 0.0.xx release package:

luarocks package is automatically versioned based on the number of commits:
https://github.com/ibhagwan/fzf-lua/blob/main/.github/workflows/luarocks-release.yaml

ibhagwan:
> We should be listed in https://0ver.org/ (:

We can consider a versioning system, I never bothered with it as I feel the project doesn’t need it, will just create an overhead to “release a version”.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/d82ead472fc4fd72bafd1d506e1a9f6d17dcbbfc, let me know if you think I missed anything, `fzf-lua.txt` should be updated by the testing to use 0.11.0 as "stable" but I didn't want to do this now as it would introduce too much noise updating the screenshots so it's better to do in another PR.

phanen:
> We can consider a versioning system, I never bothered with it as I feel the project doesn’t need it, will just create an overhead to “release a version”.

fzf-lua usually fix issues quickly on HEAD. Tag with minimal support neovim version should be enough.

ibhagwan:
I’ve been using this PR for two days now, seems stable at least on my Linux, still unsure when we should merge this, do you want to fix the Windows CI first or remove the fixup commit?

These type of changes usually have some after effect we never think about so we’ll have to be more alert in case something breaks.

phanen:
We can merge this first since gc fixes are unrelated here. gc actually work in my win10 vm when use interactively but fail in test.

I cherry pick those fixes to another PR. (that's a blocker before changing the default profile to "hide")

ibhagwan:
As always, tysm @phanen for your amazing work on this :)

PR #1961, opened 2025-04-08
fix(hide): selection is empty when do action on "multi-selected && zero-matched"

phanen:
* Regression of
https://github.com/ibhagwan/fzf-lua/commit/6e61cef02ed41059b4ae3baac791554e1b26f13a:
`not tonumber(idx)` imply list is empty, but select may not empty.
* Don't patch `"" -> nil` when user have it own `v.field_index` which is
unpredictable.
* If user set `field_index = "{q} {n} xxx"`, we shouldn't remove it.

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -1129,6 +1129,45 @@ M.set_header = function(opts, hdr_tbl)
   return opts
 end
 
+
+-- Use both {q} and {+} as field indexes so we can update last query when
+-- executing the action, without this we lose the last query on "hide" as
+-- the process never terminates and `--print-query` isn't being printed
+-- When no entry selected (with {q} {+}), {+} will be forced expand to ''
+-- Use {n} to know if we really select an empty string, or there's just no selected
+local patch_shell_action = function(v, opts)
+  local field_index = v.field_index == false and "" or v.field_index or "{+}"
+  local overide_f_idx
+  if not field_index:match("^{q} {n}") then
+    field_index = "{q} {n} " .. field_index
+    overide_f_idx = true
+  end
+  -- replace the action with shell cmd proxy to the original action
+  return shell.raw_action(function(items, _, _)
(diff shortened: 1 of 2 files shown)


ibhagwan:
Looks good @phanen, I also wanted to dedup the action patch but was lazy about it lol

Do you feel this is ready to be merged? Are you still testing the new CI with hide?

ibhagwan:
Ty @phanen for your amazing work!

phanen:
> On another note: how do you feel setting the default profile (if none specified) to `{"border-fused","hide"}`?

"hide" have been a better default. But I think "border-fused" need more border type (now "single" border only, and it's a bit tedious to config it by user own.

> due to `command too long`

Maybe it's still long if you pass base64 encoding when use `multiprocess`...
And that's depends on number of actions user provided.

ibhagwan:
> Maybe it's still long if you pass base64 encoding when use multiprocess...
And that's depends on number of actions user provided.


Correct, that’s why the issue is more prevalent the more actions we have as each action key bind turns into a long execute-silent command.


> But with https://github.com/ibhagwan/fzf-lua/pull/1955 it should be not much longer than in unix? I'm not sure what's the longest support command on windows v.s. unix.

Don’t think we’ll ever reach the Unix max, on Windows the limitation comes from cmd.exe, I didn’t test but I think it’s around 2048 chars? Or maybe less because I don’t think it was that long…

Perhaps we can work around it by using powershell.exe instead of cmd.exe as the wrapper but that has 2 drawbacks:
- Slower startup than cmd.exe (which is already slow as it is)
- Requires major refactor for which I don’t have time and no longer have access to the Windows dev VM I was loaned


> "hide" have been a better default

We should do the change then, but maybe limit it to non-Windows?

ibhagwan:
> Or I'm even not sure if "hide" really working on windows...

It works, I tested it when I had the Win VM.

Issue #1974, opened 2025-04-15
Bug: [sk] with fzf_bin = 'sk', preview not rendering properly in tmux


levihuayuzhang:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Mac

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.0 Build type: Release LuaJIT 2.1.1744318430

### Fzf version (`fzf --version`)

0.61.1 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup {
          fzf_bin = 'sk',

          -- fzf-native
          winopts = { preview = { default = 'bat' } },
          manpages = { previewer = 'man_native' },
          helptags = { previewer = 'help_native' },
          lsp = { code_actions = { previewer = 'codeaction_native' } },
          tags = { previewer = 'bat' },
          btags = { previewer = 'bat' },

          -- opens in a tmux popup (requires tmux > 3.2)
          fzf_opts = { ['--border'] = 'rounded', ['--tmux'] = 'center,80%,60%' },
        }
```

### Describe the bug / steps to reproduce

with `fzf_bin = 'sk'` (using skim).

preview not rendering properly in `tmux` (but ok without `tmux`, or says, outside `tmux`).

In `tmux`, preview panel shows: 
1. `argv[2] = nil`
2. `FzfLua Error: VimE474: Invalid argument.` (if using `bat`)
3. `cat: No such file or directory` (if using `cat`)

It seems there may have any default arguments were incompatible with `skim` or `skim-tmux`?

phanen:
I don't know why `sk --tmux` don't pass environment variable:
```
AA=1 sk --tmux --bind 'ctrl-a:execute-silent:notify-send $AA'
```

phanen:
Ok. I find a workaounrd. In https://github.com/ibhagwan/fzf-lua/pull/1975.

levihuayuzhang:
[...]

PR #1975, opened 2025-04-15
fix: `skim --tmux` didn't pass all environemnt variable

phanen:
~I search through skim docuemnt `$SKIM` seems for testing purpose only so we can use it.
Otherwise we need to pass `fzf_lua_server` in cli again...~ Ok, there's no $SKIM, it's $SKIM_xx

Since `SKIM_xx` is allowed we use `SKIM_FZF_LUA_SERVER`...

Fix https://github.com/ibhagwan/fzf-lua/pull/1974.

fzf wrap a proxy bash script with tons of lines of `export xx=xx` (`ps aux | rg fzf-temp | rg '\-\-tmux'`)...https://github.com/junegunn/fzf/blob/aefb9a5bc41f92227e4bffa050caca0270b450ba/src/proxy.go#L63.
But this `--proxy-script` is an undocumented option (at least currently)

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -280,6 +280,8 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
       ["FZF_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
       ["SKIM_DEFAULT_COMMAND"] = FZF_DEFAULT_COMMAND,
       ["FZF_LUA_SERVER"] = vim.g.fzf_lua_server,
+      -- sk --tmux didn't pass all environemnt variable (https://github.com/skim-rs/skim/issues/732)
+      ["SKIM_FZF_LUA_SERVER"] = vim.g.fzf_lua_server,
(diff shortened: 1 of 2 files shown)


phanen:
Yes. Actually prefix with `SKIM`. https://github.com/lotabout/skim/blob/5698c6d8fd29096539a269ac1ed9902cd2f4b982/skim/src/tmux.rs#L183-L188.

ibhagwan:
> > Ok, there's no $SKIM, it's $SKIM_xx
> 
> I'm not sure why I said it didn't work... Maybe I checkout from old branch to test it at that time.
> 
> > vim.env.FZF_LUA_SERVER or vim.env.SKIM_FZF_LUA_SERVER so the variable needed to be be prefixed with SKIM_ to be exported?
> 
> Yes. Actually prefix with `SKIM`. https://github.com/lotabout/skim/blob/5698c6d8fd29096539a269ac1ed9902cd2f4b982/skim/src/tmux.rs#L183-L188.

Great find! Ty @phanen!

PR #1983, opened 2025-04-16
feat: custom treesitter line parser

phanen:
Fix https://github.com/ibhagwan/fzf-lua/discussions/1982.

Maybe need a better option name.


ibhagwan:
> Maybe need a better option name.

This is very internal, I don’t expect this to ever be used outside of #1983.

Issue #1989, opened 2025-04-18
Bug: Not returning back to insert mode when starting an action from insert mode


otomn:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

WSL

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.10.4

### Fzf version (`fzf --version`)

0.60.3 (0012183e)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=fg:-1,fg+:underline,fg+:-1 --color=spinner:31,pointer:167,info:136,prompt:31,query:regular,hl+:167,hl:167,marker:31,border:238,header:247 --color=bg:-1,bg+:-1   

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({})
vim.keymap.set('i', '<c-\\>z', fzf.spell_suggest, {})
```

### Describe the bug / steps to reproduce

When I invoke fzf function in insert mode and then return, the mode is always reset to normal mode. This is particularly troublesome when I use it for spell correct.
I am interested in any workaround if possible.
Still reproducible on 40fbbb6

phanen:
Try https://github.com/ibhagwan/fzf-lua/pull/1991.

But "<cword>" seems not get current word currently if you're in the end of a word. I don't know if there any option can change the behavior.

phanen:
BTW, you can also use `<c-x><c-s>` with builtin completion.

phanen:
Let's hope one day nvim expose an `ext_completion`, then we only need to do ui work.

otomn:
Thank you sir. You are so fast. This is the fastest bug fix and merge I have ever seen in any public repo.

ibhagwan:
> Thank you sir. You are so fast. This is the fastest bug fix and merge I have ever seen in any public repo.

Ty @phanen!

otomn:
> Try [#1991](https://github.com/ibhagwan/fzf-lua/pull/1991).
> 
[...]

PR #1992, opened 2025-04-18
fix(actions): reuse actions.complete for spell

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1989#issuecomment-2814514886.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -1086,7 +1086,7 @@ M.defaults.spell_suggest        = {
     width    = 0.30,
   },
   actions = {
-    ["enter"] = actions.spell_apply,
+    ["enter"] = actions.complete,
   },
 }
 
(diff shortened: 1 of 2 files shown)


phanen:
To control word boundary we need to use `opts.word_pattern` instead of `vim.o.iskeyword` now.

ibhagwan:
Ty @phanen, great implementation for compete.

phanen:
No... I just find it didn't fix https://github.com/ibhagwan/fzf-lua/issues/1989#issuecomment-2814514886.

ibhagwan:
> No... I just find it didn't fix [#1989 (comment)](https://github.com/ibhagwan/fzf-lua/issues/1989#issuecomment-2814514886).
> 
> `fzf-lua.complete_xx` won't resume to insert mode if `ctrl-c`

That makes sense, ctrl-c stops everything, even on my regular completion with blink/cmp, escape closes completion but ctrl-c aborts everything and goes into normal mode, this behavior is consistent with neovim and shouldn’t be fixed, it’s more of a “feature”.

PR #1993, opened 2025-04-18
fix(actions): resume insertmode when complete abort

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/1989#issuecomment-2814514886 again.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -128,7 +128,11 @@ M.act = function(selected, opts)
 end
 
 -- Dummy abort action for `esc|ctrl-c|ctrl-q`
-M.dummy_abort = function()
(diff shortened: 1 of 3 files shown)


ibhagwan:
As I wrote in https://github.com/ibhagwan/fzf-lua/pull/1992#issuecomment-2814949947, I’m not sure if we want this fixed, it makes sense for ctrl-c to go back to normal mode no?

phanen:
Yeah. But if we start picker in an insert mode (for completion). I feel we should resume it somehow.

phanen:
* resume insert mode in `dummy_abort`.
* `ctrl-c` will still resume normalmode
* fix wrong suggestion when cursor in the end of word

ibhagwan:
Ty @phanen for the very thoughtful PR.

PR #1998, opened 2025-04-19
fix(zoxide): sort by score

phanen:
Fix https://github.com/ibhagwan/fzf-lua/discussions/1928.

Side note:
So the way to implement a `frecency` fzf-lua plugin is `cmd | frecency-sort | fzf --no-sort`

Edit:
That's not very good though... since fzf have to wait frecency-sort finish all sort job... 

thing would be much more easier as long as fzf can give a sort rank on something like prefix score number. then also we can just use spawn.lua to prepend a score retrieval from whatever db while processing the stream.


ibhagwan:
Ty @phanen!

PR #2003, opened 2025-04-21
fix: VimResized when fzf buffer hidden

phanen:
`FzfLua files` -> `hide` -> resize terminal -> `unhide` -> broken layout


ibhagwan:
Ty @phanen!

PR #2004, opened 2025-04-22
docs: fix title_fnamemodify example

carloscalla:
Adds missing `return` statement in docs for the `title_fnamemodify` example.
While testing I kept getting and error regarding a `nil` value so I noticed the `return` statement was missing.

Error without the `return` statement:
image


ibhagwan:
You’re absolutely correct, ty @carloscalla!

PR #2005, opened 2025-04-23
perf(previewer): don't update if entry is the same as `last_entry`

phanen:
This is the root cause of https://github.com/ibhagwan/fzf-lua/pull/1913#issuecomment-2810013998.


ibhagwan:
Ty @phanen!

ibhagwan:
Had to revert this commit, causes issue with F4 toggle, also unhide shows empty preview, technically this should take care of it:
https://github.com/ibhagwan/fzf-lua/blob/23dce7340eecc74cfd933401e1690b8416d7ff86/lua/fzf-lua/previewer/builtin.lua#L571-L574

PR #2006, opened 2025-04-23
ci: remove uncessary attr screenshots

phanen:
Those lines looks redundant, can be removed.

EDIT: weird. test passed locally


ibhagwan:
This is the reasoning, mini.test currently needs this for the comparison to work:
https://github.com/echasnovski/mini.nvim/issues/1300#issuecomment-2614542586

phanen:
`screenshot_read` also don't work as my expectation now...
https://github.com/echasnovski/mini.nvim/blob/687fa868d869078194c02065628712982d971fd6/lua/mini/test.lua#L2307-L2320

Ok, we can wait for upstream support it.

PR #2007, opened 2025-04-23
perf(previewer): don't update if entry is the same as `last_entry` (2nd try)

phanen:
Fix https://github.com/ibhagwan/fzf-lua/pull/1913#issuecomment-2810013998 again.


ibhagwan:
Ty @phanen!

Issue #2008, opened 2025-04-23
Bug: snacks.image not loaded


carloscalla:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

MacOs 15.4

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.0

### Fzf version (`fzf --version`)

0.61.3 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
  previewers = {
    builtin = {
      snacks_image = { enabled = true }
    }
  }
})
```

### Describe the bug / steps to reproduce

When running `:FzfLua files`, fzf-lua doesn't get the `snacks.image` module to load the images yet, even when set explicitly as `enabled = true` in the options. The picker would show this:

![Image](https://github.com/user-attachments/assets/2c66606c-ac5a-4a3d-8dc0-c78c337ca399)

I have debugged this and seems like `snacks.image` module is lazy loaded, even when explicitly setting the `snacks.image` module as `{enabled = true}` in the snacks config.

I noticed that the code checks for the `snacks.image` module to be loaded but it's always false because it's always lazy loaded:

![Image](https://github.com/user-attachments/assets/5f069576-0b09-4624-8189-60989bd59244)

I've managed to solve this in my config by calling `require 'snacks.image'` before fzf-lua setup:

```lua
    require 'snacks.image'
    require('fzf-lua').setup {}
```

This solves the issue for me.

Note: when I run `:FzfLua files` the images don't load, then run the snacks files picker and the images load (it calls the `snacks.image` under the hood), after this I then run again `:FzfLua files` and the images load. That's how I noticed the module is always lazy loaded. I tried via the configs to try to load the `snacks.image` module but had no success so my solution for now was to require it before the fzf-lua setup call as shown above.

[...]

PR #2009, opened 2025-04-23
Manually load snacks.image module while checking if available

carloscalla:
Fixes https://github.com/ibhagwan/fzf-lua/issues/2008


ibhagwan:
Ty @carloscalla!

PR #2011, opened 2025-04-24
fix: `:quit` on other window should not kill fzf job

phanen:
`FzfLua files` -> `<esc>` -> `new` -> `quit` -> `unhide` fail


ibhagwan:
Ty @phanen!

ibhagwan:
Unrelated but I wonder why the macOS CI keeps failing on waiting for condition even after I increased the timeout to 5s, macOS shouldn’t be slower than Windows…

PR #2013, opened 2025-04-26
fix(preview): `helptags()` cannot preview markdown help buffer

phanen:
When use `lazy.nvim`, it will auto generate markdown tags for those plugin have no vimdoc.

e.g. `nvim-bqf`
```vim
:FzfLua helptags query=nvim-bqf winopts.preview.hidden=false
```


ibhagwan:
Ty @phanen!

PR #2014, opened 2025-04-26
fix(actions): when clipboard=unnamed,unnamedplus

phanen:
* when clipboard=unnamed,unnamedplus
* Also some nil check when no selected.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -448,6 +448,7 @@ M.arg_del = function(selected, opts)
 end
 
 M.colorscheme = function(selected, opts)
+  if #selected == 0 then return end
   local dbkey, idx = selected[1]:match("^(.-):(%d+):")
   if dbkey then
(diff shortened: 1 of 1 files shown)


lua/fzf-lua/actions.lua
```diff
@@ -57,7 +57,7 @@ M.normalize_selected = function(selected, opts)
   -- The below separates the keybind from the item(s)
   -- and makes sure 'selected' contains only item(s) or {}
   -- so it can always be enumerated safely
-  if not selected then return end
+  if not selected or #selected == 0 then return end
```
ibhagwan: I don’t think we should add this, not all actions aee dependent on the items, for example say you press enter on an action after typing a query that yields no matches, for example, git_branch_add with skim, `selected` will contain no items and the branch name will be taken from `opts.last_query`.

I didn’t test so maybe I’m wrong but even dummy_abort won’t be called if you type a query long enough and there would be no matches.

ibhagwan:
Ty @phanen!

PR #2021, opened 2025-05-02
updated health check

mainendra:
Added `mini.icons` check along with `nvim-web-devicons`

### DIFF
--- a/lua/fzf-lua/_health.lua
+++ b/lua/fzf-lua/_health.lua
@@ -82,9 +82,9 @@ function M.check()
   end
 
   start("fzf-lua [optional]")
-  if pcall(require, "nvim-web-devicons") then
+  if package.loaded["nvim-web-devicons"] then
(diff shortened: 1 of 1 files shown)


lua/fzf-lua/_health.lua
```diff
@@ -84,8 +84,10 @@ function M.check()
   start("fzf-lua [optional]")
   if pcall(require, "nvim-web-devicons") then
     ok("`nvim-web-devicons` found")
+  elseif pcall(require, "mini.icons") then
```
ibhagwan: This isn’t the right way to check if mini.icons is loaded as this will cause lazy loading of the module from the check health, a better way to check is test for `package.loaded["mini.icons"]` (same can be applied to the nvim-web-devicons test).

mainendra: Done.

ibhagwan:
Ty @mainendra!

PR #2028, opened 2025-05-04
feat: show diff with any branch

mostafaqanbaryan:
I've added a new command: `git_diff` that works with a branch name.

I needed this for my work, because I don't remember what files I've changed in the past on a single branch.

So basically, this is only for me!

If you want to merge it, I'm open to suggestions.

_(I don't have enough experience with this code base, and just edited the `git_status` command to make it work.)_

### DIFF
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -92,57 +92,6 @@ M.status = function(opts)
   return core.fzf_exec(contents, opts)
 end
 
-M.diff = function(opts)
-  opts = config.normalize_opts(opts, "git.diff")
-  if not opts then
-    return
-  end
-  opts.cmd     = "git show-ref --quiet --branches " .. opts.branch .. " && git diff " .. opts.branch .. " --name-only || echo Branch " .. opts.branch .. " not found"
-  opts.preview = "git show-ref --quiet --branches " .. opts.branch .. " && git diff " .. opts.branch .. " --color {2}"
-  opts = set_git_cwd_args(opts)
-  if not opts.cwd then
-    return
-  end
-  if opts.preview then
-    opts.preview = path.git_cwd(opts.preview, opts)
-  end
-
-  -- we always require processing (can't send the raw command to fzf)
-  opts.requires_processing = true
-
-  local contents, id
-  if opts.multiprocess then
-    opts.__mt_transform = [[return require("fzf-lua.make_entry").file]]
-    contents = core.mt_cmd_wrapper(opts)
-  else
-    opts.__fn_transform = opts.__fn_transform or function(x)
-      return make_entry.file(x, opts)
-    end
-
-    -- we are reusing the "live" reload action, this gets called once
-    -- on init and every reload and should return the command we wish
-    -- to execute, i.e. `git diff`
-    opts.__fn_reload = function(_)
-      return opts.cmd
-    end
-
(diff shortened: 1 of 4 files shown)


lua/fzf-lua/providers/git.lua
```diff
@@ -92,6 +92,57 @@ M.status = function(opts)
   return core.fzf_exec(contents, opts)
 end
 
+M.diff = function(opts)
+  opts = config.normalize_opts(opts, "git.diff")
+  if not opts then
+    return
+  end
+  opts.cmd     = "git show-ref --quiet --branches " .. opts.branch .. " && git diff " .. opts.branch .. " --name-only || echo Branch " .. opts.branch .. " not found"
+  opts.preview = "git show-ref --quiet --branches " .. opts.branch .. " && git diff " .. opts.branch .. " --color {2}"
```
nenahp: Maybe this can also be implemented by passing these args to git_status

mostafaqanbaryan: In this case, `preview_pager` should be changed as well.

But we need the functionality of `git_status` and don't want to lose that.

ibhagwan:
Ty @mostafaqanbaryan!

> I've added a new command: git_diff that works with a branch name.

This is going to need a different name, I was planning on adding `git diff` which outputs all currently modified lines so we’re able t9 jump to the exact line.

Maybe `git_log_branch` or `git_status_branch` (may or may not be applicable as I haven tested to see what this does exactly).

> (I don't have enough experience with this code base, and just edited the git_status command to make it work.)

Again, haven’t looked into the code deeply yet, but is there a way to make `git_status` configurable / accept a branch param so the code isn’t duplicated?

ibhagwan:
Ty @nenahp, might take me a couple of days to take a look at this properly as I’m traveling for work.

ibhagwan:
Can this be combined to a  diff for any ref? can this be used to compare to a different commit (to HEAD or a different commit/tag)?

We can then name this new picker `git_diff_ref` so we can reserve the `git_diff` name to the current changes diff picker (for changed lines)?

mostafaqanbaryan:
> Can this be combined to a diff for any ref? can this be used to compare to a different commit (to HEAD or a different commit/tag)?
> 
> We can then name this new picker `git_diff_ref` so we can reserve the `git_diff` name to the current changes diff picker (for changed lines)?

Yes, instead of `HEAD`, we could put any branch/hash that we want. So I guess it's possible to change the name to `git_diff_ref`.

ibhagwan:
Ty @mostafaqanbaryan, merged, applied my 2 cents on the PR and fixed the previewer (wouldn't show some diffs, now matches the command), also added ref check at the start with proper warning.

ibhagwan:
Btw, didn't rename from `git_diff` as I'll name the new picker `git_hunks`

ibhagwan:
> fzf_lua.git_status can be aliased to git_diff { ref = "HEAD" } now.

`git_diff` doesn't include the stage/unstage markers and actions.

I was thinking to add these markers to the new `git_hunks` + the ability to stage/unstage hunks but it's too much work, might as well use git signs or mini.diff for that (which most users have installed).

mostafaqanbaryan:
> Ty @mostafaqanbaryan, merged, applied my 2 cents on the PR and fixed the previewer (wouldn't show some diffs, now matches the command), also added ref check at the start with proper warning.
> 
> ```lua
> :FzfLua git_diff ref=main
> ```

That's great, thank you!

PR #2033, opened 2025-05-06
ci: remove uncessary attr screenshots (2nd try)

phanen:
I realized we can access hidden mini.test helper module by `debug.getupvalue`.


ibhagwan:
Wow looks great @phanen!

PR #2036, opened 2025-05-07
ci inline screenshot

phanen:
- **ci: can assert screenshot in lua file**
- **ci: fix test**
- **ci: add case filter**


ibhagwan:
Looks good, would you like me to merge this or need more time?

PR #2039, opened 2025-05-10
expose hls.path_filename

otomn:


ibhagwan:
Ty @otomn.

I don’t think this is needed, we already have `FzfLuaFilePart` and use a formatter, e.g.:
```vim
:hi! link FzfLuaFilePart ErrorMsg
:FzfLua files formatter=path.dirname_first
``` 

This applies everywhere paths are used not just buffers, quick fix, etc:
IMG_3216

otomn:
@ibhagwan I added it because I noticed filename isn't highlighted for e.g. any of the LSP functions.
Output of 
```vim
:hi! link FzfLuaFilePart ErrorMsg
:Fzf lsp_references
```
Screenshot 2025-05-10 at 21 49 07

ibhagwan:
As I wrote in my comment, this requires a formatter:
IMG_3217

lua/fzf-lua/make_entry.lua
```diff
@@ -304,7 +306,7 @@ M.lcol = function(entry, opts)
   return string.format("%s:%s%s%s",
     -- uncomment to test URIs
     -- "file://" .. filename,
-    filename, --utils.ansi_codes.magenta(filename),
```
otomn, line 307: Looking at this code here. You can clearly see that filename had no highlight configurable at all.

otomn:
Ok, I guess having a custom formatter would work. But it feels inconsistent to me that there's no highlight group for the filename part but there is for line nr and col nr.

ibhagwan:
> Ok, I guess having a custom formatter would work. But it feels inconsistent to me that there's no native highlight group for the filename part but there is for line nr and col nr.

It would also be inconsistent to have the file part work only for some pickers (not files for example), the reason is some pickers are lua table/callback based and some are shell command output (rg/fd/etc) - having to parse the path requires extra processing (=lower perf) so it’s not enabled by default but it is consistent access all pickers that have file paths.

You can enable the formatter everywhere by default if you wish (which I do, via `defaults’):
https://github.com/ibhagwan/nvim-lua/blob/42931f61aa11629be3e6907340be9109857038e9/lua/plugins/fzf-lua/setup.lua#L167

lua/fzf-lua/make_entry.lua
ibhagwan: That was ages ago before the highlights were introduced.

ibhagwan:
Closing as we already have `FzfLuaFilePart`.

Issue #1834, opened 2025-02-17
Feature: Files with line-number


wikiliu:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

Whether fzf-lua support find file with line-number and show/goto the line directly？Just like telescope？

![Image](https://github.com/user-attachments/assets/6c3112b6-2c98-4d04-9e11-d9443f3671a2)
|
|

![Image](https://github.com/user-attachments/assets/0f4d18f4-4735-4202-abe2-170fec141862)

phanen:
If you don't care about performance cost, `fzf_live` can help you:
```lua
  local path = require('fzf-lua.path')
  path.entry_to_file = (function(overriden)
    return function(...)
      local res = overriden(...)
      res.line = g.lnum or 0
      return res
    end
  end)(path.entry_to_file)
  require('fzf-lua').fzf_live(function(q)
    g.lnum = q:match('^.*:(.*)$')
    local query = q:match('^[^:]*')
    return 'fd --type f --hidden --follow -E .git | rg --color=always ' .. query
  end, { previewer = 'builtin' })
```

ibhagwan:
@wikiliu, this has been asked numerous times, unfortunately, ATM this isn’t straight forward with fzf so any solution will be super hacky.

We can revisit this in the future when fzf adds features that would help this use case.

ibhagwan:
FYI, fzf 0.59 made this possible, update to https://github.com/ibhagwan/fzf-lua/commit/8cdf9aaf1c07ff5820b51ef7432ca45ab07ea982 and use (works with any file-type picker):

```lua
:FzfLua files line_query=true
```

Once `:<line>` is detected it will be parsed and applied to both the previewer and action (enter, etc):
![image](https://github.com/user-attachments/assets/281a309f-8d82-4a73-9449-3ee0675d5dfc)

Aiyane:
@ibhagwan  Thank you very much! However, the `line_query=true` configuration is not yet supported in the `bat_native` previewer.

phanen:
Fixed in https://github.com/ibhagwan/fzf-lua/pull/2045.

PR #2045, opened 2025-05-14
fix(previewer): opts.line_query in fzf previewer

phanen:
https://github.com/ibhagwan/fzf-lua/issues/1834#issuecomment-2878875713

I delete `result` event since it seems unused?

### DIFF
--- a/lua/fzf-lua/previewer/fzf.lua
+++ b/lua/fzf-lua/previewer/fzf.lua
@@ -27,10 +27,7 @@ function Previewer.base:preview_window(_)
   return nil
 end
 
-function Previewer.base:_preview_offset()
-  if self.opts.preview_offset or self.preview_offset then
-    return self.opts.preview_offset or self.preview_offset
-  end
+function Previewer.base:_preview_offset(lnum)
   --[[
     #
     #   Explanation of the fzf preview offset options:
@@ -42,8 +39,9 @@ function Previewer.base:_preview_offset()
     #
     '--preview-window '~3:+{2}+3/2''
   ]]
-  if self.opts.line_field_index then
-    return ("+%s-/2"):format(self.opts.line_field_index)
+  lnum = lnum or self.opts.line_field_index
+  if lnum then
+    return ("+%s-/2"):format(lnum)
   end
 end
 
@@ -224,9 +222,6 @@ Previewer.bat_async = Previewer.cmd_async:extend()
 
 ---@param lnum string?
 function Previewer.bat_async:_preview_offset(lnum)
-  if self.opts.preview_offset or self.preview_offset then
-    return self.opts.preview_offset or self.preview_offset
-  end
   --[[
     #
     #   Explanation of the fzf preview offset options:
--- a/lua/fzf-lua/previewer/fzf.lua
+++ b/lua/fzf-lua/previewer/fzf.lua
@@ -178,6 +178,7 @@ end
 
 function Previewer.cmd_async:parse_entry_and_verify(entrystr)
   local entry = path.entry_to_file(entrystr, self.opts)
+  entry.line = self.opts.line_query and tonumber(self._last_query:match(":(%d+)$")) or entry.line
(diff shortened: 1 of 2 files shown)


lua/fzf-lua/previewer/fzf.lua
```diff
@@ -260,9 +262,10 @@ end
 function Previewer.bat_async:cmdline(o)
   o = o or {}
   local act = shell.raw_preview_action_cmd(function(items, fzf_lines)
+    self._last_query = items[2] or ""
     local filepath, entry, errcmd = self:parse_entry_and_verify(items[1])
     local line_range = ""
-    if entry.ctag then
+    if entry.line then
```
Aiyane: This line would cause non-ctag previewer to apply the `--line-range` configuration, leading to unexpected rendering in the custom previewer I defined.

phanen: You're right. This also disable scroll (which is done by fzf term, but currently it's implemented by static `--preview-window`).

So maybe need some actions to apply scroll on `change` event.

phanen: @Aiyane Force pushed, could you test if it work now.

ibhagwan:
> I delete result event since it seems unused?

If I recall correctly I added `result` is needed to update the line once the fuzzy filtering is done.

I’ll do some testing on the rest of the changes.

ibhagwan:
On another note, I don’t understand this MacOS CI, I increased the condition timeout to 10s and it still fails, mostly only on “macos-latest” and only on the push run, if I re-run the failed job it mostly succeeds…

phanen:
> On another note, I don’t understand this MacOS CI, I increased the condition timeout to 10s and it still fails, mostly only on “macos-latest” and only on the push run, if I re-run the failed job it mostly succeeds…

This seems related to a commit on nightly. Maybe already existed in `v0.11.1`...

ibhagwan:
> > On another note, I don’t understand this MacOS CI, I increased the condition timeout to 10s and it still fails, mostly only on “macos-latest” and only on the push run, if I re-run the failed job it mostly succeeds…
> 
> This seems related to a commit on nightly. Maybe already existed in `v0.11.1`...

Why only on Mac though…

phanen:
The oldest failure: https://github.com/ibhagwan/fzf-lua/actions/runs/14623427422

This one looks suspicious:
https://github.com/neovim/neovim/commit/82f08f33c1017408940d5d03c19555364a8029d5

lua/fzf-lua/previewer/fzf.lua
Aiyane: @phanen It works well when the previewer is configured as `bat`, but when it's set to `bat_native`, the previewer doesn't jump to the correct line. They don't seem to be using the same previewer, so it looks like both the  `bat_async` and `bat` previewers need to be modified.

ibhagwan: > @phanen It works well when the previewer is configured as `bat`, but when it's set to `bat_native`, the previewer doesn't jump to the correct line. They don't seem to be using the same previewer, so it looks like both the `bat_async` and `bat` previewers need to be modified.

I don’t think bat_native can be fixed easily nor would I worry about it, I’d like to remove it and leave only the “bat” (async) previewer, there is no real benefit to the bat_native over regular bat anyways.

ibhagwan:
> This one looks suspicious:
https://github.com/neovim/neovim/commit/82f08f33c1017408940d5d03c19555364a8029d5

Interesting, maybe this causes an issue with the headless instance? I’m not sure, also, why would this happen only on Mac and why just the first CI?

ibhagwan:
@Aiyane, merged and fixed limited support for bat_native (because why not lol), the line will jump to the offset but there will be no line highlight as this is determined by the original preview command sent to fzf (can be fixed by as I mentioned I plan to delete bat_native).

PR #2049, opened 2025-05-16
tabs.locate positions cursor at current window

lbrayner:
Hey, @ibhagwan , can we add the ability to position the cursor at the current window to the tabs provider?

Use fn_pre_fzf to set opts.__load_pos, and core.build_fzf_cli adds the action to fzf_opts["--bind"].

### DIFF
--- a/README.md
+++ b/README.md
@@ -1057,6 +1057,7 @@ previewers = {
     prompt            = 'Tabs❯ ',
     tab_title         = "Tab",
     tab_marker        = "<<",
+    locate            = true,         -- position cursor at current window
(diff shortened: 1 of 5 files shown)


ibhagwan:
Similar to https://github.com/ibhagwan/fzf-lua/discussions/1870.

Sure I don’t see harm in adding this option, I’ll text and merge, ty @lbrayner!

I do believe this has to be limited to an Fzf version using `utility.has` as this wouldn’t work with skim.

ibhagwan:
@lbrayner, removed the core binds hack, instead used `keymap.fzf.load` directly, also limited support for fzf 0.36 as there's no `load` event prior.

PR #2059, opened 2025-05-23
feat(defaults): treesitter highlighting for command history

phanen:
image


ibhagwan:
Great idea @phanen.
 
 I didn’t know vim ts would cover lua too even when = is used.

ibhagwan:
Mine looks a bit lame though…

IMG_3269

PR #2066, opened 2025-05-25
debug.getupvalue hacks on ts context win

phanen:
- **Revert "fix(tx-context): revert 1cc70fb29e63ff26acba1e0cbca04705f8a485f1."**
- **debug.getupvalues**

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -810,10 +810,9 @@ function Previewer.buffer_or_file:populate_preview_buf(entry_str)
     assert(cached.bufnr == self.preview_bufnr)
     -- same file/buffer as previous entry no need to reload content
     -- only call post to set cursor location
-    if type(self.cached_bufnrs[tostring(self.preview_bufnr)]) == "table"
-        and ((tonumber(entry.line) and entry.line > 0 and entry.line ~= self.orig_pos[1])
-          or (tonumber(entry.col) and entry.col > 0 and entry.col - 1 ~= self.orig_pos[2]))
-    then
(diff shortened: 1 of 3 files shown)


phanen:
Sad. I think this is the minimal efforts to get thing correct. Hopefully it won't break in short term. Even `getupvalue` is used I make silent fail if upstream make changes. 

We can adapt it from time to time or forking it if thing changes a lot. Or implement it in fzf-lua is also an option since it only have 1400 loc...we can just use queries file from that repo.  
```
~/l/nvim-treesitter-context master ❯ tokei lua/
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Lua                     6         1402          928          273          201
===============================================================================
 Total                   6         1402          928          273          201
===============================================================================
```

ibhagwan:
> Sad. I think this is the minimal efforts to get thing correct. Hopefully it won't break in short term. Even getupvalue is used I make silent fail if upstream make changes.

Maybe better to err if we fail so we know we need to fix it other than just remembering ts context disappeared.

> We can adapt it from time to time or forking it if thing changes a lot. Or implement it in fzf-lua is also an option since it only have 1400 loc...we can just use queries file from that repo.

I’d rather not do that.

ibhagwan:
This is very hacky and prone to fail IMHO, I’m not sure if this is worth it just so we can avoid the edge case when the ts-context overlaps with the one from the origin window.

ibhagwan:
> I have to admit it looks hacky...although it's mainly to fix previous hacks.

I see your point, the whole thing is hacky, might as well at least work as expected, until the upstream changes…

Ty @phanen!

Issue #2064, opened 2025-05-24
Bug: snacks throws error trying to inline render images in previewer


carloscalla:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

macOS 15.5

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.1

### Fzf version (`fzf --version`)

0.62.0 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [x] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
      previewers = {
        builtin = {
          snacks_image = { enabled = true },
        },
      },
})
```

### Describe the bug / steps to reproduce

When enabling `snacks_image` and the previewer shows a file with an image in it (frequently `.md` files), snacks is trying to inline render the image and fails for some cases. This problem doesn't break the plugin but it shows the error and interrupts the workflow, after acknowledging the error it goes away and doesn't pop up again as far as I was able to test. It only throws the error the first time.

I gathered a couple of examples with the same file by using the files and live_grep pickers.

Files:

![Image](https://github.com/user-attachments/assets/448b2f9b-01f2-4e37-a796-d151fa636090)

Grep:

![Image](https://github.com/user-attachments/assets/4d963122-f16b-4a0a-877e-1c1c250472b1)

I tested this in my config and also in a pretty clean config to trace down the error and rule out that it was some other plugin causing this.

Note: when using the Snacks files picker, I believe it errors as well but silently, because some images are not inline rendered but the error is not thrown in the screen.

ibhagwan:
@phanen, maybe we should wrap the snacks call in `pcall` to ignore the error?

phanen:
iirc we cannot, since this is printed not error.

ibhagwan:
> Note: when using the Snacks files picker, I believe it errors as well but silently, because some images are not inline rendered but the error is not thrown in the screen.

@carloscalla, are you certain about this? What happens when you  open the README as a buffer? Do you get the error?

carloscalla:
[...]

PR #2074, opened 2025-05-26
feat(image): add option to control image inline render

carloscalla:
- Adds `render_inline` prop to control whether to do image inline render in fzf-lua builtin previewer or not.
- This options is added because as explained in https://github.com/ibhagwan/fzf-lua/issues/2064 if inline render is enabled in Snacks, fzf-lua is detecting this option and applying it to the previewer as well.
- Enables the user to control better customization having a separate option for the previewer.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1013,8 +1013,9 @@ function Previewer.base:attach_snacks_image_inline()
   local bufnr, preview_winid = self.preview_bufnr, self.win.preview_winid
   if not simg
       or not self.snacks_image.enabled
+      or not self.snacks_image.render_inline
       or not simg.supports_terminal()
-      or not (simg.config.doc.enabled and simg.config.doc.inline and simg.terminal.env().placeholders)
+      or not simg.terminal.env().placeholders
       or vim.b[bufnr].snacks_image_attached then
     return
   end
--- a/README.md
+++ b/README.md
@@ -744,7 +744,7 @@ previewers = {
       -- render_markdown.nvim integration, enabled by default for markdown
       render_markdown = { enabled = true, filetypes = { ["markdown"] = true } },
       -- snacks.images integration, enabled by default
-      snacks_image = { enabled = true },
+      snacks_image = { enabled = true, render_inline = true },
     },
     -- Code Action previewers, default is "codeaction" (set via `lsp.code_actions.previewer`)
     -- "codeaction_native" uses fzf's native previewer, recommended when combined with git-delta
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -218,7 +218,7 @@ M.defaults                      = {
       ueberzug_scaler   = "cover",
       title_fnamemodify = function(s) return path.tail(s) end,
       render_markdown   = { enabled = true, filetypes = { ["markdown"] = true } },
-      snacks_image      = { enabled = true },
+      snacks_image      = { enabled = true, render_inline = true },
       _ctor             = previewers.builtin.buffer_or_file,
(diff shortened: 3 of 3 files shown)


lua/fzf-lua/previewer/builtin.lua
```diff
@@ -1013,8 +1013,9 @@ function Previewer.base:attach_snacks_image_inline()
   local bufnr, preview_winid = self.preview_bufnr, self.win.preview_winid
   if not simg
       or not self.snacks_image.enabled
+      or not self.snacks_image.render_inline
       or not simg.supports_terminal()
-      or not (simg.config.doc.enabled and simg.config.doc.inline and simg.terminal.env().placeholders)
```
carloscalla, line 1017: I am removing the conditions referring to Snacks config, allowing fzf-lua config to control this behavior.

ibhagwan: I agree, this is better indeed.

ibhagwan:
It says the PR is still a draft but I believe it’s ready to be merged.

ibhagwan:
Thank you @carloscalla!

carloscalla:
> It says the PR is still a draft but I believe it’s ready to be merged.

@ibhagwan sorry, I had to step out for a moment so I left the PR in draft to ask a question before merging, though it was ready to be merged as is.

I am using this condition now:

```lua
  if not simg
      or not self.snacks_image.enabled
      or not self.snacks_image.render_inline
      or not simg.supports_terminal()
      or not simg.terminal.env().placeholders
```

So I check if `snacks_image` is enabled and then use the `render_inline` option. I was wondering if there should be a separation between image files preview and inline render in documents giving more control for the behavior. I didn't think it was worth covering this at first but I was going to ask here if needed, something like:

```lua
snacks_image = { img_preview = true, render_inline = true },
```

Where `img_preview` (or a better name) would control the image file previews itself and `render_inline` would control the inline rendering in documents. I didn't think a user would enable `snacks_image` to only enable `render_inline` but not wanting to preview images, but the inverse case implemented in this PR to preview images but turn inline off, I though was valid.

I was just going to ask if this separation was worth implementing but the more I think of it the less I think it makes sense so it seems to be good as is.

Thanks for your feedback.

ibhagwan:
> > It says the PR is still a draft but I believe it’s ready to be merged.
> 
> @ibhagwan sorry, I had to step out for a moment so I left the PR in draft to ask a question before merging, though it was ready to be merged as is.
> 
> I am using this condition now:
> 
> ```lua
>   if not simg
>       or not self.snacks_image.enabled
>       or not self.snacks_image.render_inline
>       or not simg.supports_terminal()
>       or not simg.terminal.env().placeholders
> ```
> 
> So I check if `snacks_image` is enabled and then use the `render_inline` option. I was wondering if there should be a separation between image files preview and inline render in documents giving more control for the behavior. I didn't think it was worth covering this at first but I was going to ask here if needed, something like:
> 
> ```lua
> snacks_image = { img_preview = true, render_inline = true },
> ```
> 
> Where `img_preview` (or a better name) would control the image file previews itself and `render_inline` would control the inline rendering in documents. I didn't think a user would enable `snacks_image` to only enable `render_inline` but not wanting to preview images, but the inverse case implemented in this PR to preview images but turn inline off, I though was valid.
> 
> I was just going to ask if this separation was worth implementing but the more I think of it the less I think it makes sense so it seems to be good as is.
> 
> Thanks for your feedback.

I feel this isn’t required, if someone has snacks image enabled I assume they wanna see images, I can’t think of a scenario where someone would want this enabled, but only inline and not images, maybe if something is broken and they wanna fallback on chafa but sounds like an edge case.

If this issue comes up one day we can easily add that, rn it would just add another confusing setting.

Issue #2067, opened 2025-05-25
Feature: Improve diagnostics picker


toupeira:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

I customized the `diagnostic` provider in [my dotfiles](https://github.com/toupeira/dotfiles/blob/main/vim/util/fzf-diagnostics.lua) a bit to:

- Highlight the file headers with the diagnostic color (warnings = orange etc.)
- Append the diagnostic code to the message if present

I'd like to upstream these changes, but wanted to ask first what approach you prefer:

1. Add the changes, probably enabled by default, but also add settings to disable them (e.g. `color_filenames`, `diag_code`, to match the existing `color_icons` and `diag_source`)
2. Replace the [diagnostic formatting](https://github.com/ibhagwan/fzf-lua/blob/1e0339b9c7e55d7738dc50067392dda3278e18a0/lua/fzf-lua/providers/diagnostic.lua#L211-L221) with a callback that users can override in their config. Not sure yet if this would be sufficient, I might need to hook into [`preprocess_diag`](https://github.com/ibhagwan/fzf-lua/blob/1e0339b9c7e55d7738dc50067392dda3278e18a0/lua/fzf-lua/providers/diagnostic.lua#L192) too.

Screenshots:

| Before | After |
| --- | --- |
| ![Image](https://github.com/user-attachments/assets/3f45ba21-0909-4f83-891e-5ddce55013e5) | ![Image](https://github.com/user-attachments/assets/9ccfc7d0-dd5d-44ad-b9b3-8c7be844560c)

Diff:

```diff
--- vim/util/fzf-diagnostics.lua	2025-05-24 06:19:20.003399898 +0200
+++ packages/lazy/fzf-lua/lua/fzf-lua/providers/diagnostic.lua	2025-05-20 01:34:45.999991798 +0200
@@ -196,14 +194,6 @@
                 coroutine.resume(co)
                 return
               end
-
-              local sign_def
-              if opts.__signs[diag.severity] then
-                sign_def = opts.__signs[diag.severity]
-              end
-
-              diag_entry.filename = utils.ansi_from_hl(sign_def.texthl, diag_entry.filename)
-
               local entry = make_entry.lcol(diag_entry, opts)
               entry = make_entry.file(entry, opts)
               if not entry then
@@ -212,22 +202,20 @@
               else
                 local icon = nil
                 if opts.__signs[diag.severity] then
[...]

PR #2075, opened 2025-05-26
feat(diagnostics): add options to color headings and show diag code

toupeira:
This implements the changes discussed in https://github.com/ibhagwan/fzf-lua/issues/2067:

- Highlight the file headers with the diagnostic color (warnings = orange etc.)
- Append the diagnostic code to the message if present

Both are disabled by default and can be enabled with the `color_headings` and `diag_code` options.

### DIFF
--- a/doc/fzf-lua.txt
+++ b/doc/fzf-lua.txt
@@ -1244,11 +1244,13 @@ CUSTOMIZATION                                          *fzf-lua-customization*
       },
       diagnostics ={
         prompt            = 'Diagnostics❯ ',
+        color_headings    = false,  -- display filename and source in diag color (e.g. orange for warnings)
         cwd_only          = false,
         file_icons        = true,
         git_icons         = false,
         diag_icons        = true,
-        diag_source       = true,   -- display diag source (e.g. [pycodestyle])
+        diag_source       = false,  -- display diag source (e.g. [pycodestyle])
+        diag_code         = false,  -- display diag code (e.g. [E011])
         icon_padding      = '',     -- add padding for wide diagnostics signs
         multiline         = true,   -- concatenate multi-line diags into a single line
         -- severity_only:   keep any matching exact severity
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -970,19 +970,21 @@ M.defaults.lsp.code_actions     = {
 }
 
 M.defaults.diagnostics          = {
-  previewer   = M._default_previewer_fn,
-  file_icons  = 1,
-  color_icons = true,
-  git_icons   = false,
-  diag_icons  = true,
-  diag_source = false,
-  multiline   = 2,
-  fzf_opts    = {
+  previewer      = M._default_previewer_fn,
+  file_icons     = 1,
+  color_icons    = true,
+  color_headings = false,
+  git_icons      = false,
+  diag_icons     = true,
+  diag_source    = false,
+  diag_code      = false,
+  multiline      = 2,
+  fzf_opts       = {
     ["--multi"] = true,
     ["--wrap"]  = true,
   },
-  _actions    = function() return M.globals.actions.files end,
-  _cached_hls = { "path_colnr", "path_linenr" },
+  _actions       = function() return M.globals.actions.files end,
+  _cached_hls    = { "path_colnr", "path_linenr" },
   -- signs = {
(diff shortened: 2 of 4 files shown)


lua/fzf-lua/defaults.lua
```diff
@@ -970,14 +970,16 @@ M.defaults.lsp.code_actions     = {
...
+  previewer      = M._default_previewer_fn,
+  file_icons     = 1,
+  color_icons    = true,
+  color_headings = false,
+  git_icons      = false,
+  diag_icons     = true,
+  diag_source    = false,
+  diag_code      = false,
+  multiline      = 2,
+  fzf_opts       = {
```
toupeira, line 982: > Enabled by default is fine for the source but not for the filename IMHO, if someone uses a formatter (e.g. path.dirname_first which colors the dir) this will override it, maybe we can provide an option for the filename coloring?

I ended up disabling `diag_code` by default, since `diag_source` is too.

(also I assume you wanted me to reindent here, if not I'll be happy to change it :grinning:)

doc/fzf-lua.txt
```diff
@@ -1244,11 +1244,13 @@ CUSTOMIZATION                                          *fzf-lua-customization*
       },
       diagnostics ={
         prompt            = 'Diagnostics❯ ',
+        color_headings    = false,  -- display filename and source in diag color (e.g. orange for warnings)
         cwd_only          = false,
         file_icons        = true,
         git_icons         = false,
         diag_icons        = true,
-        diag_source       = true,   -- display diag source (e.g. [pycodestyle])
+        diag_source       = false,  -- display diag source (e.g. [pycodestyle])
```
toupeira, line 1252: Docs were incorrect here, `diag_source` is disabled by default.

lua/fzf-lua/providers/diagnostic.lua
```diff
@@ -194,29 +194,41 @@ M.diagnostics = function(opts)
                 coroutine.resume(co)
                 return
               end
+
+              local sign_def = opts.__signs[diag.severity]
+
+              if opts.color_headings then
+                diag_entry.filename = utils.ansi_from_hl(sign_def.texthl, diag_entry.filename)
+              end
```
toupeira, line 202: The structure is a bit messy, we need to add highlights to the filename before passing it to `make_entry.lcol()` and `.file()`. And then the other decisions are made further down when we're actually formatting the diagnostic.

Was considering moving more logic into `preprocess_diag`, but not sure if that's a good place.

doc/fzf-lua.txt
```diff
@@ -1244,11 +1244,13 @@ CUSTOMIZATION                                          *fzf-lua-customization*
       },
       diagnostics ={
         prompt            = 'Diagnostics❯ ',
+        color_headings    = false,  -- display filename and source in diag color (e.g. orange for warnings)
```
toupeira, line 1247: Regarding naming: I figured I'd call it "headings" since the source is also colored, but I'm also aware that "headings" has a specific meaning in FZF, so I'm not sure if this is a good name :wink:

ibhagwan: I’m not sure either, maybe `color line`? Or ` color line1`?

lua/fzf-lua/defaults.lua
ibhagwan: Added a commit that made all this default :)

lua/fzf-lua/providers/diagnostic.lua
ibhagwan: Guess we could refactor this nicely but not high priority IMHO

doc/fzf-lua.txt
ibhagwan: Also changed the default here.

ibhagwan:
@toupeira, this looks really nice, I added a commit that set this all to default as I also found out the formatter actually takes precedence so it won’t overwrite user’s formatter if such one was set.

we just have to decided if `color_ headings` is the right name and  we can merge this.

Ty, this is very nice :)

doc/fzf-lua.txt
toupeira: Hmm that doesn't sound clearer to me, and I also saw now in FZF they're called "headers" rather than "headings", so maybe `color_headings` is actually fine after all?

ibhagwan: Sure thing :)

toupeira:
@ibhagwan awesome, thank you too! :grinning:

PR #2079, opened 2025-05-27
fix(ts-context): upstream changes

phanen:
https://github.com/nvim-treesitter/nvim-treesitter-context/commit/33b3f5a06ded3f3dcbf5de433d1ea35caa9b14d3

PR #2080, opened 2025-05-27
fix(ts-context): update zindex when toggle context win

phanen:
Fix https://github.com/ibhagwan/fzf-lua/pull/2079#issuecomment-2912855760.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -86,10 +86,10 @@ end
 
 ---@param winid integer
 ---@param bufnr integer
----@param opts? { setup_opts: table, zindex: integer }
+---@param opts? TSContext.UserConfig
 function TSContext.update(winid, bufnr, opts)
   opts = opts or {}
-  if not TSContext.setup(opts.setup_opts) then return end
+  if not TSContext.setup(opts) then return end
   assert(bufnr == vim.api.nvim_win_get_buf(winid))
   local render = require("treesitter-context.render")
   render.close_contexts({ winid })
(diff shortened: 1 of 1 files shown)


phanen:
Sadly I found another bug...

phanen:
The bug is `WinResized` on context window mistakenly close itself, this should never happen.

ibhagwan:
Ty @phanen!

ibhagwan:
Shoot I rushed to merge this since the code looked good, but it doesn’t work and I see no context lol

ibhagwan:
Weird, it seems that `context_ranges` and `context_lines` is empty for me at line 96.

ibhagwan:
Yes `vim.g._ts_force_sync_parsing = true` this works, but I wonder why we need this now and not with previous versions.

There’s also an issue where the first contest appears for a sec and then disappears, toggle or scroll makes it reappear, I wish we had a decent API for this.

ibhagwan:
Confusing, sometimes it works without `vim.g._ts_force_sync_parsing`, sometimes flashes, I’ll also dig into the code of ts-context tomorrow.

Maybe we can call the ts-context only after our previewer TS parsing is done? I faintly recall a condition like that in the code somewhere

phanen:
For me it work well if open picker on a empty buffer.
It will flash if your current buffer has attached context window.

phanen:
Remove `render.close_contexts({ winid })` feels better...

ibhagwan:
> Remove `render.close_contexts({ winid })` feels better...

Do we really need it? There’s enough closes in the ts context plug-in code we don’t have to call our own, it will probably get closed anyways in a cleanup.

PR #2086, opened 2025-05-29
feat(commands): include subcommands

phanen:
```lua
-- Save one more type
require('fzf-lua').commands({ flatten = { Gitsigns = true } })

-- Discovery (all commands)
require('fzf-lua').commands({ flatten = vim.api.nvim_get_commands {} })

-- Customization
require('fzf-lua').commands({ flatten = { Foo = function(opts) return { 'Bar', 'Baz --abc' } end } })
```
image


ibhagwan:
Ty @phanen!

PR #2095, opened 2025-05-31
Load content from dap server

Davidyz:
# Context: 

Some debuggers are _asynchronous_. It means that the debugging and the code execution doesn't happen at the same time. An example of this is [coredumpy](https://github.com/gaogaotiantian/coredumpy). In their implementation, the source code that triggered the error is saved in a dump file, which can be loaded later. This can even be loaded on a different machine/system. In this case, the files on the local filesystem are often ignored because they're not guaranteed to be the same as the source code which caused the error (or simply don't exist at all).

# What this PR does

This PR attempts to load the source code saved in the dump file (provided by the dap server). This enables frame previewing when working with a dump file from [coredumpy](https://github.com/gaogaotiantian/coredumpy) using [coredumpy.nvim](https://github.com/Davidyz/coredumpy.nvim).

Before:
image

After:
image

[This](https://microsoft.github.io/debug-adapter-protocol//specification.html#Requests_Source) is the dap `source` request used to fetch the source code from supported server.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -891,6 +891,8 @@ function Previewer.buffer_or_file:populate_preview_buf(entry_str)
       local lines = nil
       if entry.path:match("^%[DEBUG]") then
         lines = { tostring(entry.path:gsub("^%[DEBUG]", "")) }
+      elseif entry.content then
+        lines = entry.content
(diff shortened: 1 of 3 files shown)


ibhagwan:
Tyvm @Davidyz for this PR.

@phanen, builds on your commit 8099351

Issue #2103, opened 2025-06-06
Bug: `FzfLua treesitter` stopped working after updating `treesitter` to a `main` branch


Ax51:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Mac

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.2

### Fzf version (`fzf --version`)

0.62.0 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

-m --height 50% --border

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [x] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
    grep = {
      rg_opts = "--column --line-number --no-heading --color=always --smart-case --multiline --max-columns=4096 -e",
      rg_glob = true,
      git_icons = true,
    },
    previewers = {
      builtin = {
        extensions = {
          ["png"] = { "viu", "-b" },
          ["jpg"] = { "viu", "-b" },
        }
      }
    },
})
```

### Describe the bug / steps to reproduce

Suddenly, after `nvim-treesitter` marked `master` branch as an archived one and suggest users to move to their new `main` branch I've found that `Fzf-lua treesitter` stopped working after the migration. Likely this is caused because `nvim-treesitter` authors heavily refactored their plugin and as for this moment there is no any `locals` module in their `main` branch: 

Seems that this line causes rejection: https://github.com/ibhagwan/fzf-lua/blob/6d1d81a5c7936d82a47a41a521958e482e499fde/lua/fzf-lua/providers/buffers.lua#L524

In `master` branch they had the following lua module that was supposed to be called: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/lua/nvim-treesitter/locals.lua

But in their new `main` branch there is nothing similar: https://github.com/nvim-treesitter/nvim-treesitter/blob/main/lua/nvim-treesitter/locals.lua

Immediately after opening treesitter view I received the following error message:

```
[...]

PR #2104, opened 2025-06-06
fix(treesitter): use locals.scm only

phanen:
https://github.com/ibhagwan/fzf-lua/issues/2103

Copy paste from `locals.lua`. Not sure if this will work with `master` branch.

### DIFF
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -504,9 +504,9 @@ M.treesitter = function(opts)
     opts._bufname = utils.nvim_buf_get_name(opts.bufnr)
   end
 
-
+  local ts = vim.treesitter
   local ft = vim.bo[opts.bufnr].ft
-  local lang = vim.treesitter.language.get_lang(ft)
+  local lang = ts.language.get_lang(ft) or ft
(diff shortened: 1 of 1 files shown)


phanen:
Tested with `mini.sh` with current default branch of `nvim-treesitter`. No problem to me.

ibhagwan:
> Tested with `mini.sh` with current default branch of `nvim-treesitter`. No problem to me.

Ty @phanen!

ibhagwan:
Tested and works on master, I can’t get main to work, I also get issue with treesitter-context configs, I think we should wait with this till the main branch is stabilized and becomes default otherwise we’re just chasing our tails with patches as there might be many more changes?

phanen:
It won't work if no query file found. And no `main` branch query file is symlink in builtin rtp (`.local/share/nvim/site/queries/...` for me, can be installed by `require("nvim-treesitter").install { "lua" }`).

> I think we should wait with this till the main branch is stabilized and becomes default otherwise we’re just chasing our tails with patches as there might be many more changes?

Yeah. But with this PR, the picker only need `queries/**/locals.scm` to work. I think there's no more break changes (unless upstream also decide to remove query file...)

PR #2106, opened 2025-06-08
fix(ts-context): ignore WinResized on scrollbar to avoid closing context win

phanen:
`WinResize` -> `update_win` -> `Render.close`

(`noautocmd=true` don't ignore `WinResized`)

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -1312,6 +1312,7 @@ function FzfWin:update_preview_scrollbar()
       empty.noautocmd = true
       self._sbuf1 = ensure_tmp_buf(self._sbuf1)
       self._swin1 = utils.nvim_open_win0(self._sbuf1, false, empty)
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @phanen!

This is going to break on 0.10 without adding `fn.exists("+eventignorewin")` no?

Should we add this as a static in utils:
https://github.com/ibhagwan/fzf-lua/blob/933b06ce34353d9b295c577c64c2a500669a9429/lua/fzf-lua/utils.lua#L13-L18

and then we can also get rid of the call at:
https://github.com/ibhagwan/fzf-lua/blob/933b06ce34353d9b295c577c64c2a500669a9429/lua/fzf-lua/previewer/builtin.lua#L108

phanen:
`set_winopts` handle it:
https://github.com/ibhagwan/fzf-lua/blob/0e7f8ccee06e89d5e9da4edabcb932e714c053c1/lua/fzf-lua/win.lua#L709

ibhagwan:
Correct :)

PR #2117, opened 2025-06-18
fix(actions): ex_run don't save history

phanen:
When hit enter myself it seems never save the history for me.


ibhagwan:
Ty @phanen !

Issue #970, opened 2023-12-28
Slower `--startuptime` than expected


nvs:
### Info

- Operating System: WSL / Arch Linux
- Shell: bash
- Terminal: Windows Terminal
- `nvim --version`: v0.9.4
- `fzf --version`: 0.44.1

- [x] The issue is reproducible with `mini.sh`

### Description 

I was looking at various `--startuptime` values, and noticed that a fair bit of time was spent in `require('fzf-lua.defaults')`. The time spent here noticeably increased after 407323ade0ebca1bb4f26453c1aacbb246654139. After looking into it a bit, it seems that `vim.fn.executable` is the primary source of time usage within this file.

For my setup, using the minimal init above (with an added `--startuptime` argument) I get the following:
```txt
007.227  001.254  000.536: require('fzf-lua.utils')
007.233  001.506  000.252: require('fzf-lua.path')
008.043  000.507  000.507: require('fzf-lua.actions')
008.424  000.041  000.041: require('fzf-lua.previewer')
074.919  066.871  066.830: require('fzf-lua.defaults')
074.932  067.698  000.320: require('fzf-lua.config')
075.069  000.055  000.055: sourcing /tmp/fzf-lua.tmp/nvim/site/pack/vendor/start/fzf-lua/plugin/fzf-lua.vim
076.692  000.744  000.744: require('fzf-lua.win')
077.318  000.113  000.113: require('fzf-lua.fzf')
077.536  000.216  000.216: require('fzf-lua.libuv')
077.712  000.174  000.174: require('fzf-lua.shell')
079.389  001.676  001.676: require('fzf-lua.make_entry')
079.395  002.700  000.521: require('fzf-lua.core')
079.402  074.003  001.301: require('fzf-lua')
```
If I then edit `defaults.lua` in the following fashion:
```diff
diff --git a/lua/fzf-lua/defaults.lua b/lua/fzf-lua/defaults.lua
index bdbc67d..849934d 100644
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -141,17 +141,17 @@ M.defaults = {
       _ctor = previewers.fzf.cmd,
     },
     bat = {
-      cmd   = vim.fn.executable("batcat") == 1 and "batcat" or "bat",
+      cmd = "bat",
       args  = "--color=always --style=numbers,changes",
       _ctor = previewers.fzf.bat_async,
     },
     bat_native = {
[...]

PR #2119, opened 2025-06-20
perf: improve startup time by deferring config defaults

EdenEast:
This change wraps the function `_man_cmd_fn` which checks the system to find which man previewer is installed. This system check is expensive and does not need to be called until it is actually required. This reduces the startup time of fzf-lua. An example of the difference via [vim-startuptime](https://github.com/dstein64/vim-startuptime):

### Before:
```
event                  time percent
fzf-lua               41.34   42.38
fzf-lua.config        37.96   38.91
fzf-lua.defaults      37.57   38.51
```

### After:
```
event                  time percent
fzf-lua                3.29    5.66
fzf-lua.config         0.98    1.69
fzf-lua.path           0.67    1.15
fzf-lua.defaults       0.53    0.91
```

Relates: #970

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -190,11 +190,11 @@ M.defaults                      = {
     },
     man = {
       _ctor = previewers.builtin.man_pages,
-      cmd = M._man_cmd_fn(),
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @EdenEast!

PR #2123, opened 2025-06-21
fix(win): winopts.fullscreen can suppress cmdline area when zindex >= 200

phanen:
https://github.com/ibhagwan/fzf-lua/discussions/2121


ibhagwan:
Ty @phanen!

Issue #2122, opened 2025-06-21
Bug: `search_history` and `command history` truncate wide entries


aidancz:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

linux

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-630+g8af2aea24f

### Fzf version (`fzf --version`)

0.62.0 (d226d841)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=16 --color=fg:-1 --color=fg+:-1:reverse --color=bg:-1 --color=bg+:-1:reverse    --color=gutter:-1 --color=hl:4 --color=hl+:4:reverse --color=query:-1 --color=info:-1 --color=border:-1 --color=label:-1 --color=prompt:-1 --color=pointer:-1 --color=marker:-1 --color=spinner:-1 --color=header:-1 --bind=ctrl-a:first --bind=ctrl-e:last --no-unicode --pointer= --scrollbar= --highlight-line --scroll-off=0 --walker=file,dir,follow,hidden

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

### Describe the bug / steps to reproduce

search a string wider than `vim.o.columns`, run `:FzfLua search_history`

for example:

1. `set columns=80`
2. press `/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa<cr>`

![Image](https://github.com/user-attachments/assets/f8c9fa93-413f-43e1-9cdb-85526768273e)

the middle part is replaced by 3 dots

I recommend using vim.fn.histnr() and vim.fn.histget() instead of vim.fn.execute("history " .. str)

phanen:
~Maybe a hack is toggling `vim.o.columns`: 
[...]

PR #2124, opened 2025-06-21
fix: avoid long history entries truncated

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/2122.


ibhagwan:
Ty @phanen!

This just solves it for actions, should we also change the picker input itself to histget? Then we don’t have to modify the actions code.

phanen:
I guess that would be ~10000 times calling on `vim.fn.histget`...

phanen:
It seems not slow. I will do that.

PR #2125, opened 2025-06-21
fix(history): missing entries, also asyncify

phanen:
* `3` was used to skip the header and no longer needed now.
* asyncify to make it feels faster


ibhagwan:
This is great Ty @phanen!

PR #2132, opened 2025-06-24
perf(buffers): avoid call getbufinfo in filter_buffers

phanen:
`getbufinfo` is called twice.

(`utils.buf_is_qf` -> `getbufinfo`)


ibhagwan:
Ty @phanen!

PR #2135, opened 2025-06-24
fix(actions): make `:tabnew | h xx | only` work with &splitkeep

phanen:
`nvim --clean -u repro.lua`:
```lua
vim.o.scrolloff = math.ceil(vim.o.lines / 2)
vim.o.splitkeep = 'topline'
vim.cmd [[tabnew | help shada | only]]
```

Cursor won't focus on tags.

For fzf-lua not sure if this problem exist for old version.. since there's another bug [6a71239cd5ea2a8e993bddd8e7765d1afe3e79e4](https://github.com/phanen/neovim/commit/6a71239cd5ea2a8e993bddd8e7765d1afe3e79e4). And this just fix on 0.11 (which has `vim._with`) only.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -702,7 +702,9 @@ end
 M.help_tab = function(selected, opts)
   if #selected == 0 then return end
   -- vim.cmd("tab help " .. helptags(selected, opts)[1])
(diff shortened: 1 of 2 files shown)


ibhagwan:
Ty @phanen, seems reasonable.

PR #2141, opened 2025-06-27
feat: `spellcheck` picker, all misspelled words in buffer

ibhagwan:


ibhagwan:
@phanen, would love to have your opinion about the new spellcheck picker, the code is roughly designed after https://github.com/inkarkat/vim-SpellCheck.

Tried to find another way to list all misspelled words aside from manual navigation using `]s` but apparently there isn’t such way? Maybe there is on neovim and I missed something?

If `]s` navigation is the  solution, I’m trying to find a way to avoid the scroll redraw on the background window, it’s not pretty and also has a performance cost, I tried to temporarily set `lazyredraw` but it didn’t help - do you know if any other solution to temporarily suspend draw  until the last word is found?

Something else that’s happening that may lead to the solution is that I found out that after render-markdown.nvim is loaded (single call to `require("render-markdown")` is enough), the background window won’t scroll anymore but the interface only shows the entries only after all the entries are loaded which is odd since this is running in a coroutine+vim.schedule.

phanen:
I don't know. Just test it's a bit slow to load this picker on README.md.

`typos-lsp` code action/ diagnosctics is another OOTB way...


>    (single call to require("render-markdown") is enough), the background window won’t scroll anymore

No the bg still scroll for me..

Maybe extmark can be used to get misspell pos.

And neovim has `vim.spell.check` https://github.com/ribru17/blink-cmp-spell/blob/00bf4d9ba985e29cbe4fba70b0b77e9ab0f94894/lua/blink-cmp-spell/init.lua#L69.

ibhagwan:
> typos-lsp code action/ diagnosctics is another OOTB way...

It looks very nice but it’s a dependency, I’m trying to find a solution for the native spell checking, it’s not amazing but it’s something but I’m very surprised there isn’t a way to enumerate all misspelled words without having to traverse the document with `]s`, I’ll have to dig into neovim source to see if there’s a hack.

> And neovim has vim.spell.check

I’ve seen that one but it’s an alternative to `spellbadword`, I’ll take a look in the code and see if there are some leads there.

ibhagwan:
@phanen, check it out now, much better and faster, using the same logic as `blines` to iterate through all lines and `%w+` to iterate through all words with `vim.spell.check`.

The only issue remaining is why there's a difference between `spellbadword` and `vim.spell.check`, for example, spellbadword will skip the word `config` (considering it a "good" word), while spellcheck returns it as a bad word `{ { "config", "bad", 1 } }`.

ibhagwan:
~~I think I also need to change the logic to only consider words that have spaces around them but that will further differentiate from `spellbadword` logic.~~

Decided to go with a customizable approach using `opts.word_separator`, defaults to spaces and punctuation marks `[%s%p]`

PR #2142, opened 2025-06-28
fix(actions): `:hi ` executed twice

phanen:


nenahp:
This seems intended: https://github.com/ibhagwan/fzf-lua/pull/1541#issuecomment-2504452220...

I didn't notice since it has no problem with `require('vim._extui').enable{}`.

phanen:
Ok I cannot figure it out...

PR #2143, opened 2025-06-28
fix(actions): `:hi` executed twice

phanen:
Anyway I want to avoid show twice when enable `vim._extui`


nenahp:
previous: https://github.com/ibhagwan/fzf-lua/pull/2142.

ibhagwan:
Ty @phanen!

Issue #2149, opened 2025-07-02
Bug: Pickers are not working


Emille1723:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux & WSL

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.11.2

### Fzf version (`fzf --version`)

0.62.0 (d226d841)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=fg+:#ffffff,bg+:#343C4B,hl+:#50fa7b --color=info:#bd93f9,prompt:#bd93f9,pointer:#fa50cf --color=marker:#63f2f1,spinner:#50fa7b,header:#8ABAB8 --color=border:#6272a4,label:#6272a4,gutter:-1,separator:#bd93f9 --highlight-line --layout=reverse --no-height --border=top --preview-window="right,60%" --prompt="🔍" --pointer="█ " --marker="▐ " --marker-multi-line="▐▐▐" --margin="2.5%,0,0,0" --info=default --scrollbar=▌▐ --preview-label-pos="3:bottom" --preview-window="60%,border-rounded" --bind="ctrl-j:down" --bind="ctrl-k:up" --bind="esc:become(exit 0;)" --bind="alt-bspace:clear-query"  --bind="ctrl-c:abort" --bind="ctrl-d:preview-page-down" --bind="ctrl-u:preview-page-up" --bind="alt-v:replace-query" --bind="ctrl-y:execute(echo {} | wl-copy;)" --cycle --ellipsis=STR --ansi --wrap --ignore-case

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
local fzf_lua_opts = {
    "telescope",
	-- file_icon_padding = ' ',
	fzf_colors = {
		-- ["fg"]          = { "fg", "#d3D3D3" },
		["fg"]          = { "fg", "#bec0d3" },
		["bg"]          = { "bg", "Normal" },
		["hl"]          = { "fg", "#50fa7b" },
		-- ["fg+"]         = { "fg", "#ffffff" },
[...]

PR #2150, opened 2025-07-02
fix(config): always remove the first boolean

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/2149.


ibhagwan:
Ty @phanen!

Issue #1879, opened 2025-03-12
Feature: add a combined picker between buffers & git_files/files


LoricAndre:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

I'd love a picker that would show me my open buffers at the top, then show me files (or git_files, I'm not so sure). I tried building it from the public API but most utility functions are local which makes it quite complicated to implement robustly and maintain with future changes.
More generally (but probably harder), it would be interesting to be able to combine multiple pickers, and be able to specify the ordering.

LoricAndre:
Note: a way for me to build something like this in my config would be to expose the `contents` for specific pickers, that I could retrieve and try to combine myself

LoricAndre:
For now, I put this together, but I'm still missing the icons for the buffers:
```lua
function FzfLuaSmart(opts)
  local config = require('fzf-lua.config')
  local core = require('fzf-lua.core')
  local path = require('fzf-lua.path')
  local libuv = require('fzf-lua.libuv')
  local make_entry = require('fzf-lua.make_entry')
  opts = config.normalize_opts(opts, 'git.files')
  if not opts then
    return
  end
  opts.cwd = path.git_root(opts)
  opts.git_icons = true
  opts.file_icons = true
  if not opts.cwd then
    return
  end
  local git_cmd = core.mt_cmd_wrapper(opts)
  local git_fn = libuv.spawn_nvim_fzf_cmd({
    cmd = git_cmd,
    cwd = opts.cwd,
    cb_pid = function(pid)
      opts.__pid = pid
    end,
  })
  local contents = function(x, cb, y)
    make_entry.preprocess(opts)
    for _, buf_id in ipairs(vim.api.nvim_list_bufs()) do
      if
        vim.api.nvim_buf_is_loaded(buf_id) and vim.fn.buflisted(buf_id) == 1
      then
        local p = vim.api.nvim_buf_get_name(buf_id)
        local exists, stats = pcall(vim.uv.fs_stat, p)
        if exists and stats and stats.type == 'file' then
          cb(make_entry.file(p .. '\n', opts))
        end
      end
    end

    git_fn(x, cb, y)
  end
  opts = core.set_header(opts, opts.headers or { 'cwd' })
  return core.fzf_exec(contents, opts)
end
```

phanen:
Maybe in this way you can get icons for buffer:
```diff
@@ -31,7 +31,7 @@ function FzfLuaSmart(opts)
[...]

Issue #2063, opened 2025-05-24
Implement reload = true for custom functions


ibhagwan:
### Discussed in https://github.com/ibhagwan/fzf-lua/discussions/2062

<div type='discussions-op-text'>

<sup>Originally posted by **stuckinsnow** May 24, 2025</sup>
Hello, I am making this feature request as advised in another thread -- https://github.com/ibhagwan/fzf-lua/discussions/2061. Reload = true does not currently work for custom functions, and there's a few cases I can think of where this could be useful:

- Quickfix menu
- Loclist menu
- Custom harpoon menu 

Example loclist configuration --
```lua
  actions = {
    ["Ctrl-x"] = {
      noclose = true,
      fn = function(_, opts)
        local loclist = vim.fn.getloclist(0, { items = 0 }).items

        if not loclist or #loclist == 0 then
          print("Location list is empty.")
          return
        end

        local current_index = vim.fn.getloclist(0, { idx = 0 }).idx
        if current_index and loclist[current_index] then
          table.remove(loclist, current_index)
          vim.fn.setloclist(0, loclist, "r")
          print("Deleted item at index: " .. current_index)
        else
          print("Invalid selection.")
        end
        FzfLua.loclist(opts)
      end,
    }
  },
```
With a functional reload = true, the menu would no longer flash--much like how deleting buffers works. </div>

ibhagwan:
Still WIP try https://github.com/ibhagwan/fzf-lua/tree/dev with, the `.idx` didn't work anymore since the reload action happens in the context of fzf-lua (there's no process restart or win switch so we also have to use `CTX().bufnr`.
```lua
FzfLua.loclist({
  actions = {
    ["Ctrl-x"] = {
      reload = true,
      fn = function(sel, opts)
        local bufnr = require("fzf-lua").core.CTX().bufnr
        local loclist = vim.fn.getloclist(bufnr, { items = 0 }).items

        local bufmap = (function()
          local ret = {}
          for _, s in ipairs(sel) do
            local b = s:match("%[(%d+)%]")
            ret[b] = true
          end
          return ret
        end)()

        local newlist = {}
        for i, l in ipairs(loclist) do
          if bufmap[tostring(l.bufnr)] then
            print(string.format("Deleted index %d bufnr %d ", i, l.bufnr))
          else
            table.insert(newlist, l)
          end
        end

        vim.fn.setloclist(bufnr, newlist, "r")
      end,
    }
  },
})
```

[...]

PR #2152, opened 2025-07-04
Convert all content types to string

### LINKED ISSUE #2058: Feature: Global Search Behavior Inspired by VS Code (@, #, $, etc.)
### Have you RTFM'd?

### Feature Request

Hi everyone — first of all, thank you for this awesome plugin! It’s fast, flexible, and makes Neovim feel like a modern editor.

I’d love to suggest a feature that mimics VS Code’s global search behavior via prefixes in the `Ctrl+p` menu. The idea is to allow smart, context-aware searches by using specific starting characters. Here's what I mean:

|Prefix|Behavior|
|---------|------------|
|`@`|Search symbols in current file|
|`#`|Search workspace/global symbols (via LSP)|
|`no prefix`|Search files by name|
|`$`|Search open buffers|

This approach keeps search unified in a single interface while still offering contextual power — similar to VS Code’s `Ctrl+p`. It would reduce the cognitive load of remembering multiple commands and help users move even faster inside Neovim.

🙋 **Is this possible?**
Is it currently feasible to implement something like this via fzf-lua or fzf.vim? And if so, would it make sense to integrate this behavior into the plugin or expose an interface to let users define prefix handlers?

Thanks again for the great work, and I’d love to hear your thoughts! 🚀

--- 
Here is how I have implemented this currently using fzf

```lua
return {
  "ibhagwan/fzf-lua",
  dependencies = { "nvim-tree/nvim-web-devicons" },
  opts = {},
  config = function()
    local fzf = require("fzf-lua")
    local config = fzf.config
    local actions = config.actions

    local function vscode_search()
      vim.ui.input({ prompt = "Search: " }, function(input)
        if not input then return end
        
        if input:sub(1, 1) == "#" then
          -- Global/workspace symbols search
          fzf.lsp_workspace_symbols({ query = input:sub(2) })
        elseif input:sub(1, 1) == "@" then
          -- Document symbols search
          fzf.lsp_document_symbols({ query = input:sub(2) })
        else
          -- File search
          fzf.files({ query = input })
        end
      end)
    end

    fzf.setup({})
    
    -- other mappings
    vim.keymap.set("n", "<leader>ff", vscode_search, { desc = "VSCode-like search" })
  end,
}
```

But this doesn't feel smooth and fast.

ibhagwan:
`fzf_exec` now converts all content types (tables/funcs) to shell command strings this greatly simplifies the code as all picker inputs are now shell commands which can be used in fzf's reload, execute-silent, etc.

This will also enable us to combine pickers easily with semicolon or logical and, e.g. `cmd1; cmd2` or `cmd1 && cmd2`.

Fixes #2058 and prepares the ground for #1879, #2058.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -382,7 +382,7 @@ function M.normalize_opts(opts, globals, __resume_key)
   end
 
   -- `fzf_cli_args` is string, `_fzf_cli_args` is a table used internally
-  opts._fzf_cli_args = {}
+  opts._fzf_cli_args = type(opts._fzf_cli_args) == "table" and opts._fzf_cli_args or {}
 
   -- backward compatibility, rhs overrides lhs
   -- (rhs being the "old" option)
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -360,6 +360,21 @@ M.defaults.files                = {
   winopts                = { preview = { winopts = { cursorline = false } } },
 }
 
+M.defaults.global               = vim.tbl_deep_extend("force", M.defaults.files, {
+  cwd_prompt        = true,
+  line_query        = true,
+  pickers           = {
+    [""]  = "files",
+    ["$"] = "buffers",
+    ["@"] = "lsp_document_symbols",
+    ["#"] = "lsp_workspace_symbols",
+  },
+  fzf_opts          = { ["--nth"] = false, ["--with-nth"] = false },
+  winopts           = { preview = { winopts = { cursorline = true } } },
+  _ctx              = { includeBuflist = true }, -- we include a buffer picker
+  _fzf_nth_devicons = false,
+})
+
 -- Must construct our opts table in stages
 -- so we can reference 'M.globals.files'
 M.defaults.git                  = {
--- a/lua/fzf-lua/init.lua
+++ b/lua/fzf-lua/init.lua
@@ -289,6 +289,7 @@ do
     tmux_buffers = { "fzf-lua.providers.tmux", "buffers" },
     profiles = { "fzf-lua.providers.meta", "profiles" },
     combine = { "fzf-lua.providers.meta", "combine" },
+    global = { "fzf-lua.providers.meta", "global" },
     complete_path = { "fzf-lua.complete", "path" },
     complete_file = { "fzf-lua.complete", "file" },
     complete_line = { "fzf-lua.complete", "line" },
--- a/lua/fzf-lua/profiles/default-prompt.lua
+++ b/lua/fzf-lua/profiles/default-prompt.lua
@@ -6,6 +6,7 @@ return {
   winopts              = { title_pos = "center", preview = { title_pos = "center" } },
   -- Uses `cwd_prompt` by default
   -- files                = prompt("Files"),
+  global               = prompt("Global"),
   buffers              = prompt("Buffers"),
   tabs                 = prompt("Tabs"),
   lines                = prompt("Lines"),
--- a/lua/fzf-lua/profiles/default-title.lua
+++ b/lua/fzf-lua/profiles/default-title.lua
@@ -9,6 +9,7 @@ end
 return {
   desc                 = "defaults using title for picker info",
   winopts              = { title_pos = "center", preview = { title_pos = "center" } },
+  global               = title("Global"),
   files                = title("Files"),
   buffers              = title("Buffers"),
   tabs                 = title("Tabs"),
--- a/lua/fzf-lua/providers/meta.lua
+++ b/lua/fzf-lua/providers/meta.lua
@@ -2,6 +2,7 @@ local uv = vim.uv or vim.loop
 local core = require "fzf-lua.core"
 local path = require "fzf-lua.path"
 local utils = require "fzf-lua.utils"
+local libuv = require "fzf-lua.libuv"
 local config = require "fzf-lua.config"
 
 local M = {}
@@ -121,7 +122,7 @@ M.combine = function(t)
 
   local cmds, opts = (function()
     local ret, opts = {}, nil
-    for i, p in ipairs(t.pickers) do
+    for _, p in ipairs(t.pickers) do
       -- local ok, msg, cmd, o = pcall(FzfLua[p], opts1)
       -- if not ok or not cmd then
       local _, cmd, o = FzfLua[p](opts1)
@@ -146,6 +147,102 @@ M.combine = function(t)
   local contents = table.concat(cmds, utils.__IS_WINDOWS and "&" or ";")
 
   return core.fzf_wrap(contents, opts)
+end
+
+M.global = function(opts)
+  opts = config.normalize_opts(opts, "global")
+  if not opts then return end
+
+  if opts.line_query and not utils.has(opts, "fzf", { 0, 59 }) then
+    utils.warn("'global' requires fzf >= 0.59, reverting to files.")
+    return FzfLua.files(opts)
+  end
+
+  -- Tells fzf_wrap to not start the fzf process
+  opts._start = false
+  local pickers = {}
+  local opts_copy = vim.deepcopy(opts)
+  for c, p in pairs(opts.pickers) do
+    if FzfLua[p] then
+      if #c == 0 then
+        -- Default picker opts set the tone for this picker options
+        -- this way convert reload / exec_silent actions will use a consistent
+        -- opts ref in the callbacks so we can later modify internal values
+        pickers[p] = { FzfLua[p](opts) }
+        -- Override opts with the return opts and store a copy of `pickers[]`
+        -- as we patch the opts when switching a picker in the change event
+        opts = pickers[p][3]
+        opts._start = nil -- remove the start suppression
+        pickers[p][3] = vim.deepcopy(opts)
+      else
+        -- Each subsequent picker gets a fresh copy of the original opts
+        -- (unmodified by the default picker)
+        pickers[p] = { FzfLua[p](opts_copy) }
+      end
+    else
+      utils.warn(string.format("invalid picker '%s', ignoring.", p))
+    end
+  end
+
+  ---@param q string
+  ---@return table?, integer?
+  local get_picker = function(q)
+    q = type(q) == "string" and q or ""
+    if #q == 0 then
+      return pickers[opts.pickers[""]], 1
+    end
+    for c, p in pairs(opts.pickers) do
+      if #c > 0 and q:match("^" .. utils.lua_regex_escape(c)) then
+        return pickers[p], #c + 1
+      end
+    end
+  end
+
+  -- Copy all keys without destroying the parent opts ref
+  -- or we will mess up the actions (especially hide)
+  local patch_opts = function(t1, t2)
+    for k, _ in pairs(t1) do
+      if k ~= "actions" then t1[k] = nil end
+    end
+    for k, v in pairs(t2) do
+      (function()
+        if k == "actions" then return end
+        if type(v) == "table" then
+          t1[k] = vim.tbl_deep_extend("force", t1[k] or {}, v)
+        else
+          t1[k] = v
+        end
+      end)()
+    end
+    return t1
+  end
+
+  -- Get starting picker
+  local cur_picker, cur_sub = get_picker()
+  if not cur_picker then
+    utils.warn("default picker not found, aborting.")
+    return
+  end
+
+  table.insert(opts._fzf_cli_args, "--bind=" .. libuv.shellescape("change:+transform:"
+    .. FzfLua.shell.stringify_data(function(args, _, _)
+      local q = args[1]
+      local new_picker, new_sub = get_picker(q)
+      local reload = ""
+      if new_picker and new_picker ~= cur_picker then
+        -- New picker requested, reload the contents and transform
+        -- the search string to exclude the picker prefix
+        cur_sub = new_sub
+        cur_picker = new_picker
+        -- Patch the opts refs with important values for path parsing
+        -- e.g. formatter, path_shorten, etc
+        opts = patch_opts(opts, cur_picker[3])
+        reload = string.format("reload(%s)+", new_picker[2])
+      end
+      return reload .. string.format("search(%s)", q:sub(cur_sub))
+    end, opts, "{q}")))
+
+  return core.fzf_wrap(cur_picker[2], opts)
 end
 
 return M
--- a/lua/fzf-lua/shell.lua
+++ b/lua/fzf-lua/shell.lua
@@ -83,10 +83,12 @@ function LRU:get(id)
 end
 
 -- Cache should be able to hold all function callbacks of a single picker
--- max cache size of 50 should be more than enough, we don't want it to be
+-- max cache size of 100 should be more than enough, we don't want it to be
 -- too big as this will prevent clearing of referecnces to "opts" which
 -- prevents garabage collection from freeing the resources
-local function new_cache(size) return LRU:new(size or 50) end
+-- NOTE: with combine/global the no. of callbacks has increased significantly
+-- so monitor the number of callbacks
+local function new_cache(size) return LRU:new(size or 100) end
 local _cache = new_cache()
 
 local M = {}


ibhagwan:
@phanen, still a lot more work , but starting to look like something...

TODO:
- [x] Merge `core.mt_cmd_wrapper` into `shell.stringify` and remove all calls to `core.mt_cmd_wrapper`
- [x] Replace `shell.raw_action` and `shell.raw_preview_action_cmd` with the new `shell.stringify` (should be able to pipe any type of content into fzf)
- [x] Automatic convert `fn_transform` to strings with `string.dump` when possible (no upvalue refs)
- [x] Add `fzf_exec/fzf_live` API testing for all use cases (string, multiprocess, table, function, etc)
- [x] Deprecate `fn_{pre|post}_fzf`
- [x] Add `opts._resume_reload` which will be used by the `hide` profile to determine if we need to reload on unhide (e.g. buffer lines changes, etc).

phanen:
`:FzfLua live_grep rg_glob=true multiprocess=false` glob don't work after https://github.com/ibhagwan/fzf-lua/commit/624ae3bf06283a1151fc5aaa7f603de6f1f8f35d.

ibhagwan:
> `:FzfLua live_grep rg_glob=true multiprocess=false` glob don't work after [624ae3b](https://github.com/ibhagwan/fzf-lua/commit/624ae3bf06283a1151fc5aaa7f603de6f1f8f35d).

Ty @phanen, I’ll take a look, still a lot more  work but I’m happy with the code cleanup.

phanen:
live_grep_st use a function rather a pre-built string:
https://github.com/ibhagwan/fzf-lua/blob/74da0da81df42d3f830919fa6cd8062a1334d676/lua/fzf-lua/providers/grep.lua#L308-L313

And its glob_parse here:
https://github.com/ibhagwan/fzf-lua/blob/74da0da81df42d3f830919fa6cd8062a1334d676/lua/fzf-lua/providers/grep.lua#L64-L66

ibhagwan:
> live_grep_st use a function rather a pre-built string:

Correct, but I have to find a generic solution for this that isn’t a hack or maybe a condition in live_grep which changes `fn_reload` based on `opts.multiprocess`?

`mutliprocess=false` is just for testing/debugging issues anyways.

phanen:
I remember I've been using multiprocess=false because I don't need file_icon/git_icon and want glob feature while `live_grep_native` don't support glob.


```diff
diff --git a/lua/fzf-lua/make_entry.lua b/lua/fzf-lua/make_entry.lua
index 41e701d..7f30605 100644
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -230,7 +230,7 @@ M.preprocess = function(opts)
 
   -- nifty hack to avoid having to double escape quotations
   -- see my comment inside 'live_grep' initial_command code
-  if opts.argv_expr then
+  if opts.argv_expr and opts.cmd then
     opts.cmd = opts.cmd:gsub("{argv.*}",
       function(x)
         local idx = x:match("{argv(.*)}")
diff --git a/lua/fzf-lua/providers/grep.lua b/lua/fzf-lua/providers/grep.lua
index c667d7b..537d459 100644
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -295,18 +295,23 @@ M.live_grep = function(opts)
   end
 
   -- signal to preprocess we are looking to replace {argvz}
-  opts.argv_expr = true
+  opts.argv_expr = opts.multiprocess
 
   -- this will be replaced by the appropriate fzf
   -- FIELD INDEX EXPRESSION by 'fzf_exec'
-  opts.cmd = get_grep_cmd(opts, core.fzf_query_placeholder, 2)
-  if not opts.cmd then return end
+  opts.cmd = opts.multiprocess and
+      get_grep_cmd(opts, opts.multiprocess and core.fzf_query_placeholder or nil, 2)
 
   -- search query in header line
   opts = core.set_title_flags(opts, { "cmd", "live" })
   opts = core.set_header(opts, opts.headers or { "actions", "cwd" })
   opts = core.set_fzf_field_index(opts)
-  core.fzf_live(opts.cmd, opts)
+  core.fzf_live(opts.cmd or function(s)
+    -- can be nil when called as fzf initial command
+    local query = s[1] or ""
+    opts.no_esc = nil
+    return get_grep_cmd(opts, query, true)
+  end, opts)
 end
 
 M.live_grep_glob = function(opts)
```

live workspace symbol also broken.
```diff
diff --git a/lua/fzf-lua/providers/lsp.lua b/lua/fzf-lua/providers/lsp.lua
index 9287a1c..b41de87 100644
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -819,9 +819,9 @@ M.live_workspace_symbols = function(opts)
   opts.lsp_params = { query = opts.lsp_query or opts.query or "" }
   opts.query = opts.lsp_query or opts.query
 
-  local contents = function(query)
-    opts.query = query
-    opts.lsp_params = { query = query or "" }
+  local contents = function(s)
+    opts.query = s[1]
+    opts.lsp_params = { query = s[1] or "" }
     opts = gen_lsp_contents(opts)
     return opts.__contents
   end
```

ibhagwan:
> I remember I've been using multiprocess=false because I don't need file_icon/git_icon and want glob feature while live_grep_native don't support glob.

You’re always at a performance disadvantage this way, why not use multiprocess=true with git_icons=false and file_icons=false?

phanen:
I don't know... `FzfLua live_grep rg_glob=true multiprocess=false` is significantly faster than `FzfLua live_grep rg_glob=true multiprocess=true`

ibhagwan:
> I don't know... `FzfLua live_grep rg_glob=true multiprocess=false` is significantly faster than `FzfLua live_grep rg_glob=true multiprocess=true`

Weird, it shouldn’t be, multiprocess should always be faster as it’s an external independent command.

ibhagwan:
> live workspace symbol also broken.

Ty for the diff fixes, I’ll add these later today.

phanen:
Damn, I forgot to disable `temp-throttle.service`. So my words may be wrong, I will test it again.

ibhagwan:
Pushed your fixes in https://github.com/ibhagwan/fzf-lua/commit/497ab31d6d1083e9d80589ae4441b6533adf9750, ty @phanen!

ibhagwan:
@phanen, fixed some more stuff, still have an issue with multiprocess=false and file_icons=mini which I’m able to solve with vim.schedule by uncommenting this line:
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/libuv.lua#L215-L216

But that presents another issue, it seems that when I use vim.scheudule there the fzf pipe closes before all the reads are complete (maybe fzf closes it with process exit?), if you run `:FzfLua live_grep multiprocess=false` and type quickly you will sometimes encounter partial/empty results, or if you run `:FzfLua files` on a large folder the last read will get cut resulting in less number of files.

Had to go out so I left it commented for now but once I solve that it would also remove the need for the `ft_match` mock as calls to `fn_transform` will be executed outside of  the “fast event”.

Once I solve this we can “retire” both of these functions (I think):
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/path.lua#L612
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/path.lua#L639

ibhagwan:
Seems the issue is the pipe gets closed before finalizing all writes so this line is hit:
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/shell.lua#L328

Probably some race condition causing this to trigger:
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/shell.lua#L315-L321

ibhagwan:
Fixed in https://github.com/ibhagwan/fzf-lua/commit/2cd15285d42987fd3389bfb19885c3a93ed15b78, finish would be called before any read were scheduled so `write_cb_count` would be 0, needed to add a `read_cb_count`.

ibhagwan:
~~Only issue left `vim.filetype.match` is failing the tests (commented for now):~~
https://github.com/ibhagwan/fzf-lua/blob/dd10409aadce6f3f2f4c8e399d5d86f315e45acc/lua/fzf-lua/devicons.lua#L518-L521

~~Obviously we can leave it as is but I’d like to find out what’s happening here as this may indicate other issues when users calling vim functions with fzf-lua’s API.~~

**Scrap this issue this was due to mini.test `child.lua_get` running in a fast event, no real upside for this change anyway.**

ibhagwan:
@phanen, check it out, I think it's somewhat close to the final version, only tasks left are to add CI testing and implement the condition in the `hide` profile so that it reloads `contents` only when it needs reloading:
> `opts.__reload_cmd` is a dud now, this needs to be changed with a new `opts.__should_reload_on_resume()` or something similar
https://github.com/ibhagwan/fzf-lua/blob/df67fc26efd3a92af22684ca9cd0a86e7a7df90f/lua/fzf-lua/profiles/hide.lua#L90-L92

phanen:
Yeah now (no icons) `rg_glob=true multiprocess=true` is faster but that's becuase `rg_glob=true multiprocess=false` is slower. I now know the reason why it's on the contrary before.

Before:
* `rg_glob=true multiprocess=false` call `libuv.spawn` with `fn_transform=nil/false` 
* `rg_glob=false multiprocess=true` always call `libuv.spawn` with `fn_transform=make_entry.file`

So they go into different branch here:
https://github.com/ibhagwan/fzf-lua/blob/47b19c7c09e1ca1c98da20d78d252a126d85d845/lua/fzf-lua/libuv.lua#L166-L170

* `rg_glob=true multiprocess=false`: A1 `read proc pipe` -> A1 `write socket` -> B1 `read socket` -> B1 `write to stdout`
* `rg_glob=true multiprocess=true`: B2 `read proc pipe` -> B2 `fn_transform` -> B2 `write to stdout`

It's not werid the former can be faster than the later.  

And now both `fn_transform` are converted from default: https://github.com/ibhagwan/fzf-lua/blob/4350a64466701d71016957d2a39aec16b2ac8170/lua/fzf-lua/defaults.lua#L534-L535

So now A1 also need a dummy `fn_transfrom`, it become slower.

phanen:
Some ideas about improvement:
* `rg_glob=true` don't require `fn_transform`, only need `fn_preprocess` to just parse the glob:
https://github.com/ibhagwan/fzf-lua/blob/01ef2bbda511c6e6c02a3c6b39c5b6fdd8e7ea7b/lua/fzf-lua/make_entry.lua#L216-L218
Although st use a different codepath from mt here:
https://github.com/ibhagwan/fzf-lua/blob/4350a64466701d71016957d2a39aec16b2ac8170/lua/fzf-lua/providers/grep.lua#L64-L66
* Not necessary to do `fn_preprocess` then `uv.spawn`, there's a option to `change:transfrom:nvim -l rpc.lua {q}` to parse the glob only then `stringify_cmd` a cmd to be executed by fzf directly. With this approach we can make live_grep (with glob parse) fast as `live_grep_native`.

ibhagwan:
> Yeah now (no icons) `rg_glob=true multiprocess=true` is faster but that's becuase `rg_glob=true multiprocess=false` is slower. I now know the reason why it's on the contrary before.
> 
> Before:
> 
> * `rg_glob=true multiprocess=false` call `libuv.spawn` with `fn_transform=nil/false`
> * `rg_glob=false multiprocess=true` always call `libuv.spawn` with `fn_transform=make_entry.file`
> 
> So they go into different branch here:
> 
> 
> * `rg_glob=true multiprocess=false`: A1 `read proc pipe` -> A1 `write socket` -> B1 `read socket` -> B1 `write to stdout`
> * `rg_glob=true multiprocess=true`: B2 `read proc pipe` -> B2 `fn_transform` -> B2 `write to stdout`
> 
> It's not werid the former can be faster than the later.
> 
> And now both `fn_transform` are converted from default:
> 
> 
> So now A1 also need a dummy `fn_transfrom`, it become slower.

You’re right, when turning off all features (icons, etc) we don’t need to call `fn_transform` which makes it slower, perhaps we can nullify `fn_transform` like we do here for _narive?
https://github.com/ibhagwan/fzf-lua/blob/4350a64466701d71016957d2a39aec16b2ac8170/lua/fzf-lua/shell.lua#L178-L193

> Not necessary to do fn_preprocess then uv.spawn, there's a option to change:transfrom:nvim -l rpc.lua {q} to parse the glob only then stringify_cmd a cmd to be executed by fzf directly. With this approach we can make live_grep (with glob parse) fast as live_grep_native.

Great idea, 2 issues:
- Needs to be limited to fzf, skim needs to use the old approach
- We will need a generic entry point for `change:…` so we can also handle `line_query=true`:
https://github.com/ibhagwan/fzf-lua/blob/4350a64466701d71016957d2a39aec16b2ac8170/lua/fzf-lua/config.lua#L891-L909

ibhagwan:
I also feel the logic of my last commit https://github.com/ibhagwan/fzf-lua/pull/2152/commits/4350a64466701d71016957d2a39aec16b2ac8170  should b changed: 
- multiprocess = true should always attempt multiprocess (since API users can also use it now)
- multiprocess = 1 means optional processing with make_entry.file/preprocess
- the logic of this should be moved to normalize_opts

This way we will restore the old behavior of multiprocess=false without icons/formatting

ibhagwan:
@phanen, https://github.com/ibhagwan/fzf-lua/pull/2152/commits/8aead20f5ba7e27e4e900c3eb94aff27a89a2380 restores the speed to:
```lua
-- debug=1 will now also print the command at the first line, same as multiprocess+debug
:FzfLua files profile=default multiprocess=false file_icons=false debug=1
```

ibhagwan:
This thing got me wondering again if we can make `mutliprocess=false` with icons/processing lag free, obviously it won’t be as fast because we chose to split callback data to lines and add icons, etc but maybe we can still make it lag free?

phanen:
One way is just not prepending icons (seems what snacks/telescope did).

Here's a example to set extmarks....
https://github.com/phanen/fzf-lua-overlay/blob/709fa6c135527ffb4de71d7a83b9e0e782c3ee38/lua/fzf-lua-extra/providers/file_decor.lua#L1

This example append icons because prepend icons will break fzf match highlighting. I'm not sure if it's solvable.

ibhagwan:
> One way is just not prepending icons (seems what snacks/telescope did).

I haven’t looked at Telescode’s code in forever but when it started it prepended icons, extmark was only added to neovim much later.


> This example append icons because prepend icons will break fzf match highlighting. I'm not sure if it's solvable.

Very nice, that will definitely simplify the code and performance but we’ll have to also find a solution for formatters, path_shorten, etc.

Path shorten for example could get very complicated as the entire line isn’t always visible in fzf - it could display …<rest of line>

ibhagwan:
> This example append icons because prepend icons will break fzf match highlighting. I'm not sure if it's solvable.

This is also a problem which turns me off the idea.

Guess we have to stay with processing, still the question remains, even if slow, can we make the experience lag free with multiprocess=false?

Won’t make a difference in files/grep/etc as nobody uses (nobody should) `multiprocess=false` but it would solve the case of an API call with `fn_process` of type “function” with upvalue refs (require, usage of local/global vars outside the function), without upvalue refs I already solved it here:
https://github.com/ibhagwan/fzf-lua/blob/f276c98dcdcec021865ae9900bf696659fa8c76a/lua/fzf-lua/shell.lua#L192-L205

ibhagwan:
@phanen, not a new issue, always been the case with `multiprocess=false` I'm just trying to revisit this as the latest changes in https://github.com/ibhagwan/fzf-lua/pull/2152/commits/bd57612619e73943e5006caa28e6ababb601639e  made this more apparent and interesting.

Something is odd and I am yet to figure it out, when running a long operation with `multiprocess=false`, as simple as `FzfLua files multiprocess=false` in a large monorepo or at `$HOME`  there's an apparent lag and fzf seems to be frozen and then updated in bulks , debugging the code it seems that there are many `shelll.on_write` calls which write the data into fzf using `uv.write` but for some reason the UI isn't getting updated after all writes are done.

Commenting `fn_transform` call makes it responsive (but I believe also updated in bulk but fast?):
https://github.com/ibhagwan/fzf-lua/blob/ce9e3714df59ecb0267d2eeaf1b43b8710c96371/lua/fzf-lua/libuv.lua#L186

You were right, it is pretty damn fast though, faster than `multiprocess=true` even with icon processing, once I solve this it might be a viable option for Windows as this avoids spawning cmd.exe which is rather slow.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2152/commits/6af4ff02c720a17038459259466032645daf6254

@phanen, I updated libuv.spawn again, added queue borrowed from snacks.nvim, improves a bit but still feels laggy with bulk updates when using `multiprocess=false`, queue is enabled by default try both and see the difference:
```lua
-- Long lag but super fast, all entries appear at once
:FzfLua files multiprocess=false use_queue=false
-- Using queue, more responsive but still laggy with "bulk" updates
:FzfLua files multiprocess=false
-- Multi process (implied true), zero lag, perhaps a bit slower?
:FzfLua files
```

ibhagwan:
@phanen, will you try out my latest commit and lmk what you think? I added a few options but I'm still mighty confused why it's not freeing the lag fully with `multiprocess=false`, even when using a queue and doing the writes serially with a coroutine if I uncomment one of the sleeps below the UI will freeze until it's over, not sure why that is, why does fzf UI refuse to update when the CPU is idle?


```lua
-- uses queue and throttling by default
:FzfLua files multiprocess=false use_queue=nil throttle=nil
-- queue without throttle, not sure if the responsiveness is better
:FzfLua files multiprocess=false use_queue=nil throttle=false
-- no queue with throttle 
:FzfLua files multiprocess=false use_queue=false throttle=nil
-- no queue, no throttle, fastest but freezes the UI until all results are in
:FzfLua files multiprocess=false use_queue=false throttle=false
```

Try uncommenting either of these lines, and you'll see what I mean:
https://github.com/ibhagwan/fzf-lua/blob/c3c44d477e13e24f6aa51693d3a842a13cbe9bfa/lua/fzf-lua/shell.lua#L355-L356
https://github.com/ibhagwan/fzf-lua/blob/c3c44d477e13e24f6aa51693d3a842a13cbe9bfa/lua/fzf-lua/shell.lua#L358-L361

I tried queueing the `uv.write` calls in threads using  `uv.new_work|uv.work_queue` and still nada, do you have any idea?

Overall I find the performance of `multiprocess=false` acceptable now (seems much better) but I would prefer to solve this mystery.

ibhagwan:
Just had a shower thought, I’ve been barking up the wrong tree :)

The issue comes from the spawning process code  ( `nvim -l` ) code, I’ll try to thread/coroutinify the below:
https://github.com/ibhagwan/fzf-lua/blob/7e700679918f173d698c60035e48ea5d7a2fb6a4/lua/fzf-lua/rpc.lua#L40-L53

ibhagwan:
It seems the issue is from the blocking `vim.rpcrequest`:
https://github.com/ibhagwan/fzf-lua/blob/7e700679918f173d698c60035e48ea5d7a2fb6a4/lua/fzf-lua/rpc.lua#L61

I tried to `coroutine.wrap` it with no success, I'm trying to mult-ithread it with `uv.new_thread` but it seems that `vim.rpcrequest` and `vim.fn` are both nil in the new thread (no runtime?).

ibhagwan:
This is even more strange, I tried running the  neovim proxy `uv.listen` in a thread, didn't work for some reason so I made `rpc_nvim_exec_lua` run in it's own process:
1. The main instance
2. The shell command instance (nvim -l) acting as proxy between fzf and the main instance (pipe)
3. Another process executed by (2) using `jobstart` which does a single `vim.rpcrequest`  and still the same behavior

This makes me think having `vim.rcprequest` called in the main instance is the issue, tried to chase a lead with using pure lua msgpack-rpc instead of `vim.rpcrequest` but I'm not sure if that would help.

ibhagwan:
Alright, after a few days of messing with this I’m giving up on the idea to make multiprocess=false a lag free experience as it seems the issue stems from neovim being single threaded, so without using threads (which have no neovim runtime) it’s never going to be as smooth as multiprocess=true (which most pickers should use when processing shell output anyway).

Back to the original mission, 2 tasks left.

ibhagwan:
@phanen, checkout https://github.com/ibhagwan/fzf-lua/pull/2152/commits/32e19e0896458450898bcc910a4b9ab23323b844 proof of concept for combining pickers,  suggestions/improvements  welcome :-)

phanen:
there's a bug(maybe blame to my commit). loop: `ctrl-c` -> `FzfLua live_grep` -> `ctrl-c` -> `FzfLua resume`. no result, sometime error

ibhagwan:
> there's a bug(maybe blame to my commit). loop: `ctrl-c` -> `FzfLua live_grep` -> `ctrl-c` -> `FzfLua resume`. no result, sometime error

I’ll check, I have yet to add testing for fzf_live, once we figure this out we should a test case for this too.

ibhagwan:
@phanen, affer enough ctrl-c <-> resume I’m getting this asset:
https://github.com/ibhagwan/fzf-lua/blob/32e19e0896458450898bcc910a4b9ab23323b844/lua/fzf-lua/shell.lua#L34

Obviously with resume, clear protected is never called but seems that new functions are registered each call until the ring buffer is full, I need to find a new solutions for this, I don’t like the “protected” idea but I also don’t want a really large buffer limit occupying a lot of memory, need to thing of something else.

ibhagwan:
Not fully fixed, need to check why functions are re-registered, but fixed the resume, deprecated fzf < 0.35, using `start` event instead of modifying the initial query, that's what I used for my shell `ctrl-f` shortcut too, it's much more robust:
https://github.com/ibhagwan/fzf-lua/blob/c8b53610c868ead5b8c4057711d3e79ddf385a17/lua/fzf-lua/core.lua#L1230-L1240

ibhagwan:
The issue was that all calls to stringify were protected indiscriminately, my last commit is still bugged and will also bug out on skim as there's no reload event so the actions may get overriden for example with `git_status`, I need to think of a better solution for storing and clearing callback function pointers.

https://github.com/ibhagwan/fzf-lua/pull/2152/commits/ee7b523c066c99d56fe1ca0e7a17953b4eb9bcbd

phanen:
After cc480bf3adc38c0af827d752d0b2bbcc8d36fbb8, `actions.start` cannot be used in custom picker.
e.g. https://github.com/phanen/fzf-lua-overlay/blob/77c6bc0d6876e5fa75888ec9d326b4743655f32d/lua/fzf-lua-extra/providers/runtime.lua#L17-L21

(Btw we can have an option for fzf_live's fzf_field_index, https://github.com/ibhagwan/fzf-lua/blob/c7780cb18e438aa27a3f48d40559bc6f4f8f3e0d/lua/fzf-lua/core.lua#L188)

ibhagwan:
> After [cc480bf](https://github.com/ibhagwan/fzf-lua/commit/cc480bf3adc38c0af827d752d0b2bbcc8d36fbb8), `actions.start` cannot be used in custom picker. e.g. https://github.com/phanen/fzf-lua-overlay/blob/77c6bc0d6876e5fa75888ec9d326b4743655f32d/lua/fzf-lua-extra/providers/runtime.lua#L17-L21
> 
> (Btw we can have an option for fzf_live's fzf_field_index,
> 
> https://github.com/ibhagwan/fzf-lua/blob/c7780cb18e438aa27a3f48d40559bc6f4f8f3e0d/lua/fzf-lua/core.lua#L188
> 
> )

Technically all the fzf events working through `actions` was an unintended consequence (should be in keymap.fzf.start),  let me think how can I fix it while still keeping the refactor - suggestions welcome.

The reason I moved the function convert there is so we can store the converted opts for resume and not re-register the action id as “protected” on resume.

ibhagwan:
> After [cc480bf](https://github.com/ibhagwan/fzf-lua/commit/cc480bf3adc38c0af827d752d0b2bbcc8d36fbb8), actions.start cannot be used in custom picker. e.g. https://github.com/phanen/fzf-lua-overlay/blob/77c6bc0d6876e5fa75888ec9d326b4743655f32d/lua/fzf-lua-extra/providers/runtime.lua#L17-L21

I wonder why it doesn’t work, I’ll have to check, the call to convert should work with a custom picker too:

> fzf_live sends true in convert_actions
https://github.com/ibhagwan/fzf-lua/blob/c7780cb18e438aa27a3f48d40559bc6f4f8f3e0d/lua/fzf-lua/core.lua#L192

> Which calls convert exec silent:
https://github.com/ibhagwan/fzf-lua/blob/c7780cb18e438aa27a3f48d40559bc6f4f8f3e0d/lua/fzf-lua/core.lua#L215-L218

phanen:
This work for me
```diff
diff --git a/lua/fzf-lua/core.lua b/lua/fzf-lua/core.lua
index f718696..97a6d84 100644
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -1231,7 +1231,7 @@ M.setup_fzf_interactive_flags = function(command, fzf_field_index, opts)
       string.format("change:reload:%s%s", no_query_condi, reload_command)))
     if utils.has(opts, "fzf", { 0, 35 }) then
       opts._fzf_cli_args = opts._fzf_cli_args .. string.format(" --bind=%s",
-        libuv.shellescape(string.format("start:reload:%s%s", no_query_condi, reload_command)))
+        libuv.shellescape(string.format("start:+reload:%s%s", no_query_condi, reload_command)))
     end
   end
 
```

ibhagwan:
Im too deprived of sleep lol, just realized you’re using fzf_live which now makes use of the start event too:
https://github.com/ibhagwan/fzf-lua/blob/c7780cb18e438aa27a3f48d40559bc6f4f8f3e0d/lua/fzf-lua/core.lua#L1233-L1234

Now this comment makes sense (which is the right solution):
> (Btw we can have an option for fzf_live's fzf_field_index,

Another (breaking) change is that now the callback param is an array (I updated the wiki), so the callbacks should be:
```lua
require 'fzf-lua'.fzf_live(
  function(args)
    local q = args[1]
   …
```

ibhagwan:
> This work for me
> 
> ```diff
> diff --git a/lua/fzf-lua/core.lua b/lua/fzf-lua/core.lua
> index f718696..97a6d84 100644
> --- a/lua/fzf-lua/core.lua
> +++ b/lua/fzf-lua/core.lua
> @@ -1231,7 +1231,7 @@ M.setup_fzf_interactive_flags = function(command, fzf_field_index, opts)
>        string.format("change:reload:%s%s", no_query_condi, reload_command)))
>      if utils.has(opts, "fzf", { 0, 35 }) then
>        opts._fzf_cli_args = opts._fzf_cli_args .. string.format(" --bind=%s",
> -        libuv.shellescape(string.format("start:reload:%s%s", no_query_condi, reload_command)))
> +        libuv.shellescape(string.format("start:+reload:%s%s", no_query_condi, reload_command)))
>      end
>    end
>  
> ```

This is a great fix! we should implement this in the convert actions too:

https://github.com/ibhagwan/fzf-lua/blob/f972ad787ee8d3646d30000a0652e9b168a90840/lua/fzf-lua/core.lua#L1255-L1269



The only thing I can’t recall is if this `start:+…` logic new in fzf?

Do you recall which fzf version added this logic so we can limit it (if it requires > 0.36)?

ibhagwan:
Apparently the + logic in binds existed since fzf 0.20, I’ll make the change.

ibhagwan:
@phanen, https://github.com/ibhagwan/fzf-lua/pull/2152/commits/3d73bf5ce9cc3dc0e358534a6dd4ec80f77d2387.

ibhagwan:
Should also change `line_query` and hide profile events to fzf_cli_args with + so we don’t override user events:
https://github.com/ibhagwan/fzf-lua/blob/3d73bf5ce9cc3dc0e358534a6dd4ec80f77d2387/lua/fzf-lua/config.lua#L924

phanen:
Even more..
```diff
diff --git a/lua/fzf-lua/config.lua b/lua/fzf-lua/config.lua
index 1225b88..74bb403 100644
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -922,7 +922,7 @@ function M.normalize_opts(opts, globals, __resume_key)
   elseif opts.line_query then
     utils.map_set(opts, "winopts.preview.winopts.cursorline", true)
     utils.map_set(opts, "keymap.fzf.change",
-      "transform:" .. FzfLua.shell.stringify_data(function(q, _, _)
+      "transform:+" .. FzfLua.shell.stringify_data(function(q, _, _)
         local lnum = q[1]:match(":(%d+)$")
         local new_q, subs = q[1]:gsub(":%d*$", "")
         -- No subs made, no ":" at end of string, do nothing
diff --git a/lua/fzf-lua/previewer/builtin.lua b/lua/fzf-lua/previewer/builtin.lua
index 2641019..c8d97e7 100644
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -469,7 +469,7 @@ function Previewer.base:zero(_)
   -- https://github.com/junegunn/fzf/issues/3516
   --
   self._zero_lock = self._zero_lock or path.normalize(vim.fn.tempname())
-  local act = string.format("execute-silent(mkdir %s && %s)",
+  local act = string.format("+execute-silent(mkdir %s && %s)",
     libuv.shellescape(self._zero_lock),
     shell.stringify_data(function(_, _, _)
       vim.defer_fn(function()
diff --git a/lua/fzf-lua/profiles/hide.lua b/lua/fzf-lua/profiles/hide.lua
index 635ddcc..fcd5068 100644
--- a/lua/fzf-lua/profiles/hide.lua
+++ b/lua/fzf-lua/profiles/hide.lua
@@ -84,7 +84,7 @@ return {
         return act
       end, opts.actions)
       -- Hijack the resize event to reload buffer/tab list on unhide
-      opts.keymap.fzf.resize = "transform:" .. shell.stringify_data(function(_, _, _)
+      opts.keymap.fzf.resize = "transform:+" .. shell.stringify_data(function(_, _, _)
         if opts._unhide_called then
           opts._unhide_called = nil
           if opts.__reload_cmd then
diff --git a/lua/fzf-lua/providers/buffers.lua b/lua/fzf-lua/providers/buffers.lua
index 4017991..733621d 100644
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -384,7 +384,7 @@ M.tabs = function(opts)
   if opts.locate and utils.has(opts, "fzf", { 0, 36 }) then
     -- Set cursor to current buffer
     utils.map_set(opts, "keymap.fzf.load",
-      "transform:" .. FzfLua.shell.stringify_data(function(_, _, _)
+      "transform:+" .. FzfLua.shell.stringify_data(function(_, _, _)
         local pos = 0
         for tabnr, tabh in ipairs(vim.api.nvim_list_tabpages()) do
           pos = pos + 1
```

ibhagwan:
> Even more..

It is better to also move these to fzf_cli_args to nor override the users keymap.fzf.

ibhagwan:
I’ll only get to this later today as AFK now, if you need it sooner feel free to PR.

ibhagwan:
@phanen, check out https://github.com/ibhagwan/fzf-lua/commit/9f6c1a09d359ee0570750a8cde20abc517b5e383, refactored all internal events to use `_fzf_cli_args` so `keymap.fzf.<event>` from the user is never overwritten, also added a test  for  events + fzf index:
https://github.com/ibhagwan/fzf-lua/blob/9f6c1a09d359ee0570750a8cde20abc517b5e383/tests/api_spec.lua#L177-L213

For some odd reason the GC called test now fails, I can't test it out ATM as I have to leave, might be related to this change but I'm not sure:
https://github.com/ibhagwan/fzf-lua/blob/9f6c1a09d359ee0570750a8cde20abc517b5e383/lua/fzf-lua/test/helpers.lua#L142-L145

ibhagwan:
@phanen, almost there, to fix the gc collection I had to change to opening files without a previewer, for some odd reason when using the builtin previewer the gc isn't called?
https://github.com/ibhagwan/fzf-lua/blob/2f87829214e577728dde85ecc7384c8bb54ac219/tests/win_spec.lua#L49

The api event test is also failing on Windows, unsure why will have to test further, tricky as I can only test via the CI as I don't have access to a Windows VM anymore...

ibhagwan:
@phanen, can you take a look at this run?
https://github.com/ibhagwan/fzf-lua/actions/runs/16281474523/job/45971839527

Aside from the obvious failure of the new test on Windows something wild is happening, even though `new_child` is called in ui_spec, all tests fail due to having 2 more entries which are the `$FZF_PROMPT` and `$FZF_TOTAL_COUNT`, I still haven’t figured this out, commenting the new api/event test will make ui_spec pass on windows.

It seems as if the start/load event somehow become persistent after the api test:
https://github.com/ibhagwan/fzf-lua/blob/a54935ba2a7555a0cba3236d307b84c63bae6984/tests/api_spec.lua#L192-L206

 May or may not be related, this is also a mystery, defining the load event in keymap.fzf.load vs _fzf_cli_args makes the GC test fail with the builtin previewer:
https://github.com/ibhagwan/fzf-lua/blob/a54935ba2a7555a0cba3236d307b84c63bae6984/tests/win_spec.lua#L48-L50

The offending change:
https://github.com/ibhagwan/fzf-lua/blob/a54935ba2a7555a0cba3236d307b84c63bae6984/lua/fzf-lua/test/helpers.lua#L142-L147

phanen:
ok i will look into it when i have time
iirc that is because some function in registry is never clean

phanen:
I walk the `_G` table with these two line, some `_registry` is always keeped in entry: https://github.com/ibhagwan/fzf-lua/blob/454855f4c2db0791692716f0c4fcc6ea869574a1/lua/fzf-lua/core.lua#L371-L378


https://github.com/phanen/fzf-lua/commit/54a662339253b85ff81e08009100cbbe0945c1ad

Reference chain:
```
.                    = (table: 0x7f8f60720d78)
.[FzfLua]            = (table: 0x7f8f60241998)
.[shell]             = (table: 0x7f8f60246cf8)
.[register_func]     = (function: 0x7f8f601eca10)
u[_registry]         = (table: 0x7f8f60246d40)
.[2]                 = (function: 0x7f8f600ec1b0)
u[fn]                = (function: 0x7f8f600ec0f0)
u[contents]          = (function: 0x7f8f600ebf58)
u[fn]                = (function: 0x7f8f600ebee8)
u[self]              = (Object)
.[win]               = (table: 0x7f8f601359b0)
```

Still not sure why this line matter here: https://github.com/ibhagwan/fzf-lua/blob/454855f4c2db0791692716f0c4fcc6ea869574a1/lua/fzf-lua/test/helpers.lua#L143

phanen:
I believe this related to the side effect of something:
```diff
diff --git a/lua/fzf-lua/core.lua b/lua/fzf-lua/core.lua
index 6e05507..defb29b 100644
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -592,6 +592,9 @@ M.create_fzf_colors = function(opts)
 end
 
 M.create_fzf_binds = function(opts)
+  do
+    local _ = "execute-silent:" .. shell.stringify_data(function() end, {})
+  end
   local binds = opts.keymap.fzf
   if not binds or utils.tbl_isempty(binds) then return {} end
   local combine, separate = {}, {}
diff --git a/lua/fzf-lua/test/helpers.lua b/lua/fzf-lua/test/helpers.lua
index a6c8478..5c4ccfa 100644
--- a/lua/fzf-lua/test/helpers.lua
+++ b/lua/fzf-lua/test/helpers.lua
@@ -132,7 +132,10 @@ M.new_child_neovim = function()
       require("fzf-lua").setup(vim.tbl_deep_extend("keep", ..., {
         %s
         winopts = {
-          on_create = function() _G._fzf_lua_on_create = true end,
+          on_create = function()
+            _G._fzf_lua_on_create = true
+            vim.defer_fn(function() _G._fzf_load_called = true end, 20)
+          end,
           on_close = function()
             _G._fzf_lua_on_create = nil
             _G._fzf_postprocess_called = nil
@@ -141,10 +144,10 @@ M.new_child_neovim = function()
         },
         -- TODO: why is gc called with builtin with this one?
         -- keymap = { fzf = { true, load = function() _G._fzf_load_called = true end } }
-        fzf_cli_args = "--bind=" .. FzfLua.libuv.shellescape("load:+execute-silent:"
-          .. FzfLua.shell.stringify_data(function(_, _, _)
-              _G._fzf_load_called = true
-            end, {}))
+        -- fzf_cli_args = "--bind=" .. FzfLua.libuv.shellescape("load:+execute-silent:"
+        --   .. FzfLua.shell.stringify_data(function(_, _, _)
+        --       _G._fzf_load_called = true
+        --     end, {}))
       }))
     ]])
         -- using "FZF_DEFAULT_OPTS" hangs the command on the
diff --git a/tests/win_spec.lua b/tests/win_spec.lua
index 322dea3..b193174 100644
--- a/tests/win_spec.lua
+++ b/tests/win_spec.lua
@@ -46,8 +46,8 @@ T["win"]["hide"]["ensure gc called after win hidden (#1782)"] = function()
       end
     end
     -- TODO: why isn't gc called with the builtin previewr
-    -- child.lua([[FzfLua.files{ previewer = 'builtin' }]])
-    child.lua([[FzfLua.files{ previewer = false }]])
+    child.lua([[FzfLua.files{ previewer = 'builtin' }]])
+    -- child.lua([[FzfLua.files{ previewer = false }]])
     child.wait_until(function()
       return child.lua_get([[_G._fzf_load_called]]) == true
     end)
```
apply this on this branch, an no-op stringify_data resolve the test.

ibhagwan:
> I walk the _G table with these two line, some _registry is always keeped in entry:

The registry is never cleared, even after `clear_protected` is called, it’s a ring buffer, function pointers (which hold ref to `opts`) only leave the registry after a full circle of the registry when a new entry overrides the same index.

> Still not sure why this line matter here:

This got me too, I looked at the generated command with debug=true, the only difference was the order of the callback function ids, for example in one the preview callback can be id 3 and in the other command (with _fzf_cli_args), it’s 1, otherwise the command is the same.

Another interesting fact is that I wasn’t able to reproduce this outside of testing, if I paste the same testing code into a lua file and `:source` it, I am only able to get the gc to be called with `previewer=false`, with “builtin” the gc is never called, with or  without `keymap.fzf`.

ibhagwan:
Interesting @phanen, what’s this for? Just to cycle the registry to skip 1?
```lua
do  
  local _ = "execute-silent:" .. shell.stringify_data(function() end, {})
end
```

ibhagwan:
Interesting, changing the `do end` block to the below also makes the CI pass, as I suspected it's skipping a slot in the registry which makes the difference... wild stuff
```lua
  do
    local _ = shell.register_func(function() end)
  end
```

phanen:
I consider the parity of regostry MAX_LEN... but this seems irrelated.

Also need `defer_fn` replace the action... otherwise test still fail. 😢

ibhagwan:
> I consider the parity of regostry MAX_LEN... but this seems irrelated.
> 
> Also need `defer_fn` replace the action... otherwise test still fail. 😢

I found out mini.test runs the test many times in the same child causing the registry to overflow and hence throw away the function ptr which causes the test to succeed, need another `shell.clear_registry()` which empties the registry, after which calling `collectgarbage` triggers the gc clear, it's working for me outside of the testing (in a simple lua source file)

Try sourcing this file:
```lua
require("fzf-lua")
FzfLua.setup({})
FzfLua.utils.setmetatable__gc = function(t, mt)
  local prox = newproxy(true)
  print("setting __gc proxy")
  getmetatable(prox).__gc = function()
    print("__gc called")
    mt.__gc(t)
  end
  t[prox] = true
  return setmetatable(t, mt)
end
FzfLua.files({
  previewer = "builtin",
  keymap = { fzf = { load = function() print("load called (keymap)") end } },
  -- actions = { load = { fn = function() print("load called (actions)") end, exec_silent = true, } },
})
```

Add this to `shell.lua`:
```lua
function M.clear_registry()
  _index = 0
  _registry = {}
  _protected = {}
end
```

After hide call:
```lua
:lua FzfLua.shell.clear_registry()
:lua collectgarbage("collect")
```

And you'll see the message `__gc called`.

Still figuring out why it's not working in the test even after adding `clear_registry()`.

ibhagwan:
mini.test wasn't running fzf-lua many times, it was the fact you called `FzfLua.files` inside the `child..wait_until`, that was opening and hiding fzf-lua many times thus cycling the function ids until one of the original ones would be kicked out of the ring buffer and then `collectgarbage` would dispose of it.

Trying a different strategy with a single run and `clear_registry`.

phanen:
yeah but better handle it in the api rather than the test if user shouldn't take care of `clear_registry()`.

ibhagwan:
> yeah but better handle it in the api rather than the test if user shouldn't take care of `clear_registry()`.

For sure, but this is just a test, technically we can replace `clear_protected` with `clear_registry`, unless we are resuming or unhiding I don't see any reason to keep any function pointers in the registry at all.

ibhagwan:
See https://github.com/ibhagwan/fzf-lua/pull/2152/commits/5c37803b1b42813733904deb01b488d6dceade28, yes `clear_registry` is better automated but I can't think where is the right place to put it, for example in testing since we use `_fzf_cli_args` in setup to trigger the load event calling `clear_registry` removes that pointer and messes up the test, hence:
https://github.com/ibhagwan/fzf-lua/blob/5c37803b1b42813733904deb01b488d6dceade28/tests/win_spec.lua#L60-L67

Now the only mystery left is why the api events test fails and why does it mess up all the other tests on Windows by adding `FZF_PROMPT/FZF_TOTAL_COUNT` to every single command output...

ibhagwan:
I’m thinking the best way forward would be to modify the function registry, ditch the “protected” idea, and make it an MRU store, once the store is full every new register would kick out the LRU function ptr.

As long as we have enough space in the store for one picker and its functions should never be an issue.

phanen:
Yeah. I don't see how protected is used... even remove this table all test passed... 


https://github.com/ibhagwan/fzf-lua/blob/5c37803b1b42813733904deb01b488d6dceade28/lua/fzf-lua/test/helpers.lua#L142-L147

* `fzf_cli_args` use once registered function will be clear after ring overflow. (no matter protected or not since each `fzf_exec` clear the protect)
* `keymap.fzf` is evaluated each time.
* `previewer="builtin"` means we have register function (while `previewer=false` means we won't register any function ... if we don't define other action)
* `wait_until` can simply timeout (false positive)

I believe there's no problem of previous gc test... if we increase the timeout and register at least one function to ensure the queue is running each fzf_exec ... (**EDIT**: also need to figure out a way to protect load function from cleared...)

ibhagwan:
> Yeah. I don't see how protected is used... even remove this table all test passed...
> fzf_cli_args use once registered function will be clear after ring overflow. (no matter protected or not since each fzf_exec clear the protect)

Correct, that’s why I will change it to an MRU cache, every time a function is accessed it moves to the top of the cache, when we run out of space in the store we remove the bottom (LRU) - should solve everything and no more protected.

> I believe there's no problem of previous gc test... if we increase the timeout and register at least one function to ensure the queue is running each fzf_exec ...

But why? Isn’t this test better? It simulates hide + registered functions rotation + new picker followed by correct garbage collection of the original “hidden” window - what would be the benefit of the previous test? It’s just longer and avoids the call to `clear_registry` (which I don’t mind being a “test only” API).

phanen:
I'm not sure I get it... if we only register temporary function, LRU will be naive as fifo:

Maybe you want to "access" when we call function? e.g.
```diff
diff --git a/lua/fzf-lua/shell.lua b/lua/fzf-lua/shell.lua
index dfd850a..34ef697 100644
--- a/lua/fzf-lua/shell.lua
+++ b/lua/fzf-lua/shell.lua
@@ -17,7 +17,7 @@ local _index = 0
 local _registry = {}
 local _protected = {}
 
-function M.register_func(fn)
+function M.lru_put(fn)
   repeat
     _index = _index % _MAX_LEN + 1
   until not _protected[_index]
@@ -59,7 +59,8 @@ end
 function M.pipe_wrap_fn(fn, fzf_field_index, debug)
   fzf_field_index = fzf_field_index or "{+}"
 
-  local receiving_function = function(pipe_path, ...)
+  local receiving_function
+  receiving_function = function(pipe_path, ...)
     local pipe = uv.new_pipe(false)
     local args = { ... }
     -- unescape double backslashes on windows
@@ -77,6 +78,7 @@ function M.pipe_wrap_fn(fn, fzf_field_index, debug)
           module.__INFO or {}, { selected = args[1][1] })
       end
     end)
+    M.lru_put(receiving_function)
     uv.pipe_connect(pipe, pipe_path, function(err)
       if err then
         error(string.format("pipe_connect(%s) failed with error: %s", pipe_path, err))
@@ -88,7 +90,7 @@ function M.pipe_wrap_fn(fn, fzf_field_index, debug)
     end)
   end
 
-  local id = M.register_func(receiving_function)
+  local id = M.lru_put(receiving_function)
 
   -- this is for windows WSL and AppImage users, their nvim path isn't just
   -- 'nvim', it can be something else
```

But I forgot why protected is needed?
`clear_protected` only called when we start a new picker.
`set_protect` only to avoid being wiped becuase call `register_func` within the picker's lifetime. If we use LRU, `register_func` new function will still wipe old one.

The problem is not to delete the non-frequent one but to avoid to delete unexpectedly.

phanen:
maybe u r right
the problem is the timing of clear_registry , clear_protect
if it can not be determine easily

use lru seems ok

one more concern... lru feels non-determintic anther way is keep a pattern to never call shell.stringify outside of api

ibhagwan:
> Maybe you want to "access" when we call function? e.g.

Yes, that’s what I meant, when we “get” a function we bubble it to the top, so functions that are used will be cleared last.

> But I forgot why protected is needed?
> clear_protected only called when we start a new picker.

It’s not needed anymore, sometime in the past before reload I needed a way to tell actions that can be called multiple times to never be cleared out of the ring buffer, for example git_status stage/unstage - but due to logic changes there is no such issue.

> set_protect only to avoid being wiped becuase call register_func within the picker's lifetime. If we use LRU, register_func new function will still wipe old one.

If we have enough room in the cache and we “access” on get then it won’t be deleted as it will be at the top of the cache.

> The problem is not to delete the non-frequent one but to avoid to delete unexpectedly.

Im trying to think when do we have a case of a non frequent function that should still be kept, maybe if the user has a custom action for a picker in `setup` and then doesn’t use the picker in a while?


> the problem is the timing of clear_registry , clear_protect if it can not be determine easily

Im not sure if it can be determined, since we have unhide/resume - we can never know if the user decided to resume a picker or start a new one.

Only if we store the function IDs of function pointers for each picker and clear only these ones? Sounds a bit over complicated.


> one more concern... lru feels non-determintic anther way is keep a pattern to never call shell.stringify outside of api

Maybe something in between? Allow it to be called but have some known limitations?


Bottom line, IMHO these are all edge cases that will only be encountered by the very sophisticated user customizing fzf-lua like what you did with fzf-lua-extra, in such case we can always guide the user to a better way of writing the code so I think LRU cache is the way to go here.

ibhagwan:
@phanen, https://github.com/ibhagwan/fzf-lua/pull/2152/commits/4ce85af7404e042e3ad7c6fa9a44b2108497c9e0.

Seems to work great but the GC test fails on Windows now, unsure why, shouldn't be a difference as all refs are clear, also unsure why this part?

I'm assuming it's leftover from when it was encapsulated in `child.wait_until` and the Windows pickers weren't closing fast enough,

https://github.com/ibhagwan/fzf-lua/blob/4ce85af7404e042e3ad7c6fa9a44b2108497c9e0/tests/win_spec.lua#L39-L46

Also, any ideas about how to fix the api events test on Windows and the weirdness that comes with it, if you look at the failed screenshots it somehow pollutes all other ui_spec tests with the output of the FZF_PROMPT / FZF_TOTAL COUNT env vars

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2152/commits/4ee6d64276e176a7317cab2b7480e1755783fac9

Alright, so I gave up on the Windows oddities with `%FZF_PROMPT%` messing up everything, instead on Windows we use `{fzf:prompt}`, it is deprecated but still working, at least we know the actions/execute-silent combo works:
https://github.com/ibhagwan/fzf-lua/blob/4ee6d64276e176a7317cab2b7480e1755783fac9/tests/api_spec.lua#L191-L202

I still don't know I'munable to use `%FZF_PROMPT%` on Windows, might be an fzf, fzf-lua or maybe just CI/mini.test bug? I can't test it as I no longer have access to Windows VM.

I restored the GC test to it's original code with a slight hack:
https://github.com/ibhagwan/fzf-lua/blob/4ee6d64276e176a7317cab2b7480e1755783fac9/tests/win_spec.lua#L39-L42

2 more tasks and this PR might be ready to merge:
- Get rid of `fn_pre_fzf`, `fn_post_fzf`, shouldn't be a big deal as it's used by only a few pickers
- Implement `reload_on_hide()` where it makes sense:
https://github.com/ibhagwan/fzf-lua/blob/4ee6d64276e176a7317cab2b7480e1755783fac9/lua/fzf-lua/profiles/hide.lua#L92-L97

phanen:
for gc i only know windows has a upstream bug it must manually kill the job 

not sure if %FZF_PROMPT% can be used... it's env var but i didn't see this in fzf's test case

ibhagwan:
Ok, with https://github.com/ibhagwan/fzf-lua/pull/2152/commits/787139925f39e1e17e561d68f0d1475b93c6b342 I feel this PR is almost done, could probably have a bit more polish here and there but most of what I wanted to achieve is complete.

I didn’t fully implement `_resume_reload` on all pickers where it matters and also a bit lazy on `blines` reloading indiscriminately (vs only when buffer changed) but the framework is there and it can easily be added at any time, any new picker that needs reloading we just add `_resume_reload = true` in its default options or for more complex scenarios `_resume_reload = function() … end`.

There are also obvious improvements to the `:FzfLua combine` picker which I snuck in this PR - all can be handled later.

We’ll give it a few days, see if we can think of some other changes or something will come up in testing.

ibhagwan:
@phanen, how do you feel about merging this?

ibhagwan:
Lol, review by Grok4, a bit repetitive but mostly on point:

Overview of the Pull Request
This PR (#2152) in the ibhagwan/fzf-lua repository is a significant refactoring effort aimed at simplifying the codebase for fzf-lua, a Neovim plugin that integrates the fzf fuzzy finder. The core change is to convert all content types (e.g., tables, functions) passed to fzf_exec into shell command strings. This allows all picker inputs to be treated uniformly as shell commands, making it easier to use fzf features like reload and execute-silent. It also enables new functionality, such as combining multiple pickers using semicolons (;) or logical AND (&&), e.g., :FzfLua combine pickers=buffers,files.
The PR includes 24 commits (with multiple force-pushes), resulting in +2,565 lines added and -1,710 removed. It fixes issue #2058 (related to global search behavior) and lays groundwork for #1879 (VS Code-inspired search features like @, #, $). It bumps the minimum fzf version requirement from >0.25 to >0.36 (released in Jan 2023), deprecating support for older versions. The changes touch configuration, core logic, providers, utilities, and tests, with a focus on performance, async handling, and code cleanup.
Based on the PR description, comments, and code diffs, the changes appear well-motivated and progressively refined through iterative fixes. However, I’ll highlight potential flaws, bugs, or issues below after summarizing the key changes.
Key Changes and Refactoring
The diffs span multiple files, primarily in the lua/fzf-lua/ directory. Here’s a consolidated summary from the extracted diffs:
	•	CI and Config Files:
	◦	.github/workflows/ci.yaml: Updated Neovim to v0.11.2 and fzf to v0.64.0 in CI setup for better compatibility testing.
	◦	.luarc.jsonc: Added testing globals (describe, it) to diagnostics for improved Lua LSP support.
	•	Documentation:
	◦	README.md: Updated fzf minimum version and fixed a config example (only_valid → valid_only in quickfix previewer).
	◦	doc/fzf-lua.txt: Minor formatting updates (e.g., file end markers), no substantive changes.
	•	Actions and Completion Logic (lua/fzf-lua/actions.lua, lua/fzf-lua/complete.lua):
	◦	New function list_del added to delete entries from quickfix/location lists based on selections (using buffer numbers).
	◦	Updated grep_lgrep to use multiprocess (typo fixed from multiprcess) for handling globs in grep commands.
	◦	Removed core.mt_cmd_wrapper calls in completion functions (path, file), directly passing opts.cmd to core.fzf_exec for simplification.
	•	Configuration Normalization (lua/fzf-lua/config.lua):
	◦	Extensive refactoring in normalize_opts: Consolidated option handling, added support for new features like picker combining.
	◦	Deprecated old keys (e.g., fn_pre_fzf, fn_post_fzf) and removed related hooks.
	◦	Added defaults for new options like _resume_reload for hide/unhide behavior.
	•	Core Logic (lua/fzf-lua/core.lua):
	◦	Major refactoring: fzf_exec now stringifies all inputs (tables/functions to shell cmds). Removed wrappers like mt_cmd_wrapper.
	◦	Added stringify for content conversion, enabling piping of any content type into fzf.
	◦	Introduced fzf_live for live searching, with better handling of multiprocess modes.
	◦	Deprecated fn_pre/post_fzf and related functions; replaced with string-based transformations.
	◦	Bug fixes for async reading/writing, including counters for read/write callbacks to prevent pipe closure races.
	•	Providers and Pickers (e.g., lua/fzf-lua/providers/buffers.lua, lua/fzf-lua/providers/files.lua, lua/fzf-lua/providers/grep.lua):
	◦	Unified input handling to strings across pickers (e.g., buffers, files, grep, lsp).
	◦	Added support for combining pickers; first picker’s opts determine settings (window, previewer).
	◦	Fixes for glob patterns in live_grep (e.g., when rg_glob=true multiprocess=false).
	◦	Performance tweaks: Better handling of icons (file/git) in non-multiprocess modes, using vim.schedule to avoid yielding across C boundaries.
	◦	Workspace symbol fixes for live updates.
	•	Shell and Utilities (lua/fzf-lua/shell.lua, lua/fzf-lua/utils.lua):
	◦	New stringify function to convert functions/tables to dumpable strings (using string.dump where possible, no upvalues).
	◦	Removed raw_action and raw_preview_action_cmd; replaced with string-based piping.
	◦	Added resume/reload logic for dynamic content (e.g., buffer changes on unhide).
	•	Window and UI (lua/fzf-lua/win.lua):
	◦	Updated keymap handling (e.g., toggle_help with format_bind fixes).
	◦	Added _resume_reload for reload on unhide in hide profiles.
	•	Tests (Various test files under tests/):
	◦	Adjustments for API testing across content types (string, multiprocess, table, function).
	◦	Fixes for test failures, e.g., in win_spec.lua (concatenation errors with function values).
	•	Other Minor Changes:
	◦	Replaced vim.v.argv with _G.arg for consistency.
	◦	Inline functions like setup_fzf_interactive_native.
	◦	Workarounds for title flags in non-multiprocess modes.
Overall, the refactoring reduces code duplication by treating all inputs as strings, improves modularity (e.g., easier picker combining), and enhances performance in async scenarios.
Potential Major Flaws, Bugs, or Issues
The PR seems stable based on the iterative fixes in commits and comments (e.g., bugs reported by @phanen were addressed promptly). However, here are some potential concerns, drawn from code diffs, TODOs, and discussions:
	1	Incomplete or Unfinished Work (TODOs):
	◦	The initial comment lists several TODOs: Merge core.mt_cmd_wrapper into shell.stringify (partially done via removals); auto-convert fn_transform to strings with string.dump (implemented, but warns on upvalues); add comprehensive API testing for all input types (adjustments made, but not exhaustive); deprecate fn_pre/post_fzf (done in commits); add opts._resume_reload for hide/unhide reloads (added).
	◦	While many are addressed in later commits, the PR was described as “still a lot more work” early on. If not fully completed, this could leave edge cases unhandled, especially for dynamic content like buffer changes.
	2	Race Conditions and Async Issues:
	◦	Fixed a race where pipes closed before all reads/writes completed in non-multiprocess modes (added read_cb_count and write_cb_count). Also used vim.schedule to avoid “attempt to yield across C-call boundary” errors.
	◦	Potential flaw: These fixes rely on callback counting, which could be error-prone in high-load scenarios (e.g., large folders or rapid typing in live_grep). If not thoroughly stress-tested, partial/empty results might reoccur, as noted in comments.
	3	Performance Regressions:
	◦	Discussions highlight inconsistencies: multiprocess=false was reported as faster than true in some cases (e.g., live_grep with globs), but shouldn’t be since multiprocess offloads to external commands. This was attributed to temp-throttle services, but could indicate unoptimized string conversion or icon handling overhead.
	◦	New string dumping for functions might add latency if functions have upvalues (code handles this by falling back, but could degrade in complex setups).
	4	Breaking Changes and Deprecations:
	◦	Bumping fzf to >0.36 deprecates older versions, potentially breaking setups for users on outdated fzf (though it’s 2+ years old).
	◦	Removal of fn_pre/post_fzf, raw_action, etc., could break custom configurations or plugins relying on them. No migration guide in diffs, though deprecations are noted.
	◦	Picker combining is new but warns about mismatching previewers (e.g., combining incompatible types might crash or error).
	5	Bug Fixes Indicate Prior Issues:
	◦	Fixed glob not working in live_grep after specific commits; live workspace symbols broken and repaired; partial results in large file listings.
	◦	A test fix in win_spec.lua addressed concatenation of function values, suggesting type-handling bugs were introduced mid-PR.
	◦	Multiprocess=false (debug mode) had icon-related issues, fixed but highlights that debug paths might still be fragile.
	6	Testing and Coverage:
	◦	API test adjustments were made, but the TODO emphasizes testing all use cases (string, multiprocess, table, function). If incomplete, regressions in less-common scenarios (e.g., function-based inputs with upvalues) could slip through.
	◦	No new integration tests for picker combining or resume reloads visible in diffs.
	7	General Code Quality:
	◦	Error-prone areas: String conversion assumes dumpable functions; if upvalues reference non-dumpable state, it falls back but might lead to silent failures.
	◦	No major security issues (e.g., shell command injection), as inputs are plugin-controlled, but user-provided cmds in custom pickers could be risky if not escaped properly (existing utils handle this).
	◦	The force-pushes (e.g., 8+ times) suggest rapid iteration, which is fine but increases risk of overlooked changes.
In summary, this PR introduces valuable simplifications and features without obvious critical bugs in the final state, but its WIP nature early on and remaining TODOs suggest it might benefit from more testing before merging. If you’re using fzf-lua, test in a staging environment, especially for multiprocess=false or custom functions. For the latest status, check the PR directly, as it’s open as of July 16, 2025. 0 1

ibhagwan:
With https://github.com/ibhagwan/fzf-lua/commit/34f7ef5a675fc88f3958012a16ff888f28b405da and some improvements to `combine` and `global` this PR should close #1879, #2058 and #2063.

phanen:
> how do you feel about merging this?

I think it's fine. The only break change for public api is `fzf_live`: https://github.com/search?q=fzf_live%20lang%3Alua%20NOT%20is%3Afork&type=code

ibhagwan:
> > how do you feel about merging this?
> 
> I think it's fine. The only break change for public api is `fzf_live`: https://github.com/search?q=fzf_live%20lang%3Alua%20NOT%20is%3Afork&type=code

Alright, I'll merge it tomorrow, see if you can think of any other issues.

Also, the new `global` picker has an issue with resume (that isn't an "unhide"), probably has more issues that I can't think about ATM but it's a new picker, I can fix post merge.

ibhagwan:
@phanen, updated to 0.11.3, added retires on win/Mac due to their flakey nature with the wait callbacks, let’s hope this ends the manual retries I have to do :)

https://github.com/ibhagwan/fzf-lua/pull/2152/commits/d1bfe75ff9c3c9e9c11c7d28295e7ae7aa08c368

ibhagwan:
I’m not super happy with the hacky opts propagation in the new global picker, I’ll merge this anyway and work on this later, need to find a better solution that propagates consistent opts to both actions and convert actions (for reloading and hide profile).

PR #2153, opened 2025-07-04
fix move into stringify

phanen:
- **refactor: move mt_cmd_wrapper into stringify**

### DIFF
--- a/lua/fzf-lua/complete.lua
+++ b/lua/fzf-lua/complete.lua
@@ -93,8 +93,7 @@ M.path = function(opts)
     end
   end)()
   opts = set_cmp_opts_path(opts)
-  local contents = core.mt_cmd_wrapper(opts)
-  return core.fzf_exec(contents, opts)
+  return core.fzf_exec(opts.cmd, opts)
 end
 
 M.file = function(opts)
@@ -115,8 +114,7 @@ M.file = function(opts)
     end
   end)()
   opts = set_cmp_opts_path(opts)
-  local contents = core.mt_cmd_wrapper(opts)
-  return core.fzf_exec(contents, opts)
+  return core.fzf_exec(opts.cmd, opts)
 end
 
 M.line = function(opts)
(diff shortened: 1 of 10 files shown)


ibhagwan:
Ty @phanen, I didn’t know you wanted to join in on the fun, let  me know when you want me to merge this into the PR :)

phanen:
Actually these codes from my git stash long time ago... I was trying to refactor this part but half way done.

ibhagwan:
> Actually these codes from my git stash long time ago... I was trying to refactor this part but half way done.

Nice, this will make things much simpler as everything lua  <-> shell will have just one entry point with `stringify`.

phanen:
Hmm.. so `config.bytecode` was one more round trip to fetch the config from editor instance? (To make base64 payload shorter?)

ibhagwan:
> Hmm.. so `config.bytecode` was one more round trip to fetch the config from editor instance? (To make base64 payload shorter?)

Correct, I figured if I transfer full functions (via string.dump) in the `spawn_stdio` string it will be too long so for some  parameters which are a  function (formatters, rg_glob_fn, etc) it’s easier to grab the string.dump output over the pipe, only did this for parameters which won’t break the multiprocess if I run the string as standalone (i.e. no formatter will produce different output but still work).

phanen:
I stop the PR here. Hope I didn't mess more thing up.

> grab the string.dump output over the pipe

The reason why we have to play with string -> the only way to interact with host is `nvim_exec_lua`... hope thing will be better when upstream support registering custom rpc handler:
* https://github.com/neovim/neovim/issues/31868
* https://github.com/neovim/neovim/issues/34592

ibhagwan:
> I stop the PR here. Hope I didn't mess more thing up.

Ty! I’ll merge it when I continue work on this, I anticipate there will be a lot of things to fix anyways but the end result would be great.

> The reason why we have to play with string -> the only way to interact with host is nvim_exec_lua... hope thing will be better when upstream support registering custom rpc handler:

Looks interesting.

PR #2154, opened 2025-07-05
ci: format with stylua

PeterCardenas:
Formats all files with the current stylua config and adds a ci check for whether running stylua formatting is needed.


PeterCardenas:
would appreciate feedback on whether the current stylua config is actually what we want. Feels like too many changes for me to think that the stylua config is what we want.

ibhagwan:
Ty @PeterCardenas  but we use `lua_ls` to format the code, see `.editorconfig`.

PR #2156, opened 2025-07-05
fix(btags): correctly parse and cache entry

phanen:


ibhagwan:
I just committed a fix in my local branch which included only this :)
``` lua
  local bufnr = not opts._ctag and idx > 1 and entry:sub(1, idx):match("%[(%d+)%]") or nil
```

Will take a look at the rest of the changes.

ibhagwan:
Looks good, is this ready to be merged or you need more changes?

phanen:
Maybe we can make `cache` usuable for all pickers (if possible).

Rather than add `:entry_to_file` I think we should add a `:parse_entry_cachable` then revert `:parse_entry` to the old behavior (don't parse entry.tick).

ibhagwan:
> Maybe we can make `cache` usuable for all pickers (if possible).
> 
> Rather than add `:entry_to_file` I think we should add a `:parse_entry_cachable` then revert `:parse_entry` to the old behavior (don't parse entry.tick).

Why not, sounds great but maybe better to have a separate PR for this as it's a bigger change?

phanen:
Yeah. Maybe I can do it in another PR after more consideration.

ibhagwan:
Ok merging this one then.

PR #2161, opened 2025-07-06
refacator contents

phanen:
- **chore: expand_query**
- **chore: vim.v.argv -> _G.arg**
- **chore: types**
- **chore: set default fn_selected in fzf_wrap**

### DIFF
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -299,7 +299,8 @@ M.live_grep = function(opts)
 
   -- this will be replaced by the appropriate fzf
   -- FIELD INDEX EXPRESSION by 'fzf_exec'
-  opts.cmd = opts.multiprocess and get_grep_cmd(opts, core.fzf_query_placeholder, 2)
+  local cmd = get_grep_cmd(opts, core.fzf_query_placeholder, 2)
+  opts.cmd = opts.multiprocess and cmd
 
   -- search query in header line
   opts = core.set_title_flags(opts, { "cmd", "live" })
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -334,10 +334,13 @@ M.live_grep_native = function(opts)
   opts = opts or {}
   opts.git_icons = false
   opts.file_icons = false
+  opts.file_ignore_patterns = false
+  opts.strip_cwd_prefix = false
   opts.path_shorten = false
   opts.formatter = false
+  opts.multiline = false
   opts.rg_glob = false
-  opts.multiprocess = true
+  opts.multiprocess = 1
   return M.live_grep(opts)
 end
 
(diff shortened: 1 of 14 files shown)


ibhagwan:
Ty @phanen! let me know when this contains all changes you want and I’ll merge it

phanen:
Just found (also in `main` branch) `alt-i`/`alt-h` don't set_title_flags  when `FzfLua live_grep multiprocess=false` https://github.com/ibhagwan/fzf-lua/blob/8d4d92073454bcd8090bee1e100dcbcb6eb7b189/lua/fzf-lua/core.lua#L772-L774

ibhagwan:
> Just found (also in `main` branch) `alt-i`/`alt-h` don't set_title_flags when `FzfLua live_grep multiprocess=false`
> 
> https://github.com/ibhagwan/fzf-lua/blob/8d4d92073454bcd8090bee1e100dcbcb6eb7b189/lua/fzf-lua/core.lua#L772-L774

Nice catch, maybe it’s time to move the action definitions into defaults as `= { fn = … header = <definition> }`?
https://github.com/ibhagwan/fzf-lua/blob/9be1552e7de8d72fa280b87ede3642e8bfb4c998/lua/fzf-lua/core.lua#L16-L28

phanen:
Ok. That's all my changes.

> Nice catch, maybe it’s time to move the action definitions into defaults as = { fn = … header = <definition> }?

Not sure about how it work. But I've push a quick hack... maybe you know more about how to better handle this case.

lua/fzf-lua/providers/grep.lua
```diff
@@ -299,7 +299,8 @@ M.live_grep = function(opts)
 
   -- this will be replaced by the appropriate fzf
   -- FIELD INDEX EXPRESSION by 'fzf_exec'
-  opts.cmd = opts.multiprocess and get_grep_cmd(opts, core.fzf_query_placeholder, 2)
+  local cmd = get_grep_cmd(opts, core.fzf_query_placeholder, 2)
```
ibhagwan, line 302: Not sure I understand this change, isn’t the end result of these 2 lines exactly the same?

ibhagwan:
> Not sure about how it work. But I've push a quick hack... maybe you know more about how to better handle this case.

Seems about right, so the issue with title flags was only with live_grep/grep?

lua/fzf-lua/providers/grep.lua
phanen: No `get_grep_cmd` has a side effect to set `opts._cmd`. Previously it's not executed since short circuit. SO this is a hack..

phanen:
Yeah, mt version pre-built a `opts.cmd` while st version didn't.

lua/fzf-lua/providers/grep.lua
ibhagwan: Oh you’re right, yet one more of my “undocumented” hacks lol

Maybe it’s better to add a comment there so we don’t get over eager “optimizing” it one day, or make the hack less hacky.

ibhagwan:
Squash merged manually due to conflicts as I modified https://github.com/ibhagwan/fzf-lua/pull/2152/commits/5a170899ec92cd54422a5c8d22eb6f5bd3c2c00c.

PR #2165, opened 2025-07-06
fix(call_hierarchy): Don't open fzf window with 1 result and `opts.jump1`

andrewbraxton:
With 1 result, the cursor will jump to the location but the fzf window also opens

Avoid opening the window by relying on the jump1 logic inside gen_lsp_contents()


ibhagwan:
Ty @andrewbraxton! I missed that on the latest refactor :-)

PR #2168, opened 2025-07-07
feat: allow keymap to focus preview window

PeterCardenas:
Allow focusing the preview window, since there's already a keymap to focus the fzf input from the preview window.

PR #2169, opened 2025-07-07
fix: formatting + remove logging statment

PeterCardenas:
cleanups from this pr: https://github.com/ibhagwan/fzf-lua/pull/2168


ibhagwan:
My bad, missed this on the review from the phone.

PR #2172, opened 2025-07-09
chore: remove stylua config

PeterCardenas:
It's pretty confusing that we have a stylua config file when we don't use stylua for formatting at all. So removing it here.

### DIFF
--- a/.stylua.toml
+++ b/.stylua.toml
@@ -1,7 +1,0 @@
-column_width = 100
-line_endings = "Unix"
-indent_type = "Spaces"
-indent_width = 2
-quote_style = "AutoPreferDouble"
-call_parentheses = "Always"
-collapse_simple_statement = "Never"


ibhagwan:
Ty @PeterCardenas but I don’t think this needs to be removed, it’s not that big of a deal one day we might change our mind and there isn’t too many contributors anyway.

FYI tests are failing because you didn’t update the screenshots.

PeterCardenas:
@ibhagwan it's a bit annoying for me because if i want auto formatting for my config i have to make an explicit exception for this repo to not use stylua even though the config exists

ibhagwan:
> @ibhagwan it's a bit annoying for me because if i want auto formatting for my config i have to make an explicit exception for this repo to not use stylua even though the config exists

Gotcha, I have a second keybind to test with stylua to see if it’s less opinionated, haven’t used it in a while, either way not a big deal I’m sure this can be excluded somehow.

PR #2177, opened 2025-07-16
fix(picker): file picker not respecting .gitignore

amlanjlahkar:
By default `fd` only respects global or local gitignore rules if searching inside a git repository. However in certain scenarios, e.g. when using [jujutsu](https://github.com/jj-vcs/jj) instead of git, this approach never takes effect since it can't find a .git directory in cwd.

The proposed `--no-require-git` flag allows `fd` to respect all git related ignore rules, regardless of whether it's searching inside a git repo or not.

I think it's safe to include this flag in default cmd.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -349,7 +349,7 @@ M.defaults.files                = {
     "git", "-c", "color.status=false", "--no-optional-locks", "status", "--porcelain=v1" },
   find_opts              = [[-type f \! -path '*/.git/*']],
   rg_opts                = [[--color=never --files -g "!.git"]],
-  fd_opts                = [[--color=never --type f --type l --exclude .git]],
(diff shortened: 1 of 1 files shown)


ibhagwan:
Tyvm @amlanjlahkar for this thoughtful PR, seems this flag existed since fd v8.7 which is old enough, so this goes in :)

https://github.com/sharkdp/fd/releases/tag/v8.7.0

amlanjlahkar:
From the release note you mentioned, this behaviour is confirmed for ripgrep as well and I guess it also should be the default for grepping in general

ibhagwan:
> From the release note you mentioned, this behaviour is confirmed for ripgrep as well and I guess it also should be the default for grepping in general

We can add it to rg as well it it makes sense and is old enough to not cause issues with older rg versions.

ibhagwan:
Already have two issues with this, may have to revert it and you can just add this to your fd_opts instead, #2180, #2179.

amlanjlahkar:
okay, no problem

Issue #2188, opened 2025-07-18
Bug: Error opening Fzf in make_entry


Grueslayer:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Windows 11

### Shell

cmd

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-782+gfb0dc825e9

### Fzf version (`fzf --version`)

0.53.0 (c4a9ccd)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require("fzf-lua").setup({
        fzf_opts = {
          ['--layout'] = 'default'
        },
        winopts = {
          height = 0.95,
          width = 0.95
        },
        actions = {
          files = {
            ["default"] = file_drop_or_qf,
            ["ctrl-s"]  = actions.file_split,
            ["ctrl-v"]  = actions.file_vsplit,
            ["ctrl-t"]  = actions.file_tabedit,
            ["alt-q"]   = actions.file_sel_to_qf,
            ["alt-l"]   = actions.file_sel_to_ll,
            ["ctrl-d"]  = file_vdiffsplit
          },
          buffers = {
            ["default"] = actions.buf_edit,
            ["ctrl-s"]  = actions.buf_split,
            ["ctrl-v"]  = actions.buf_vsplit,
            ["ctrl-t"]  = actions.buf_tabedit,
          },
        },
        files = {
          formatter      = { "path.filename_first", 2 },
        },
        buffers = {
          formatter      = { "path.filename_first", 2 },
        },
        oldfiles = {
          formatter      = { "path.filename_first", 2 },
        },
        git = {
          branches = {
            actions = {
              ["default"] = git_switchlocal,
            }
          },
        },
      })
```

### Describe the bug / steps to reproduce

This came once and is not reproducable anymore:

```
[...]

PR #2189, opened 2025-07-18
fix: nil-check opts.cmd #2188

phanen:


phanen:
I believe those `loadstring` don't make sense cause if you going to avoid circular ref you should just avoid require the module. This don't avoid load the module. 

Since `require` has already been wrapped in a function, I think they are not necessary, this is the same:
```
local func = loadstring(...)
local func = function() ... end
```

Maybe I can use `_G.FzfLua` for all, but I'm not sure if some module function can be called before `require('fzf-lua')`

ibhagwan:
Ty @phanen !

PR #2193, opened 2025-07-19
refactors

phanen:
- **refactor: patch_shell_action -> shell.stringify_data2**
- **refactor: move ctx into a new module**


ibhagwan:
This is great @phanen! Let me know when you’re fully done with this and I’ll test/merge.

phanen:
ready for review. Just some renaming and moving. 
Having both `resume_get` and `get_info` feels not necessary. Maybe merge them in future.

phanen:
Two new bug fix in the last two commit:
`require('fzf-lua').fzf_live('echo ', { query = '%%', debug = true })`

* eat one `%`
* also command has repeated `... {q} {q}`

Not sure if there's better way to fix them.

phanen:
I'm considering move `stringify_mt` out from `stringify`...

ibhagwan:
> I'm considering move `stringify_mt` out from `stringify`...

Guess it could make sense since it’s only good for shell commands.

No rush, whenever  you feel you’re done with this let me know and I’ll do some testing too.

I want to add a new live_grep test with all special chars:

https://github.com/ibhagwan/fzf-lua/blob/07326bb95827f1f4b7d7aead5e8094d844588d08/lua/fzf-lua/utils.lua#L168-L170

phanen:
https://github.com/ibhagwan/fzf-lua/pull/2193/commits/effc34a00898d65958a710a124c4ab7ae1550732 impl live_grep_glob_native.

Even no rg_glob it's the same fast:
```
:FzfLua live_grep debug=true multiprocess=false rg_glob=false fn_transform=false fn_preprocess=false fn_postprocess=false
:FzfLua live_grep debug=true multiprocess=true  rg_glob=false fn_transform=false fn_preprocess=false fn_postprocess=false
```

ibhagwan:
> [effc34a](https://github.com/ibhagwan/fzf-lua/commit/effc34a00898d65958a710a124c4ab7ae1550732) impl live_grep_glob_native.
> 
> Even no rg_glob it's the same fast:
> 
> ```
> :FzfLua live_grep debug=true multiprocess=false rg_glob=false fn_transform=false fn_preprocess=false fn_postprocess=false
> :FzfLua live_grep debug=true multiprocess=true  rg_glob=false fn_transform=false fn_preprocess=false fn_postprocess=false
> ```

Amazing! I’ll have to look what you did here when I get home.

ibhagwan:
`global` picker was much slower to open and also failed trying to switch to the other pickers (buffers, symbols), 
https://github.com/ibhagwan/fzf-lua/pull/2193/commits/4a69f9bf457a0840b8593dafe3f5b9c32874e794 - this fixes the global picker as it wasn't inheriting the opts from the lsp/buffer defaults which was causing the LSP calls to be made not asynchronously.


https://github.com/phanen/fzf-lua/blob/4a69f9bf457a0840b8593dafe3f5b9c32874e794/lua/fzf-lua/config.lua#L157-L159
```lua
  if opts._normalized then
    return opts
  end
```
This is obviously the right way to go just needs to be tested, could affect pickers that call  other pickers, in this case global but also tags/btags and maybe others.

Also if a user had an action/keybind that switches say from files to git_status / buffers /etc and reuses the options from the callback `actions = { ["ctrl-g"] = function(sel, o) FzfLua.buffers(o) end }`, I'm not even sure if we should support this but just something to thing about.

ibhagwan:
> https://github.com/ibhagwan/fzf-lua/commit/effc34a00898d65958a710a124c4ab7ae1550732 impl live_grep_glob_native
> Even no rg_glob it's the same fast:

This fails for me, seems it doesn't replace the last arg `{argvz}`:
```
:FzfLua live_grep debug=1 multiprocess=true  rg_glob=false fn_transform=false fn_preprocess=false fn_postprocess=false
```
image

phanen:
iirc argvz replaced in preprocess
i'm not sure how i test...maybe preprocess=true

ibhagwan:
> iirc argvz replaced in preprocess i'm not sure how i test...maybe preprocess=true

Yap it is in preprocess, maybe in the new +transform you added (I didn’t get to it yet)

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2193/commits/b794e049c72cc889792010116ff78f36682f7607

https://github.com/junegunn/fzf/blob/master/CHANGELOG.md#0450

phanen:
no i seems use fn-tramsform=false only test

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2193/commits/9492fee05b568c8ca55d1a12836ab3e47b73bddd added CI for grep'ing regex with all special chars, semi-useful on Windows where a lot of the lines are ignored due to backslash <-> forward slash replacement but at least we know it works and matches the correct line (with  `multiprocess = {false|true}`).

nenahp:
Force push because my skim fix  break my fix on duplicated `{q} {q}`.

ibhagwan:
Good call, I saw the dead code and forgot to remove, wrap spawn makes a lot more sense in shell.

ibhagwan:
I’m going to merge this unless you have further objections

phanen:
ok

PR #2198, opened 2025-07-21
refactor: simplify serialization

phanen:


scripts/headless_fd.sh
```diff
@@ -132,28 +132,25 @@ if [ $file_icons != "nil" ]; then
 fi
 
 # VIMRUNTIME=/usr/share/nvim/runtime \
-nvim -u NONE -l ${BASEDIR}/../lua/fzf-lua/spawn.lua "return
+nvim -u NONE -l "${BASEDIR}"/../lua/fzf-lua/spawn.lua "return
```
ibhagwan: Probably doesn’t make a difference but maybe better to wrap the entire path as string?
```sh
nvim -u NONE -l "${BASEDIR}/../lua/fzf-lua/spawn.lua"
```

phanen: That's automatically format by my `shfmt` (Actually it change all the file and I forget revert this part

ibhagwan: I thought you had an issue running this due to a username with spaces or special chars.

ibhagwan:
Great changes, let me know when you want this merged.

phanen:
Amened to the first comment: move `loadstring` into `deserialze`

nenahp:
Oh I should remove renamed screenshots.

tests/actions_spec.lua
```diff
@@ -0,0 +1,37 @@
...
+local MiniTest = require("mini.test")
+local helpers = require("fzf-lua.test.helpers")
+local child = helpers.new_child_neovim()
+local expect, eq = helpers.expect, helpers.expect.equality
+local new_set = MiniTest.new_set
+
+---@format disable-next
+local reload = function(config) child.unload(); child.setup(config) end
+local sleep = function(ms) helpers.sleep(ms, child) end
+local exec_lua = child.lua
```
ibhagwan, line 11: I noticed this makes the lua code inside the exec_lua call have proper lua syntax highlighting, haven’t figure out why yet but it’s nice :)

nenahp: https://github.com/nvim-treesitter/nvim-treesitter/blob/692b051b09935653befdb8f7ba8afdb640adf17b/runtime/queries/lua/injections.scm#L45-L52

I can use ts hl and ts related plugins in codeblock

nenahp: This is wildly used in neovim's testsuite

ibhagwan: > This is wildly used in neovim's testsuite

I see, I was wondering why renaming the function makes it behave this way, some of the reason I created `helpers.FzfLua` was to have syntax highlighting too but it does a bit more, also serializes functions as bytecode which is nice.

ibhagwan:
> Oh I should remove renamed screenshots.

Yes it doesn’t break the tests when new files are added anymore because I excluded the tests folder, nice side effect it also tests `file_ignore_patterns`.

ibhagwan:
Looks great @phanen, shall I merge this or you need more time?

nenahp:
ok

PR #2200, opened 2025-07-22
chore(luals): more config typings

phanen:
Mostly an ai generated template... we can improve them gradually.


lua/fzf-lua/path.lua
```diff
@@ -412,12 +412,12 @@ function M.entry_to_location(entry, opts)
 end
 
 ---@param entry string
----@param opts fzf-lua.Config?
+---@param opts fzf-lua.Config
 ---@param force_uri boolean?
 ---@return fzf-lua.path.Entry
 function M.entry_to_file(entry, opts, force_uri)
   -- NOTE: see note in meta.lua:global regarding alt options
```
ibhagwan: I think we should leave this line as is, many users use `entry_ to_file` and do not necessarily provide opts.

scripts/mini.sh
```diff
@@ -46,6 +46,6 @@ else
 fi
 
 # Run neovim
-HOME=${TEMPDIR} PACKPATH=${packpath} ${nvim_bin} -u ${tmp_rtp}/${plug_name}/scripts/init.lua
+NVIM_APPNAME=fzf-lua HOME=${TEMPDIR} PACKPATH=${packpath} ${nvim_bin} -u ${tmp_rtp}/${plug_name}/scripts/init.lua
```
ibhagwan: If we start using NVIM_APPNAME we can probably remove other stuff too like $HOME, when I wrote this originally NVIM_APPNAME didn’t exist.

ibhagwan:
> Mostly an ai generated template... we can improve them gradually.

Sure we can give this PR some time until we’re happy with it.

scripts/mini.sh
phanen: that would need 0.10.0+

i add it since it mistakenly load my stdpath(config) maybe we also need to assign xdg_xx env var

lua/fzf-lua/path.lua
phanen: right i ll revert this

scripts/mini.sh
ibhagwan: > that would need 0.10.0+

Shouldn’t be a problem, it up’s the min requirement just for mini.sh, only used for troubleshooting anyways.

ibhagwan:
This is looking really good, maybe we can look into auto generating docs from the annotations and ditch the forever WIP OPTIONS.md :)

phanen:
Mostly cleanup or diagnostics fix.

[`e8f4308`](https://github.com/ibhagwan/fzf-lua/pull/2200/commits/e8f43086341bf273ed4492276ba0fe553a73f56f) fix a regression from `opts._normalized` checking.

phanen:
> the forever WIP OPTIONS.md

Yeah, essentially this just convert the task from writing docs to solving typings puzzles...

ibhagwan:
> > the forever WIP OPTIONS.md
> 
> Yeah, essentially this just convert the task from writing docs to solving typings puzzles...

lol

ibhagwan:
I’m assuming this needs a squash/rebase before merge, let me know when you feel it’s ready

phanen:
Idk why I cannot rename typings use luals, but work with https://github.com/CppCXY/emmylua-analyzer-rust, also still beta but this is so much faster.

phanen:
I'm try to hide some field from cmp via `@field private xx` but it don't seems work. Anyway this don't hurt since before we get no cmp from `FzfLua.files{ xx = }`

ibhagwan:
> I'm try to hide some field from cmp via `@field private xx` but it don't seems work. Anyway this don't hurt since before we get no cmp from `FzfLua.files{ xx = }`

If we can auto-generate options.md from annotations we will get perfect completion for `:FzfLua files …` as the command completion is auto generated from it.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/cdfac04b264cd924173e7f63a00b4883e2b98564 FYI, it was lazydev messing up the local library loading and vim runtime.

~~Unsure why, getting similar failures you mentioned on manual `make lint` (CI is ok):~~ I’m a dodo, forgot to `make deps` lol

phanen:
Have to afk for few months, maybe merge this or left it here and I'll rebase when I got back.

The next step may require refactor on `normalize_opts` (for each picker), to convert partial types to exact types.

ibhagwan:
> Have to afk for few months, maybe merge this or left it here and I'll rebase when I got back.
> 
> The next step may require refactor on `normalize_opts` (for each picker), to convert partial types to exact types.

If you feel this is ready, I’ll work on the rebase so it’s less of a mess with the renames.

Have fun on your journey @phanen and ty for your help!

PR #2204, opened 2025-07-24
fix(fzf_live): shell nop since we use `--start` event

phanen:


lua/fzf-lua/libuv.lua
```diff
@@ -422,6 +422,11 @@ M.spawn_stdio = function(opts)
     io.stdout:write("[DEBUG] [mt] " .. opts.cmd .. EOL)
   end
 
+  if not fn_transform and not fn_postprocess and not _is_win and opts.cmd and pcall(require, "ffi") then
```
ibhagwan, line 425: Interesting, does this make a big difference on performance for multiprocess with glob processing only?

The conditions for this fork to be met will be quite rare, no icons, no formatting, no file_ignore_patterns… and if there’s no fn_transform uv.spawn callback sends the data to the pipe as is (without line split which is pretty fast), so I’m wondering maybe we’re even losing performance from spawning a new process and having to re-execute `sh -c` vs just uv.spawn?

phanen: Yeah this is just an idea to show it's possible.

We can avoid shell, or avoid fork by shell `exec` or bash 5.5's `${ cmd; }`

But this don't seem faster than `io.stdout:write`, maybe `rg` is the bottleneck rather than lua in this case.

ibhagwan: Got it.

We can merge it just as a reminder of what’s possible, 99% of the users won’t hit this code fork anyway :)

PR #2236, opened 2025-08-09
chore(lua_ls): make some fields optional

MariaSolOs:
I just noticed that I had some LSP warnings about required missing fields in my configuration.
image

There might be other types that have optional fields, but I'm just modifying the ones I'm confident about for now.


ibhagwan:
Ty @MariaSolOs!

PR #2244, opened 2025-08-12
fix: FzfLua blines profile=ivy winopts.treesitter.enabled=false

phanen:
- **refactor(config): extract normalize_tbl to expand dot keys**
- **fix(config): `FzfLua blines profile=ivy winopts.treesitter.enabled=false`**

Not sure if this a correct fix but it work for me.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -144,21 +144,39 @@ local eval = function(v, ...)
   return v
 end
 
+
+---expand opts that were specified with a dot
+---@param opts table
+local normalize_tbl = function(opts)
+  -- convert keys only after full iteration or we will
+  -- miss keys due to messing with map ordering
+  local to_convert = {}
+  for k, _ in pairs(opts) do
+    if k:match("%.") then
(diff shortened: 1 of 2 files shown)


ibhagwan:
Thought you’re on vacation for 2 months lol

Take a look at the dev branch, I did some changes to address something similar where profile= wasn’t inheriting opts from setup, so if you had profile= and other opts (e.g fzf_opts history) - the others would be ignored.

I also split symbols opts to document/workspace so I could enable Ivy for doc symbols, also enabled for treesitter but the hls aren’t working yet as there’s no reverse hl - will try to complete foreground later today to get the in_matched - maybe you have another idea.

I modified the logic for line parsing too.

phanen:
actually not awk but just have few time on keyborad in recent months

i ll give it a test next time
this pr mainly for report the issue: swiper like preivewer dot't work ootb with my custom extended opts

ibhagwan:
> this pr mainly for report the issue: swiper like preivewer dot't work ootb with my custom extended opts

Yes I figured from the code, I’ll test this later and merge when not AFK.

PR #2258, opened 2025-08-16
fix(hide): actions should be nop when nothing match

phanen:
Regression from f0e830e278ad954a4e545928f8bfe08725a71173

fix: opts.__pid -> opts.PidObject


phanen:
e.g. `FzfLua files query=nothing` then `<c-t>` should not open new tab

ibhagwan:
Ty @phanen!

PR #2265, opened 2025-08-20
feat: support toggle_behavior="extend" for all pickers

phanen:
- **chore: never SIGWINCH when fzf < 0.46**
- **fix: SIGWINCH on toggle_preview for any native previewer**
- **feat: fallback to winopts.toggle_behavior**
- **feat: support toggle_behavior="extend" for all pickers**
- **refactor: normalize picker specs**


lua/fzf-lua/win.lua
```diff
@@ -152,6 +153,9 @@ function FzfWin:setup_keybinds()
         })
       end
     end
+    if not self._fzf_toggle_prev_bind then
```
ibhagwan: Maybe we don’t even need the bind anymore, if we trigger SIGWINCH we can hide close the preview in the transform no?

phanen:
Fix a typo in condition, now:
```lua
  if not self:has_previewer()
      or (self.previewer_is_builtin and (self.preview_hidden and self.toggle_behavior ~= "extend"))
      or (not self.previewer_is_builtin and (self.toggle_behavior ~= "extend" or not self.preview_hidden))
```

lua/fzf-lua/win.lua
phanen: No, now we don't use fzf's `toggle-preview`, so we must feed the key now.

that also means `require("fzf-lua.win").toggle_preview()` have to run in terminal mode, I try replace feedkey with nvim_chan_send but don't seem fix.

phanen: it work with `ctrl-xx`. not with `alt-xx`, no idea why atm

phanen:
```sh
fzf --bind 'alt-a:toggle-preview' --preview 'echo 1'
```
```vim
:=vim.api.nvim_chan_send(vim.bo.channel, vim.keycode('<a-a>'))
```

ibhagwan: > ```shell
> fzf --bind 'alt-a:toggle-preview' --preview 'echo 1'
> ```
> 
> ```viml
> :=vim.api.nvim_chan_send(vim.bo.channel, vim.keycode('<a-a>'))
> ```

Maybe `<M-a>` will work?

ibhagwan:
@phanen, this looks good, I made tiny changes to the last commit:
1. Restored the builtin picker dimensions and made previewer hidden by default so it's not a breaking change.
2. Restored ivy `toggle_behavior = "extend"` to only the grep/lines pickers, it looks horrible with the default `files` (and every other picker that doesn't have the "fullscreen-up" layout.
3. Ability to hide the"no toggle-preview keymap" with `silent=true` (some users copy keymaps table without `[1] = true` so they don't inherit the defaults so they'll be annoyed by the message).


This isn't perfect yet, when using `bat` as previewer (or any native) the layout isn't a perfect match, I'm missing lines when toggling.

Default files with previewer:
image

After toggle, one line less:
image

Ivy style:
image


After toggle, again missing a line:
image

phanen:
seems related to height calculation...floor/ceil

phanen:
i see the code again, the actual problem is border width/height is handled in last step

if border is none or fused or.. then its width/height is not 2 which defeat the assumption of previous calculation from percentage

so to fix we should get layout metadata and border first then refactor all subsequnces height/width calculation

lua/fzf-lua/win.lua
```diff
@@ -397,7 +402,7 @@ function FzfWin:normalize_border(winopts, metadata)
     rounded               = "rounded",
     solid                 = "solid",
     empty                 = "solid",
-    shadow                = "shadow",
+    shadow                = "shadow", -- TODO:... fuck
```
ibhagwan: lol, what happened

phanen: haha forgot it, i guess shadow is partial border

phanen: i'm wrong, actually it's not

phanen:
revert last fix
position in the center of screen rather than the center of window seems better, also avoid introduce large diff before another possible refactor

ibhagwan:
I still see you’re making changes to this, lmk when you feel it’s ready and I’ll test/merge this

phanen:
Nothing more now.

The last commit use sigwinch to toggle-preview if possible.
Not sure why test fail when winopts.split on windows, have to fallback to feedkey in this case.

Also in any case emit `SIGWINCH` if possible to trigger `change-window-preview` event. Because `change-window-preview` won't happen if preview is hidden ($FZF_PREVIEW_LINES expand to empty): https://github.com/ibhagwan/fzf-lua/blob/1e00da7dcc2ddddd2f9d27756f29208dbd0e9534/lua/fzf-lua/core.lua#L381-L382)

ibhagwan:
> Nothing more now.

The latest changes broke Windows CI, I’ll wait until you’re ready.

phanen:
Not sure why, use SIGWINCH to toggle_preview fail windows test, fix by feedkey on windows.

phanen:
ready for review now.

ibhagwan:
> ready for review now.

Ty @phanen!

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2265/commits/9e2da1b15ff85cef7f082d4eaa6f92ccb3a22e85 due to change in `test/helpers.lua:487`

ibhagwan:
I noticed `FzfWin.on_SIGWINCH` generates a new resize event callback for each new scope which translates to another independant neovim headless callback, maybe Windows CI may have failed due to the command being too long?

Wouldn't it be better to have single entry point from fzf, call the headless wrapper once `on_SIGWICH` and run the callback for each scope from the `stringify_data` callback?

phanen:
Still fail, apparently it's not about cmdline len.

ibhagwan:
> Still fail, apparently it's not about cmdline len.

Windows be windows I guess :)
Was worth a shot

phanen:
Not sure why ci fail again... will check later.

phanen:
Windows never stop surprising me: https://github.com/ibhagwan/fzf-lua/actions/runs/17273388520/job/49023852462?pr=2265

ibhagwan:
> Windows never stop surprising me: https://github.com/ibhagwan/fzf-lua/actions/runs/17273388520/job/49023852462?pr=2265

I’m not a fan of Windows to say the least…

ibhagwan:
What’s printing all these argv in the tests?

ibhagwan:
I see it libuv_spec.lua:181

phanen:
Now it's segment fault. Amazing.

ibhagwan:
> Now it's segment fault. Amazing.

ha ha

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/2c4c75e71eedbedd55ea13ab3400f72a2336e672, hopefully this doesn't create a conflict, if you wish, once you're done I can rebase this on main and resolve the conflicts.

tests/libuv_spec.lua
```diff
@@ -143,4 +144,57 @@ describe("Testing libuv module", function()
     assert.are.same(libuv.unescape_fzf([[foo\]], 0.52, true), [[foo\]])
     assert.are.same(libuv.unescape_fzf([[foo\\]], 0.52, true), [[foo\\]])
   end)
+
+  it("field index", function()
+    local fzf = require("fzf-lua.fzf")
+    local selected, exit_code
+    coroutine.wrap(function()
+      selected, exit_code = fzf.raw_fzf("echo foo\nbar", {
```
ibhagwan: I don’t think `echo foo\nbar` works in windows, this is the reason the `{}` returns nothing, I think this can be written as `echo foo&& echo bar` on Windows.

ibhagwan:
Rebased to main, tested `echo foo&& echo bar` - still no luck on getting Windows to show `{}`, something is off.

ibhagwan:
Oh shoot, I’m sorry I force pushed I had another idea to try the `one` event but I didn’t see your also working on this, I won’t touch it till you’re done

ibhagwan:
My apologies, hope I didn’t mess up your commit

phanen:
No, I didn't make some meaningful change. The ci seems hang.

ibhagwan:
Something is totally off with this test on Windows maybe we should give up on it lol

phanen:
`Assertion failed: self, file D:\a\neovim\neovim\src\nvim\event\multiqueue.c, line 136`. I did nothing but replace choco to scoop.

phanen:
ok green now, I leave the bad result in test. no more windows. 
Maybe sometime it's red again and not a segment fault or assert failure then we know windows ci finally do thing right.

ibhagwan:
Weird, tried running the job manually again, still segfaults, maybe we should disable the field index test on Windows altogether, it’s not really testing anything anyways since the indexes don’t expand.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2265/commits/a9cb3962a54bd9891e425215f345cf9ba357878f - this should be the correct way to skip with mini.test

ibhagwan:
Do you want me to squash all the fixup commits or you prefer to do it?

You can squash my commits too into the previous ones

phanen:
y free feel to sqash, i am afk now

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2265/commits/6b8b6ddbe293317151a5d4037f89fbe63b368f6d 

What do you think @phanen? could probably use a tiny refactor for your preference, not too happy about this, any ideas where to move this are welcome:
https://github.com/phanen/fzf-lua/blob/6b8b6ddbe293317151a5d4037f89fbe63b368f6d/lua/fzf-lua/win.lua#L1613-L1616

Also not optimal that core:379 `win.on_SIGWINCH`  is always sent so technically we send the `change-preview-window` twice but I guess it's not that critical since fzf ignores the 2nd identical change transform?

ibhagwan:
Latest change also messed up the "should not freeze on content error" test only on Windows, no idea why...

ibhagwan:
Found the issue on Windows, you already had the lead in the code... why does this fail the content error CI I have no idea:
https://github.com/phanen/fzf-lua/blob/b45a376dfe44be55abad22277907659479860950/lua/fzf-lua/win.lua#L150
https://github.com/phanen/fzf-lua/blob/b45a376dfe44be55abad22277907659479860950/lua/fzf-lua/win.lua#L164

The question is, is this only failing the CI and the rotation/toggle works on Windows? I no longer have access to Windows so I cannot test :-)

ibhagwan:
Rebased to HEAD, I’ll do some more testing today and hopefully merge this.

PR #2279, opened 2025-08-27
fix: correctly handle URIs without `//` after scheme

### LINKED ISSUE #2278: Bug: when an LSP that send a URI starting with `file:` (instead of `file://`) is used, the URI gets incorrectly prepended with `file://`
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Arch linux

### Shell

zsh

NVIM v0.11.3

0.65.1 (e5cd7f0a)

nil

- [ ] My issue is reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

1. Use a language server that responds with URIs similar to `file:/path/to/file` (instead of `file:///path/to/file`). https://github.com/GroovyLanguageServer/groovy-language-server is one example.

It's worth noting that this is a valid URI per https://datatracker.ietf.org/doc/html/rfc3986#section-3 . The syntax of an URI is defined as

```
      URI         = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

      hier-part   = "//" authority path-abempty
                  / path-absolute
                  / path-rootless
                  / path-empty
```
the `//` after the `scheme` and `:` it's not mandatory. Neovim, internally, with its `vim.lsp.buf.references/definition/etc` handles this correctly.

2. `:FzfLua lsp_definitions` with the cursor above any symbol recognized by the language server
3. The error `Error executing Lua callback: /usr/share/nvim/runtime/lua/vim/lsp/util.lua:947: Cursor position outside buffer` is shown. This is because Neovim it's trying to open the file `/file:/path/to/file` (because `fzf-lua` prepended the correctly formed URI `file:/path/to/file` with `file://`), and, since it's an empty buffer (because of the incorrect name), the cursor can't be set to the location specified by the LSP.

This was introduced by https://github.com/ibhagwan/fzf-lua/issues/1317

TheLeoP:
Fixes #2278

### DIFF
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -93,7 +93,7 @@ local function location_handler(opts, cb, _, result, ctx, _)
   -- HACK: make sure target URI is valid for buggy LSPs (#1317)
   for i, x in ipairs(result) do
     for _, k in ipairs({ "uri", "targetUri" }) do
-      if type(x[k]) == "string" and not x[k]:match("://") then
+      if type(x[k]) == "string" and not x[k]:match("^([a-zA-Z]+[a-zA-Z0-9.+-]*):.*") then
         result[i][k] = "file://" .. result[i][k]
       end
     end


nenahp:
(Test failure is unrelated, fixed on another branch.)

ibhagwan:
Ty @nenahp, I’ll merge it then.

Ty @TheLeoP!

Issue #2184, opened 2025-07-17
Feature: global picker - could be useful in LSP-less buffers too


liskin:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

I just tried the new global picker, but if I start it in a new vim instance with no LSP attached, I get
```
[Fzf-lua] LSP: no client attached
[Fzf-lua] LSP: no client attached                                                                                                       
Press ENTER or type command to continue                                                                                                 
```
which is quite annoying.

Can it perhaps fallback to btags and tags instead of lsp_document_symbols and lsp_workspace_symbols if no LSP is available?

Although, in a completely new vim instance that would still presumably result in something like
```
[Fzf-lua] 'btags' is not available for unnamed buffers.
[Fzf-lua] Tags file ('…/tags') does not exist. Create one with ctags -R
Press ENTER or type command to continue                                                                                                 
```

Can we perhaps silence that somehow?

ibhagwan:
> Can we perhaps silence that somehow?

Most warnings can be disabled by adding `silent=true` but picker contents weren’t always included in the silent condition, I can add that easily.

> Can it perhaps fallback to btags and tags instead of lsp_document_symbols and lsp_workspace_symbols if no LSP is available?

This is a new picker, the current selection is customizable, at first step you can replace lsp with tags during setup, here’s how the defaults are configured, copy this table and place it under `global.pickers` in setup:

https://github.com/ibhagwan/fzf-lua/blob/491b0882e470662218de2aa8dbdac3a9f134fde1/lua/fzf-lua/defaults.lua#L363-L371

I’ll look into doing this automatically if no LSP is attached, shouldn’t be a problem.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/4c3e5cd2680a01940289b7b6ac66e5231613c1d6 should fallback silently to tags, ctags binary error message suppressed.

The only semi-issue left with is due to the different previewer classes, the tags aren't previewed and highlighted (as in `btags|tags`) the file is displayed, enter will take you to the right place but the previewer won't show the exact location.

I'll have to try and think of a non-hacky solution for previewer fallbacks.

ibhagwan:
@phanen, maybe you have a nice non-hacky idea how to switch previewers in the global picker?

I’m thinking I can call `detach_previewer|attach_previewer` when switching pickers, i.e. here:
[...]

PR #2280, opened 2025-08-28
fix(config): pass `silent` to lsp provider

Parsifa1:
for `Global` picker:

i want to write like:
```lua
pickers = {
    { "frecency", desc = "Files" },
    { "buffers", desc = "Bufs", prefix = "$" },
    { "lsp_document_symbols", desc = "Symbols (buf)", prefix = "@", { silent = true } },
    { "lsp_workspace_symbols", desc = "Symbols (project)", prefix = "#", { silent = true } },
},
```
but `silent` peremeter seems useless when i passing, so i passed here.

i don't know if this "not pass `opts.silent`" is designed.

### DIFF
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -895,7 +895,7 @@ local function wrap_fn(key, fn)
     opts.lsp_handler.capability = opts.lsp_handler.server_capability
 
     -- check_capabilities will print the appropriate warning
-    if not check_capabilities(opts.lsp_handler) then
+    if not check_capabilities(opts.lsp_handler, opts.silent) then
       return
     end
 


Parsifa1:
seems passed wrong opts 😢

Parsifa1:
oh, seems nothing wrong? I'm confused.
if you think this is the right way, just fix it.

with this change, the lsp-related pickers in global will auto-setting `silent = true`.

---

图片

for command like `FzfLua lsp_document_symbols`,
behavior won't change.

ibhagwan:
Not sure what you are trying to do, take a look at the defaults you can add silent with `opts = …`:
https://github.com/ibhagwan/fzf-lua/blob/2c4c75e71eedbedd55ea13ab3400f72a2336e672/lua/fzf-lua/defaults.lua#L400-L404

You can also call global with silent, `:FzfLua global silent=true`.

Parsifa1:
> You can also call global with silent, `:FzfLua global silent=true`.

```lua
global = {
	pickers = {
	    { "frecency", desc = "Files" },
	    { "buffers", desc = "Bufs", prefix = "$" },
	    { "lsp_document_symbols", desc = "Symbols (buf)", prefix = "@" },
	    { "lsp_workspace_symbols", desc = "Symbols (project)", prefix = "#" },
	},
}
```
i rewrited my `global` config, and i realized that  `:FzfLua global silent=true` 's `silent = true` wont pass to the `lsp_ducument_symbols`
图片

`{ "lsp_document_symbols", desc = "Symbols (buf)", prefix = "@", opts = { silent = true } },` also useless

---

https://github.com/ibhagwan/fzf-lua/issues/2184#issuecomment-3085906990

>Most warnings can be disabled by adding silent=true but picker contents weren’t always included in the silent condition, I can add that easily.

maybe related to this?

---

with this pr, `silent = true` will correctly passed:

图片

ibhagwan:
Got it, ty @Parsifa1

PR #2283, opened 2025-08-29
fix(previewer): re-detect filetype on poor filetype

phanen:
Example: `/path/to/xx.txt` with modeline in it saying it's `cmake`.


ibhagwan:
Ty @phanen !

PR #2286, opened 2025-08-30
traceback

phanen:
- **fix(live_grep_st): slient_fail need to be apply again if we transform the cmd**
- **fix: reuse xpcall to get full traceback**

Now I know why it used to be xpcall, since pcall truncated traceback...
https://www.lua.org/pil/8.5.html


ibhagwan:
Is this ready to merge?

phanen:
y

PR #2287, opened 2025-08-30
feat!: default profile set to `border-fused+hide`

ibhagwan:


ibhagwan:
Breaking but not too breaking, easily reversible, what do you think @phanen?

phanen:
good

Issue #2010, opened 2025-04-24
Feature: forward query to (live) binary in space-separated fashion


aktau:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

I'm trying to use `live_grep` with a custom command. This command accepts a query as such:

```bash
$ mycommand --opt1 --opt2 -- queryatom1 queryatom2 queryatom3
```

But when I do the following:

```
require("fzf-lua").live_grep({ cmd = table.concat({ "mycommand", "--opt1", "--opt2", "--" } , " ") })
```

And I type `queryatom1 queryatom2 queryatom3`. It actually sends the following:

```
$ mycommand --opt1 --opt2 -- 'queryatom1 queryatom2 queryatom3'
```

This is different from what I want. The command interprets the latter as "this exact phrase", while the format is "at least one reference to each of the atoms".

I tried reading the docs and the source code, and eventually was able to hack a way to do this:

```diff
M.fzf_field_expression = function(opts)
  -- fzf already adds single quotes around the placeholder when expanding.
  -- for skim we surround it with double quotes or single quote searches fail
-  return opts and opts._is_skim and [["{}"]] or "{q}"
+  return opts and opts._is_skim and [["{}"]] or "$FZF_QUERY"
end
```

I got this idea from https://github.com/junegunn/fzf/issues/4033. I haven't found a way to configure this from the options, and I also can't keep this as a local hack because it (predictably) makes the default live grep with `rg` (and probably lots of other things) fail.

Did I just miss something in the docs? In that case, I apologize. It would be great to have an example of this though.

ibhagwan:
Use a custom `rg_glob_fn`: https://github.com/ibhagwan/fzf-lua/wiki#live-grep-custom

Or `fn_transform_cmd`: https://github.com/ibhagwan/fzf-lua/issues/1927#issuecomment-2764743237

Add `debug=1` to your opts and you’ll be able to see your generated command in the first line.

ibhagwan:
Btw, if you’re just looking for glob support, you can do this out of the box, for example search for `foo -- *.lua *.md`

https://github.com/ibhagwan/fzf-lua/wiki#can-i-use-ripgreps---globiglob-option-with-live_grep

aktau:
> Btw, if you’re just looking for glob support, you can do this out of the box, for example search for foo -- *.lua *.md

[...]

PR #2288, opened 2025-08-30
support function for multiprocess

phanen:
- **refactor: check_upvalue**
- **chore: unused**
- **fix: safe=false**
- **fix: support function for opts.multiprocess**
- **test?**

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -116,7 +116,7 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
     -- })
   end
 
-  if opts.debug then
+  if opts.debug and type(opts.debug) ~= "number" then
     utils.info("FZF_DEFAULT_COMMAND: %s", tostring(FZF_DEFAULT_COMMAND))
     utils.info("fzf cmd: %s", table.concat(cmd, " "))
   end
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -412,14 +412,14 @@ M.spawn_stdio = function(opts)
   local argv = function(i)
     local idx = tonumber(i) or #_G.arg
     local arg = _G.arg[idx]
-    if opts.debug == "v" or opts.debug == "verbose" then
+    if opts.debug == "v" or opts.debug == 2 then
       io.stdout:write(("[DEBUG] raw_argv(%d) = %s" .. EOL):format(idx, arg))
     end
     -- TODO: maybe not needed anymore? since we're not using v:argv
     if FzfLua.utils.__IS_WINDOWS then
       arg = M.unescape_fzf(arg, FzfLua.utils.has(opts, "fzf", { 0, 52 }) and 0.52 or 0)
     end
-    if opts.debug == "v" or opts.debug == "verbose" then
+    if opts.debug == "v" or opts.debug == 2 then
       io.stdout:write(("[DEBUG] esc_argv(%d) = %s" .. EOL):format(idx, M.shellescape(arg)))
     end
     return arg
@@ -445,7 +445,7 @@ M.spawn_stdio = function(opts)
     opts.no_ansi_colors = true
   end
 
-  if opts.debug == "v" or opts.debug == "verbose" then
+  if opts.debug == "v" or opts.debug == 2 then
     for k, v in vim.spairs(opts) do
       io.stdout:write(string.format("[DEBUG] %s=%s" .. EOL, k, vim.inspect(v)))
     end
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -63,7 +63,7 @@ local function load_config_section(s, datatype, optional)
       ok, res = pcall(loadstring, res)
     end
     ---@diagnostic disable-next-line: undefined-field
-    if _G._debug == "v" or _G._debug == "verbose" then
+    if _G._debug == "v" or _G._debug == 2 then
       ---@diagnostic disable-next-line: undefined-field
       io.stdout:write(("[DEBUG] [load_config] %s = %s" .. (_G._EOL or "\n"))
         :format(s, not ok and errmsg or res))


phanen:
Not sure why windows ci fail on DEBUG line... why it didn't fail before...

ibhagwan:
> Not sure why windows ci fail on DEBUG line... why it didn't fail before...

Strange indeed, should be ignored in screen_opts anyways

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2288/commits/8e7cb0ae144d6e79bd668600697ce6202f84bc05 - seems so :)

ibhagwan:
Pretty cool stuff what you did here

phanen:
... because I mistakenly set argv_expr in fzf_exec, which then set fn_preprocess

ibhagwan:
Not rebasing this one yet as I’m not sure if you wish to rebase this on main or on #2301

phanen:
before just to check the ci. rebase to main now.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2288/commits/f3fcb9bfde3d272bda43b5ed6b7b866dcba0f9f4 - I thought it was just the escaping that was wrong on Window (using the wrong slash) but it seems it’s a different issue, will remove this commit.

ibhagwan:
Rebased since https://github.com/ibhagwan/fzf-lua/commit/3d2cbb8e3b02d371fcc341fa61fa3bd18ce372ab was merged into main in #2314

lua/fzf-lua/shell.lua
```diff
@@ -222,6 +222,7 @@ M.stringify_mt = function(contents, opts)
       "fn_transform",
       "fn_preprocess",
       "fn_postprocess",
+      "fn_reload", -- TODO: just a boolean, maybe rename it
```
ibhagwan: Maybe `_is_live` or without underscore? Or anything else better you can think of

ibhagwan:
Also removed c8f697f4d4f75cb984cf8fad524ead09fd996a48 due to https://github.com/ibhagwan/fzf-lua/commit/31b671350d16934b0a8a8748a91d23dd86ee8369.

lua/fzf-lua/shell.lua
phanen: https://github.com/ibhagwan/fzf-lua/pull/2288/commits/adf6ee840d1e767bc0b7b30ba1d01377900db190

ibhagwan:
Rebased to main

ibhagwan:
Also modified https://github.com/ibhagwan/fzf-lua/pull/2288/commits/3f5b9186876277cf463617e40f14591d37640598 to adjust for new border calc on main (not sure why layout was different with multiprocess=1)

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2288/commits/f775fe7a9e782b3486601729168e7be732bad387

Installed a Windows VM, the reason the CI was failing is due to the native `exec_empty_query` condition (which I could never fix with all the special chars, maybe worth a revisit or to remove altogher):
IMG_3647

ibhagwan:
My above commit "fix" is just a workaround, the NEQ shouldn't be set at all when fzf_live contents is a function since `exec_empty_query` is tested in the function.

Also due to this PR, `live_grep_native` isn't really native anymore as it uses a contents function, technically we want native to use `rg {q}` directly in the `reload` statement with zero callbacks to neovim, this way we eliminate any code routes into fzf-lua, for that we need to still use a string command so that's it's the equivalent of `rg <initial search> | fzf --bind=change:reload:rg <args> {q}`.

I personally think we should remove the `no_query_condi` and document it as a "known limitation" for the native version, we can still ignore the initial query on the first spawn so if the user runs `:FzfLua live_grep_native exec_empty_query=false` the interface opens with an empty query but if they type one char+backspace it will execute the empty query - what do you think?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/ffa52d3ffbccdd7e4c268ed8544b658569b53b58 - removed `no_query_condi`, still need to restore `live_grep_native` to be "native" again.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/b94ca6157dc888cb5687d292d022c01ade2a25f7 - restored live_grep_native to string command, exec_empty_query only takes effect on picker open

ibhagwan:
~~I’m not sure I’m getting the functionality of `core.can_transform` now, it seems to replace `reload:cmd` with `transform:reload:cmd` which is essentially the same…

IIRC you made it for performant rg_glob with multiprocess=false and no icons, maybe it’s not working as expected post this PR?~~

I remember now, it removes the need to run through the neovim headless wrapper by reloading with rg output directly.

phanen:
> I remember now, it removes the need to run through the neovim headless wrapper by reloading with rg output directly.

Yeah. in mt, posix_exec is used, in st, transform:reload is used

phanen:
We can also add parametrize multiprocess to top of grep_spec/file_spec if we want...

Edit:
after review  i found most sceenshots are the same, maybe not to put in top but only limit to content related scope

phanen:
exec_empty_query don't seem work now, it just silent_fail with a `{q}` query

set `--pcre2` in rg_opts trigger exec_emptry_query

a minimal repro
```
require('fzf-lua').fzf_live('echo <query>')
```

ibhagwan:
Rebased on main, how do you feel about merging this soon or do you have more refactoring plans?

phanen:
no more now.

one thing i didnt test if fn_transform_cmd still work in mt/st

ibhagwan:
> no more now.
> 
> one thing i didnt test if fn_transform_cmd still work in mt/st

`rg_glob_fn` works in both st/mt, `fn_transform_cmd`  works in st, ~~doesn't work in mt:~~

**Idk what happened, wasn’t consistent, sometimes worked in mt sometimes didn’t, had to go AFK so will continue testing later**

> Example from #1927

```lua
require("fzf-lua").live_grep({
  debug = true,
  multiprocess = false,
  previewer = false,
  fzf_opts = { ["--wrap"] = true },
  no_esc = true,
  search =  [[cop -- LIC*]],
  cmd = "git grep -i --line-number --column --color=always",
  fn_transform_cmd = function(query, cmd, _)
    local search_query, glob_str = query:match("(.-)%s-%-%-(.*)")
    if not glob_str then return end
    local new_cmd = string.format("%s %s %s", cmd, vim.fn.shellescape(search_query), glob_str)
    return new_cmd, search_query
  end
})
```

> St
image

nenahp:
just gussing maybe have to write in setup(opts)

ibhagwan:
> just gussing maybe have to write in setup(opts)

I’ll test more throughly later, in any event, should work in both setup and in direct call.

phanen:
for some reason `load_config` don't get fn_transform_fn

ibhagwan:
> for some reason `load_config` don't get fn_transform_fn

I see, the load was deleted in the refactor

ibhagwan:
reworded your commit to reflect the fix.

https://github.com/ibhagwan/fzf-lua/pull/2288/commits/a090590aba0e2375a36a85319681c165db885b42 - so we can `FzfLua live_grep debug=1` and avoid printing the command into `:messages` (if you just want to see the fzf debug line - this is how it worked before).

ibhagwan:
For some reason load_config for `fn_transform_cmd` is called twice, we should probably refactor loading config values in mt/headless:

image

ibhagwan:
`get_grep_cmd` loads the value from config twice, that's 1 rpc call we can avoid:
```lua
M.get_grep_cmd = function(opts, search_query, no_esc)
  opts = _G._fzf_lua_is_headless and setmetatable(vim.deepcopy(opts), { __index = opts2 }) or opts
  ...
  if opts.fn_transform_cmd then
    local new_cmd, new_query = opts.fn_transform_cmd(search_query, command, opts)
```

ibhagwan:
Tested this PR some more works, great, few issues we can improve on (but aren't critical):
- Loading `fn_transform_cmd` twice over RPC (perhaps refactor loading values over RPC)
- No upvalue alert for opts.cmd/opts.contents, e.g.:
```lua
local up = { "a" }
require 'fzf-lua'.fzf_exec(function(fzf_cb)
  for i = 1, 10 do
    fzf_cb(tostring(i) .. up[1])
  end
  fzf_cb()
end, { debug = 1, multiprocess = true })
```

Saw you already left a comment in the code about this, this diff seems to work properly not sure if there's any downside?
```diff
diff --git a/lua/fzf-lua/shell.lua b/lua/fzf-lua/shell.lua
index b055144..586bcf7 100644
--- a/lua/fzf-lua/shell.lua
+++ b/lua/fzf-lua/shell.lua
@@ -518,7 +518,7 @@ M.wrap_spawn_stdio = function(opts)
   local is_win = utils.__IS_WINDOWS
   local nvim_bin = os.getenv("FZF_LUA_NVIM_BIN") or vim.v.progpath
   -- TODO: should we check "cmd"?
-  for _, k in ipairs({ "fn_transform", "fn_preprocess", "fn_postprocess" }) do
+  for _, k in ipairs({ "cmd", "contents", "fn_transform", "fn_preprocess", "fn_postprocess" }) do
     if type(opts[k]) == "function" then
       -- opts[k] = M.check_upvalue(opts[k], "opts." .. k)
       M.check_upvalue(opts[k], "opts." .. k)
```

In any event, none of the above is urgent so I'm going to merge this.

ibhagwan:
Also, maybe we should refactor `opts.cmd|opts.contents` and combine into one argument? I find it confusing to have both now :-)

`contents` is more fitting as contents can be a shell cmd / function /table.

aktau:
161f9c7640b2571a9c28d37921a7b9c30fd1617b broke my use case (see #2010), so it is likely not a pure refactor . I have the following setup:

```lua
  vim.keymap.set('n', '<leader>cs', function()
      require("fzf-lua").live_grep({
        cmd = "/usr/bin/ugcsearch --",  -- The '--' prevents interpreting query fragments as options.
        field_index = "$FZF_QUERY",
        cwd = findroot(vim.uv.cwd()),
        filter = 'sed s@/mnt@@',
      })
    end,
    { desc = "search code" })
```

Before this commit, search terms were being sent as different argv items to `cmd`. After this commit, they are all in a single arg.

Using execsnoop, before:

```
ugcsearch          354080 354077   0 /usr/bin/ugcsearch -- f:dir1 f:dir2
```

After:

```
nvim             354825 354823   nvim -u NONE -l $HOM/.local/share/nvim/site/pack/core/opt/fzf-lua/lua/fzf-lua/spawn.lua return [==[e2NvbG9yX2ljb25zID0gdHJ1ZSwgZm5fcmVsb2FkID0gdHJ1ZSwgZ2l0X2ljb25zID0gZmFsc2UsIHJnX2dsb2IgPSBmYWxzZSwgY29udGVudHMgPSAo f:dir1 f:dir2
sh               354825 354823   0 /bin/sh -c /usr/bin/ugcsearch -- 'f:dir1'  | se
ugcsearch          354826 354825   0 /usr/bin/ugcsearch  -- f:dir1
```

For some reason, the second argument (`f:dir2`) is just clipped off now. There also appear to be extra processes in the game.

I suspect native mode is getting disabled somehow (I've noticed that `$FZF_QUERY` used to not work right if I didn't apply the following configuration):

```lua
    require("fzf-lua").setup({
      defaults = {
        -- Setting the following to false leads to better performance. It
        -- enables the "native" mode (no interposition) even when the
        -- non-native variants (e.g.: live_grep) are executed.
        file_icons = false,
        git_icons = false,
      },
      grep = {
        rg_glob = false, -- Inhibits native mode, and I never use it.
      },
      fzf_colors = {
        true, -- It looks ugly without this.
      },
    })
```

So this change (native -> non-native) likely triggered that other bug that I had been successfully avoiding with native mode (just my theory).

aktau:
I'll file a new issue about this, that's easier. 

Update: created #2343

PR #2291, opened 2025-08-30
fix: missing M for alt in conv_map

shushtain:
Many people use `<M-` in Neovim instead of `<A-`, and this PR ads support for that, solving [this issue](https://github.com/ibhagwan/fzf-lua/discussions/2290#discussioncomment-14264048).


ibhagwan:
Ty @shushtain!

PR #2292, opened 2025-08-30
fix(actions): switch to correct buffer window on special set buf autocmd

eyalz800:
Hi @ibhagwan,

I encountered an edge case where if vim.api.nvim_win_set_buf results in autocmd that eventually decides to open the buffer elsewhere, the current window is switched back to the place where it was, I don't know if that's a neovim bug or something that is by design, here is my attempt to fix this by checking whether the current window contains the buffer, and if not - switch to the window where the buffer is.

The use case for such autocmd to decide to open the buffer elsewhere is for example if the buffer is about to open in like a file explorer or something. The winfixbuf doesnt resolve this nicely, this is what I am doing:

https://github.com/eyalz800/nvim.lua/blob/main/lua/plugins/pin.lua

What happens is that nvim_win_set_buf triggers my open_in_best_win function which opens the buffer in a different window and sets as current window, but somehow nvim_win_set_buf by fzf-lua returns with the current window being the previous window.

I hope you can review and accept/suggest a different fix.

Thank you for this awesome plugin!

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -215,10 +215,10 @@ M.vimcmd_entry = function(_vimcmd, selected, opts)
             then
               vim.bo.bufhidden = "wipe"
             end
-            -- NOTE: nvim_win_set_buf will load the buffer if needed
+            -- NOTE: nvim_set_current_buf will load the buffer if needed
             -- calling bufload will mess up `BufReadPost` autocmds
             -- vim.fn.bufload(bufnr)
-            local ok, _ = pcall(vim.api.nvim_win_set_buf, 0, bufnr)
+            local ok, _ = pcall(vim.api.nvim_set_current_buf, bufnr)
             -- When `:set nohidden && set confirm`, neovim will invoke the save dialog
             -- and confirm with the user when trying to switch from a dirty buffer, if
             -- user cancelles the save dialog pcall will fail with:


eyalz800:
This seems to happen after commit 519f765d where nvim_win_set_buf was introduced

ibhagwan:
Tysm @eyalz800!

The fix seems good, I’m trying to think  if there are any drawbacks to this approach (opening buffer from Nvim-tree, etc)?

Also wondering why using `:e` and `:b` didn’t have this issue.

eyalz800:
It didnt have issue before this change I posted the commit that started this which is https://github.com/ibhagwan/fzf-lua/commit/519f765d7c1782d70652b3187bf466b773b8e86b

eyalz800:
@ibhagwan this could be a neovim bug, nvim_win_set_buf seems to change the current window to the current window before the call, I guess, that's the only thing I can come up with. 

Regarding any drawbacks- I could not think of any, this code will execute only if the current buffer is not the buffer you set for the current window, in which case the logic that follows will start to set the cursor position in the wrong window where the buffer we wanted is not even loaded.

A few possible outcomes:
1. This is indeed a neovim bug, and we need to submit a ticket and meanwhile workaround the bug by something like this PR until the neovim bug is resolved.
2. It's not a neovim bug, shouldn't use nvim_win_set_buf because it changes the current window back to where we called it from. Revert back to the previous buffer switching that doesn't change back the window, or settle for working around that API in the suggested way / a better way.
3. Regardless of neovim bug or not, keeping things as is - if the buffer opens in a different window - then the rest of the logic is going to change cursor positions in the current window which doesnt have the buffer. - I think this is less acceptable and is going to make me use an old version or patch it myself and maintain a fork.

eyalz800:
I am not really familiar with neovim code but at least by the looks it seems to restore the window in the end of the nvim_win_set_buf, so this seems almost by design that we see the window switch back after the do_buffer:
https://github.com/neovim/neovim/blob/1b3abfa688865d95b85c27d3e523c087669604a2/src/nvim/window.c#L751
```cpp
/// Sets the current buffer in a window, without side effects
///
/// @param window   |window-ID|, or 0 for current window
/// @param buffer   Buffer id
/// @param[out] err Error details, if any
void nvim_win_set_buf(Window window, Buffer buffer, Error *err)
  FUNC_API_SINCE(5)
  FUNC_API_TEXTLOCK_ALLOW_CMDWIN
{
  win_T *win = find_window_by_handle(window, err);
  buf_T *buf = find_buffer_by_handle(buffer, err);
  if (!win || !buf) {
    return;
  }

  if (win == cmdwin_win || win == cmdwin_old_curwin || buf == cmdwin_buf) {
    api_set_error(err, kErrorTypeException, "%s", e_cmdwin);
    return;
  }
  win_set_buf(win, buf, err);
}

void win_set_buf(win_T *win, buf_T *buf, Error *err)
  FUNC_ATTR_NONNULL_ALL
{
  const handle_T win_handle = win->handle;
  tabpage_T *tab = win_find_tabpage(win);

  // no redrawing and don't set the window title
  RedrawingDisabled++;

  switchwin_T switchwin;
  int win_result;

  TRY_WRAP(err, {
    win_result = switch_win_noblock(&switchwin, win, tab, true);
    if (win_result != FAIL) {
      const int save_acd = p_acd;
      if (!switchwin.sw_same_win) {
        // Temporarily disable 'autochdir' when setting buffer in another window.
        p_acd = false;
      }

      do_buffer(DOBUF_GOTO, DOBUF_FIRST, FORWARD, buf->b_fnum, 0);

      if (!switchwin.sw_same_win) {
        p_acd = save_acd;
      }
    }
  });
  if (win_result == FAIL && !ERROR_SET(err)) {
    api_set_error(err, kErrorTypeException, "Failed to switch to window %d", win_handle);
  }

  // If window is not current, state logic will not validate its cursor. So do it now.
  // Still needed if do_buffer returns FAIL (e.g: autocmds abort script after buffer was set).
  validate_cursor(curwin);

  restore_win_noblock(&switchwin, true);
  RedrawingDisabled--;
}
```
So it's up to us whether we just want to switch buffers with a buffer command which follows the buffer for us, or if we use nvim_win_set_buf then we need to follow that buffer with our own logic into the new window where the buffer is opened. @ibhagwan it's your decision now :)

eyalz800:
Im going to check if using this is better, it seems to not try to restore the window back:
```cpp
/// Sets the current window's buffer to `buffer`.
///
/// @param buffer   Buffer id
/// @param[out] err Error details, if any
void nvim_set_current_buf(Buffer buffer, Error *err)
  FUNC_API_SINCE(1)
  FUNC_API_TEXTLOCK
{
  buf_T *buf = find_buffer_by_handle(buffer, err);

  if (!buf) {
    return;
  }

  TRY_WRAP(err, {
    do_buffer(DOBUF_GOTO, DOBUF_FIRST, FORWARD, buf->b_fnum, 0);
  });
}
```

eyalz800:
@ibhagwan it seems that the correct api to use here is indeed vim.api.nvim_set_current_buf, I've updated the pull request to do so, that way we don't experience switching back to the current window.

ibhagwan:
What a research and solution @eyalz800! :)

`vim.api.nvim_set_current_buf` is much better indeed and what I thought `nvim_win_set_buf(0, …)` would be.

eyalz800:
Thank you @ibhagwan and please keep up the amazing work!

ibhagwan:
> Thank you @ibhagwan and please keep up the amazing work!

Peace be upon you אח 🙏❤️☺️

PR #2298, opened 2025-09-01
fix(actions): vim.cmd.drop on grep result

phanen:


ibhagwan:
TBH, we can probably ditch the `<auto>`  logic now, it was only meant to use `:e` or `:b` if bufnr was available but since we now use bufadd / set_current_buf it’s not required anymore.

ibhagwan:
`| <auto>` is now removed anyways, maybe we can use different logic or an `opts._` var for bufadd/set_buf instead?
https://github.com/ibhagwan/fzf-lua/blob/091bcd0c925199b1371ea81ec79278611d299627/lua/fzf-lua/actions.lua#L179

lua/fzf-lua/actions.lua
```diff
@@ -263,29 +263,28 @@ end
...
-  local vimcmd = "tabnew | setlocal bufhidden=wipe | <auto>"
-  M.vimcmd_entry(vimcmd, selected, opts)
+  opts._is_buf_edit = true
+  M.vimcmd_entry("tabnew | setlocal bufhidden=wipe", selected, opts)
 end
 
 M.file_open_in_background = function(selected, opts)
-  local vimcmd = "badd"
-  M.vimcmd_entry(vimcmd, selected, opts)
+  opts._is_buf_edit = true
```
ibhagwan: I’m not sure if `badd` needs buf edit, this will change the buffer in the current win no?

lua/fzf-lua/actions.lua
```diff
@@ -163,72 +210,25 @@ M.vimcmd_entry = function(_vimcmd, selected, opts)
...
+        vim.cmd(("%s %s"):format(vimcmd, relpath))
+      elseif not entry.uri then
+        -- <auto> (without prefix, formerly `:b|e`) replace the current buffer
+        local will_replace_curbuf = #vimcmd == 0
+        if will_replace_curbuf and utils.wo.winfixbuf then
+          utils.warn("'winfixbuf' is set for current window, will open in a split.")
+          vimcmd, will_replace_curbuf = "split", false
         end
+        if #vimcmd > 0 then vim.cmd(vimcmd) end
+        edit_entry(entry, fullpath, will_replace_curbuf, opts)
```
ibhagwan: We need to bail here if vim.api.nvim_set_current_buf failed as this means there was a dirty buffer with `vim.o.confirm=true` and the user cancelled the save dialog.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2298/commits/8d870a632366e2ac00f1cb640f38dd0955ae0b30

Quite a bit of edge cases with `set confirm | set nohidden` & `set noconfirm | set nohidden` as well as URIs, couldn't get away from having to dup the code to test if the requested buffer is already loaded.

phanen:
if set to the third arg, this will broken..
function(...) return action.xx(...) end

maybe there's no reason to wrap a single function though

user can just change few line of code

ibhagwan:
> if set to the third arg, this will broken.. function(...) return action.xx(...) end

I’m not sure why, 3rd arg won’t be sent this way unless they call `vimcmd_entry(…)`? 

> maybe there's no reason to wrap a single function though
>
> user can just change few line of code

I’m fine with whatever you feel is best, most users don’t get to this level of customization, although it’s nice that’s it’s working I don’t consider `vimcmd_entry` an API.

phanen:
yeah, unless user call vimcmd_entry
when call vimcmd_entry, i guess this also make sense since user don't want buf edit beahvior, so it's fine.

ibhagwan:
Lmk if you feel this is done I can merge it

phanen:
fine, I want to add some test later, but before I want to improve the test framework...

phanen:
https://github.com/ibhagwan/fzf-lua/pull/2301

ibhagwan:
Alright I’ll merge it then

PR #2301, opened 2025-09-02
nostring

phanen:
- **test: split to more test**
- **test: add testnvim.exec_lua**
- **test: serialize**

### DIFF
--- a/lua/fzf-lua/core.lua
+++ b/lua/fzf-lua/core.lua
@@ -228,7 +228,6 @@ M.fzf_resume = function(opts)
     return
   end
   opts = utils.tbl_deep_extend("force", config.__resume_data.opts, opts or {})
-  utils.set_info(opts.__INFO) -- restore original picker info
   assert(opts == config.__resume_data.opts)
   opts.cwd = opts.cwd and libuv.expand(opts.cwd) or nil
   M.fzf_wrap(config.__resume_data.contents, config.__resume_data.opts)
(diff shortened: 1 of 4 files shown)


phanen:
* print in exec_lua will be populated to parent instance
* exec_lua can use simple upvalue for simple datatype

~TODO: support simple upvalue for args~ done

phanen:
cannot use this method for fzf_live...
since the child is just a relay for "live content"...
`test runner -> child -> headless wrapper`
the upvalue will be dropped in the second sent..

phanen:
rebase on https://github.com/ibhagwan/fzf-lua/pull/2298.

phanen:
It seems working well now, as long as we don't use upvalue in multiprocess contents.

ibhagwan:
Rebased on main

ibhagwan:
I’ll hopefully test this later today, would you like me to merge this or are there more changes planned?

phanen:
ok. no more change

ibhagwan:
Ty @phanen, this is really fantastic :)

ibhagwan:
Oops, not sure why yet, this messes up `:FzfLua global line_query=true` (the default, works with files though).

ibhagwan:
I think it’s because the opts ref is different now in the previewer/action and the new path.entry_to_file line condition doesn’t work, it’s tricky because each picker has it’s own opts table.

PR #2306, opened 2025-09-04
test: only :drop in on drop test

phanen:
"drop" shouldn't be always executed...


phanen:
That's explaining why the screenshots looks strange to me. `file_open_in_background` should not load into visible buffer.

PR #2308, opened 2025-09-04
zf

phanen:
`:FzfLua files profile=zf`

- **chore: fzf_exec don't have info now**
- **feat(zf): support zf**

some zf limit
* don't have `{q}`/`{n}`, so it seems also previewer is unusable.
* no custom key bindings
* no ansi parsing in list, also no parsing in preview term

https://github.com/ibhagwan/fzf-lua/discussions/2307


phanen:
Not sure if we really need to support this, it has few flags.

Maybe https://github.com/alexpasmantier/television looks more interesting.

ibhagwan:
> Not sure if we really need to support this, it has few flags.
> 
> Maybe https://github.com/alexpasmantier/television looks more interesting.

I looked into television before and deemed it wasn’t ready and would require too much work.

lua/fzf-lua/utils.lua
```diff
@@ -1401,6 +1401,11 @@ function M.sk_version(opts)
   return M.parse_verstr(out), rc, out
 end
 
+function M.zf_version(opts)
+  local out, rc = M.io_system({ opts and opts.fzf_bin or "zk", "--version" })
```
ibhagwan, line 1405: Is this a typo? Shouldn’t this be `zf` not `zk`?

lua/fzf-lua/config.lua
```diff
@@ -689,13 +690,20 @@ function M.normalize_opts(opts, globals, __resume_key)
...
-  else
+  elseif opts._is_skim then
     local SK_VERSION, rc, err = utils.sk_version(opts)
     opts.__SK_VERSION = SK_VERSION
     if not opts.__SK_VERSION then
       utils.error("'sk --version' failed with error %s: %s", rc, err)
       return nil
     end
+  elseif opts._is_zf then
+    local ZF_VERSION, rc, err = utils.sk_version(opts)
```
ibhagwan, line 701: sk_version of zf_version?

ibhagwan:
Ty @phanen!

The question is, do we really want to support yet another fzf wannabe project with barely any features and last commit 5 months ago?

If we support this we will have to fix issues if this doesn’t work as expected or make all kinds of code hacks to make the bare necessity work and then get stuck with code we cannot remove which holds us back from  exciting new features in fzf itself.

I would be more inclined to say yes if this was something new and exciting but this seems to be a somewhat stalled project, maybe a finder with typo resistant features?

lua/fzf-lua/utils.lua
phanen: yes should be zf

lua/fzf-lua/config.lua
phanen: so zf version may not needed

phanen:
yeah after finish it I don't think we should support it now since all fzf-lua enhance dont even work

it's easy to write this as another plugin with few line of code

ibhagwan:
> yeah after finish it I don't think we should support it now since all fzf-lua enhance dont even work
> 
> it's easy to write this as another plugin with few line of code

If the project was showing a lot of activity and bright future I would also say yes but nothing currently competes with the development speed of fzf

Issue #2315, opened 2025-09-06
Bug: Window sizing miscalc for results when winopts.border=false


nhurlock:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Mac

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-339+gc2d218d0c6

### Fzf version (`fzf --version`)

0.65.2 (brew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color=bg+:#363a4f,bg:#24273a,spinner:#f4dbd6,hl:#ed8796 --color=fg:#cad3f5,header:#ed8796,info:#c6a0f6,pointer:#f4dbd6 --color=marker:#b7bdf8,fg+:#cad3f5,prompt:#c6a0f6,hl+:#ed8796 --color=selected-bg:#494d64 --info=inline-right --padding=1 --bind='ctrl-j:down' --bind='ctrl-k:up' --bind='ctrl-u:preview-half-page-up' --bind='ctrl-d:preview-half-page-down' --bind='ctrl-p:toggle-preview' --bind='ctrl-z:abort' --bind='ctrl-w:backward-kill-word' --bind='ctrl-b:beginning-of-line' --bind='ctrl-e:end-of-line' --bind='alt-a:toggle-all' --border=none --reverse --multi --wrap

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
  winopts = {
    border = false
  }
})
```

### Describe the bug / steps to reproduce

The results sizing seems to be miscalculated in comparison to the preview when `winopts.border = false`.
This change was introduced in commit d783364

before (commit: 8ebb4f5):
<img width="1882" height="1261" alt="Image" src="https://github.com/user-attachments/assets/017001a6-6059-4484-93cb-e161e31e5588" />

after:
<img width="1878" height="1257" alt="Image" src="https://github.com/user-attachments/assets/79ac989d-58b5-4608-8833-851c147da0bb" />

ibhagwan:
[...]

PR #2316, opened 2025-09-07
fix(borderless-full): pad fzf win to align with previewr

phanen:
fix https://github.com/ibhagwan/fzf-lua/issues/2315#issuecomment-3263230118.


phanen:
didt notice ci fail, i will fix later

ibhagwan:
Tysm @phanen for fixing this and adding the ci :heart:

PR #2320, opened 2025-09-08
run fzf-lua in cli

phanen:
**EDIT from snacks.gh credit to folke:**

https://github.com/ibhagwan/fzf-lua/discussions/2319

(currently `nvim -l` won't load your `init.lua`)
```sh
nvim -l scripts/cli.lua files
nvim -l scripts/cli.lua live_grep
nvim -l scripts/cli.lua serverlist
```

Use the first config in the order of:
* `$XDG_CONFIG_HOME/fzf-lua.lua`
* `$XDG_CONFIG_HOME/fzf-lua/init.lua`
* `$XDG_CONFIG_HOME/fzf-lua/config.lua`

config example:
```lua
require('fzf-lua').setup({
  { 'cli' }, -- inherit cli profile
  keymap = {},
  actions = {},
  files = {},
  -- ...
})

-- add new custom picker
FzfLua.new_picker = function()
  return FzfLua.fzf_exec(function(cb)
    cb('new')
    cb('picker')
    cb(nil)
  end, {
    preview = 'echo this is a new picker',
    actions = {
      ['enter'] = function()
        if not s[1] then return end
        io.stdout:write(s[1])
        vim.cmd.qall()
      end,
    },
  })
end
```

### DIFF
--- a/scripts/cli.lua
+++ b/scripts/cli.lua
@@ -14,14 +14,6 @@ if vim.v.servername and #vim.v.servername > 0 then
   pcall(fn.serverstop, vim.v.servername)
 end
 
-api.nvim_create_autocmd("Signal", {
-  callback = function(ev)
-    vim.tbl_map(function(pid)
-      FzfLua.libuv.process_kill(pid, ev.match)
-    end, api.nvim_get_proc_children(fn.getpid()))
-  end
-})
-
 require("fzf-lua").setup({ "cli" })
 
 -- load user config


ibhagwan:
Since we can derive the location of fzf-lua from `scripts/cli.lua` (basically ..) - wouldn’t this be better used with `nvim -u NONE`? This way we can eliminate any conflict with the users' config and make this script independent

phanen:
y -l seems better

ibhagwan:
> y -l seems better

I agree, in any event we can add the root dir to pack path and run it without the user config

phanen:
`-l` work now, `uv.run` seems block infinitely.

`vim.wait(0)` is documented
```
What we consider the "Nvim event loop" is actually a wrapper around `uv_run` to
handle both the `fast_events` queue and possibly (a suitable subset of) deferred
events. Therefore "raw" `vim.uv.run()` is often not enough to "yield" from Lua
plugins; instead they can call `vim.wait(0)`.
```

https://github.com/neovim/neovim/blob/442f297c633375074a88133a9b3d8c08cad58dd0/src/nvim/event/loop.c#L74-L76

https://github.com/neovim/neovim/blob/b986fd96bc8d9ef8739efd9c5c4221f04a2c3463/src/nvim/lua/executor.c#L495-L502

basically it's `while true do uv.run('nowait') end` + dummy timer. I'm not sure if they are different here but both work.

ibhagwan:
This is pretty cool and very fast too, I am able to have live_grep with fuzzy toggle without any hassle in one command:
IMG_3752
IMG_3753

ibhagwan:
For customizing, we can read config options from `$XDG_CONFIG_HOME/fzf-lua/config.lua` and make this a full fledged replacements for [my fzf shell scripts](https://github.com/ibhagwan/dots) :)

phanen:
I move cli to a new profile. Add a serverlist picker (https://github.com/ibhagwan/fzf-lua/discussions/2409).

ibhagwan:
Wow @phanen, this is really nice! Thank you so much.

Btw, I’ve been thinking of changing LSP to https://github.com/EmmyLuaLs/emmylua-analyzer-rust - what do you think about it? At first test it seems annotations transfer almost as is? Might have to reformat or maybe just move formatting to stylua and call it a day?

phanen:
make types gotodef work in dotfile，i update readme in last commit

ibhagwan:
Another unrelated question - which AI LM+plugin do you use and do you find it helpful? It seems to fail more complicated tasks but maybe I haven’t given it enough chance or using it wrong

phanen:
i tried emmylua before but didnt get it work correctly but iirc it use the same formatter/format as luals

ibhagwan:
> i tried emmylua before but didnt get it work correctly but iirc it use the same formatter/format as luals

Yes I also tried it in the past but I think it may have made a decent leap in the last couple of months

phanen:
> Another unrelated question - which AI LM+plugin do you use and do you find it helpful? It seems to fail more complicated tasks but maybe I haven’t given it enough chance or using it wrong

mainly codecompanion with mcphub.nvim recently, although most of time i just prompt it to do some quick inline fix...

phanen:
~~Not sure how to make emmylua-check load the dependencies.~~

phanen:
emmylua-check is added to lint ci.

ibhagwan:
> emmylua-check is added to lint ci.

Wow, tysm @phanen for this incredible work

ibhagwan:
I must say emmylua-rust feels much faster than lua_ls

phanen:
preview image will flicker when u repeatly enter the same entry. I'm not sure if the expected behavior of `do_not_cache`. But at least I found for kitty it can work with cache.

phanen:
I'm so stupid to debug windows ci... I realize it's a known windows issue. `{q}` will be replaced with "" if empty....

ibhagwan:
> I'm so stupid to debug windows ci... I realize it's a known windows issue. `{q}` will be replaced with "" if empty....

Windows is stupid

Let me know if you need any help with the Windows CI I have a Windows VM on my wife’s computer

phanen:
I'm not sure we should fix it here. maybe it's intended? Now I avoid use `{q}` in windows ci now.
I'll open an discussion on upstream to ask for if this behavior is intended.

ibhagwan:
> I'm not sure we should fix it here. maybe it's intended? Now I avoid use `{q}` in windows ci now. I'll open an discussion on upstream to ask for if this behavior is intended.

We can skip it until further clarification from upstream

phanen:
one more entry appear. I recall this happended before.
```
04|~      │> 100                                    301/301 │  
```

ibhagwan:
Or we can inout random string in the query with `query=…` instead (if it doesn’t have the “” issue).

phanen:
~query="" won't fail on master iirc~
field_index is essentially the same as query maybe..

I'll leave now.

ibhagwan:
> one more entry appear. I recall this happended before.
> 
> ```
> 04|~      │> 100                                    301/301 │  
> ```

Correct, I forgot what it was, I recall I made the CI picker scroll to bottom of screen so I could see the last line and it was what helped me fix it.

I’ll try to recall what the issue was

phanen:
`param-type-not-match` rename to `param-type-mismatch` on latest version.

emmylua-check seems to check more error than emmylua-ls.  (but it's flaky)

phanen:
maybe actions should not be deep extended. In this case, field_index is mistakenly inherit.
```lua
      require('fzf-lua').setup {
        actions = {
          files = { ['ctrl-s'] = { fn = file_sel_to_qf, field_index = '{+f}', reload = true } },
        },
        git = {
          status = {
            actions = {
              ['ctrl-s'] = { fn = fa.git_stage_unstage, , reload = true },
            },
          },
        },
      }
```

ibhagwan:
> maybe actions should not be deep extended. In this case, field_index is mistakenly inherit.

You’re right

ibhagwan:
I see  https://github.com/ibhagwan/fzf-lua/pull/2320/commits/ae02f2fc2b17512b83a5f2ea79232cdd8fc0b7e6  reverts fb654cb but I’m assuming ok due to `win.lua:1395: local ctx = utils.__CTX() or {}`?

phanen:
It seems not hard to write a script to generate `OPTIONS.md` via  emmylua_doc_cli. Latest commit is modified from snippets generated by copilot.

ibhagwan:
> It seems not hard to write a script to generate `OPTIONS.md` via emmylua_doc_cli. Latest commit is modified from snippets generated by copilot.

This is amazing @phanen!

phanen:
emmylua resolve types either by inferring or from doc. In the former case it's a bit strict. Even some literal value is treated as a type in json output while it also don't give literal value. We can add type annotation when it infer to a strict type. And I read the default.lua as a fallback to get default value.

Maybe current options.md is already generated from default.lua?

We can adding annotaion to defaults.lua as the single source of truth...
```lua
---@class fzf-lua.config.SpellSuggest: fzf-lua.config.Base
---The pattern used to match the word under the cursor. Text around the cursor position that matches will be used as the initial query and replaced by a chosen completion. The default matches anything but spaces and single/double quotes.
---@field word_pattern? string
M.defaults.spell_suggest         = {
```

`nvim -l scripts/gen_option.lua spellsuggest`:
```
##### spell_suggest.word_pattern

Type: `string?`, Default: `nil`

The pattern used to match the word under the cursor. Text around the cursor position that matches will be used as the initial query and replaced by a chosen completio
n. The default matches anything but spaces and single/double quotes.
```

ibhagwan:
> Maybe current options.md is already generated from default.lua?

Yes we can do that, the current options.md is missing  a lot of options / values.

> We can adding annotaion to defaults.lua as the single source of truth...

I agree, that would be the right place to do add the doc annotations

ibhagwan:
Do you think we can merge this, I realize `gen_options.lua` isn't ready yet but this way we can work on it on the main branch as it's not really affecting anything until we add a workflow for it?

phanen:
yeah, I stop work this recently and don't have much time if next few weeks. can merge it if you want.

ibhagwan:
> Do you think we can merge this, I realize `gen_options.lua` isn't ready yet but this way we can work on it on the main branch as it's not really affecting anything until we add a workflow for it?

Of course, I also don't have a lot of time recently but when I have some time I want to work on `gen_options.lua` -> `OPTIONS.md` conversion.

ibhagwan:
And also start using the cli to replace my custom shell scripts.

ibhagwan:
Forced pushed to restore ci on pull requested (commented in ci.yaml), will merge when all tests pass.

phanen:
forgot gen_options.lua +1 file again, regenerate screenthots...

ibhagwan:
> forgot gen_options.lua +1 file again, regenerate screenthots... 

Crap I just force pushed screenshots too lol

ibhagwan:
I'll let you re-push since I just messed up I still had the non-squashed `c` commit?

phanen:
average flaky

ibhagwan:
Ok to merge now?

phanen:
y. but latest emmylua seems more strict...more lint warnings now
i didnt enable warnings as error for emmylua_check is flaky even when no lsp diagnostics

ibhagwan:
> y. but latest emmylua seems more strict...more lint warnings now i didnt enable warnings as error for emmylua_check is flaky even when no lsp diagnostics

But the ci still uses `make lint` no and we can fix `make emmylua_check` later?

Whatever you prefer, I can leave this unmerged if you think it’s safer for now

phanen:
luals is still in used, which treat warning as error
i can fix all emmyluals wanrnings when back in few weeks, also there seems some bugs in emmyluals to wait for upstream to fix

phanen:
> Whatever you prefer, I can leave this unmerged if you think it’s safer for now

ok let's it open here

ibhagwan:
rebase/resolved the `complete_bline` conflict after https://github.com/ibhagwan/fzf-lua/commit/abad16fa530926b34413f7763cb2c202bcf5b392

ibhagwan:
**EDIT: CI is fine now, no more pushes until tomorrow** 
**EDIT**: the tree is buffed, 7/8 should be further indented below 2 and 13 should be indented below 9

WIP picker for `:Undotree`, still needs work on builtin preview / swiper/ivy preview:
image

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/c4ff8f4a786f403435d27fc868e3cc94982285ed

I believe the tree is accurate now (`:Undotree` on the right):
image

ibhagwan:
@phanen, finished the undotree plugin, pretty cool , with the builtin you can press `F8` to toggle the file view so it behaves like `:Undotree` (without modifying the buffer:
```lua
:FzfLua undotree
```
image

After pressing `F8`:
image

Same as code action can use native fzf previews + delta as pager
```lua
:FzfLua undotree previewer=undo_native
```
image

ibhagwan:
FYI, not sure yet what changed in nightly but all the CIs are failing due to an extra line in the layout with `fzf://` in it (haven't investigated yet but looks like the statusline).

phanen:
https://github.com/neovim/neovim/commit/8a626e5c4a5254c9ec952abe6ffe1e9b6fef9b2b
https://github.com/neovim/neovim/commit/31c814a0ed0eca1822ecb0409b06bab24d1c4792

ibhagwan:
Ty @phanen!

Which upstream changes to emmylua are we still waiting for? I did notice it could be a bit flakey sometimes but overall it’s usable.

phanen:
https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/824
flacky seems another issue

ibhagwan:
> [EmmyLuaLs/emmylua-analyzer-rust#824](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/824) flacky seems another issue

If you feel comfortable we can merge this

phanen:
ok, let me do a few fixup then.

phanen:
ok, no lsp diagnostics now. emmylua_check is still flaky.

phanen:
The sigwinch originally it attempt to fix the redraw issue on kitty, but don't work. So remove it.

Then I found simply `:|fzf --height 50%` on the bottom of screen then resize/shrink terminal also repro the same issue.

I found if bind this in shell (for fish: `bind ctrl-f 'nvim -l scripts/cli.lua files; commandline -f repaint '`),  it will magically work well, although idk why.

ibhagwan:
> I found if bind this in shell (for fish: bind ctrl-f 'nvim -l scripts/cli.lua files; commandline -f repaint '), it will magically work well, although idk why.

TIL about `commandline -f repaint` (I use zsh, not fish)

> ok, no lsp diagnostics now. emmylua_check is still flaky.

You feel ok with merging this or need more changes? We can keep the CI using `make lint` for now, no?

phanen:
yes, and no more changes.

PR #2325, opened 2025-09-10
fix(previewer): The preview window does not highlight during fuzzy se…

Joreh-T:
When using the fuzzy search function, the preview window does not highlight the searched keywords. This issue is likely caused by `opts.search` being set to empty first during fuzzy search.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1207,7 +1207,7 @@ end
 function Previewer.buffer_or_file:set_cursor_hl(entry)
   local mgrep, glob_args = require("fzf-lua.providers.grep"), nil
   local regex = self.opts.__ACT_TO == mgrep.grep and self.opts._last_query
-      or self.opts.__ACT_TO == mgrep.live_grep and self.opts.search or nil
+      or self.opts.__ACT_TO == mgrep.live_grep and self.opts._last_query or nil
   if regex and self.opts.fn_transform_cmd then
     local _, query = self.opts.fn_transform_cmd(regex, self.opts.cmd, self.opts)
     regex = query or regex


ibhagwan:
This does the reverse for me, before the PR the the highlight is accurate after a `ctrl-g` after the PR it doesn't work.

Can you share the reproduction steps to see the issue?

Joreh-T:
> This does the reverse for me, before the PR the the highlight is accurate after a `ctrl-g` after the PR it doesn't work.
> 
> Can you share the reproduction steps to see the issue?

I think I know why. If you don't input any characters when using 'FzfLua live_grep' (or something else regexp search), directly use <C-g> to enter fuzzy search mode, and then input characters to search, you can reproduce my problem.

Joreh-T:
> > This does the reverse for me, before the PR the the highlight is accurate after a `ctrl-g` after the PR it doesn't work.
> > Can you share the reproduction steps to see the issue?
> 
> I think I know why. If you don't input any characters when using 'FzfLua live_grep' (or something else regexp search), directly use to enter fuzzy search mode, and then input characters to search, you can reproduce my problem.

I found this issue because I directly used 'grep_project' for searching, but the keyword was not highlighted in the search results preview window.

ibhagwan:
> I found this issue because I directly used 'grep_project' for searching, but the keyword was not highlighted in the search results preview window.

I still can't reproduce this, I start with `grep_project` enter a term, then press `ctrl-g` and I see my regex highlighted properly.

When you start with fuzzy/grep_project (also fuzzy) the regex isn't highlighted as there's no regex, you're in fuzzy mode and fzf should take care of the highlighting in the main window (reverse by default as treesitter is enabled) but the preview window doesn't have highlighting as it's not a regex.

ibhagwan:
Can you post a screenshot of when you believe your regex should be highlighted and isn't so I can verify we're talking about the same thing?

Joreh-T:
> > I found this issue because I directly used 'grep_project' for searching, but the keyword was not highlighted in the search results preview window.
> 
> I still can't reproduce this, I start with `grep_project` enter a term, then press `ctrl-g` and I see my regex highlighted properly.
> 
> When you start with fuzzy/grep_project (also fuzzy) the regex isn't highlighted as there's no regex, you're in fuzzy mode and fzf should take care of the highlighting in the main window (reverse by default as treesitter is enabled) but the preview window doesn't have highlighting as it's not a regex.

According to your current design logic, the existing behavior is indeed normal. I originally assumed that even with fuzzy search, the keywords would be highlighted in the preview window. I apologize for not fully understanding the correct usage.

ibhagwan:
The reasoning behind he logic is that we can’t match non regex without implementing our own algo to highlight, when your fuzzy query is a whole word/regex this might work but imagine you type `ll` you’d need to highlight the first and last letters of `local`, you see where I’m going with this.

Joreh-T:
> The reasoning behind he logic is that we can’t match non regex without implementing our own algo to highlight, when your fuzzy query is a whole word/regex this might work but imagine you type `ll` you’d need to highlight the first and last letters of `local`, you see where I’m going with this.

Yes, I got it, thank you.

PR #2331, opened 2025-09-11
fix: contents/cmd are ambiguous

phanen:
- **chore: table.move don't exist in puc-lua**
- **chore**
- **chore**
- **chore**


ibhagwan:
Is this ready to merge @phanen?

phanen:
y
opts.cmd seems still needed in preprocess

ibhagwan:
> y opts.cmd seems still needed in preprocess

Got it, ty @phanen!

Issue #971, opened 2023-12-28
【Question or Suggestion】 Regarding Whitespace Stripping


RsrnlWysxr:
### Description 

This is a question or suggestion (in case the feature isn't available yet), rather than a troubleshooting issue.

fzf-lua is an amazing plugin, but I have a minor functionality question. Since I work with indentation-based programming languages like Python, there are many whitespace characters at the beginning of each line. How can I strip off these space characters? I couldn't find any configuration options for this.

![image](https://github.com/ibhagwan/fzf-lua/assets/41769433/8369a81a-1e13-42b4-9294-70be14ef7d71)

ibhagwan:
Do you have `rg` installed?

If so you can remove these by adding `--trim` to the command (or `rg_opts`), from `man rg`:
```
   --trim
       When set, all ASCII whitespace at the beginning of each line
       printed will be trimmed.
```

Try it out with:
```lua
:lua require("fzf-lua").live_grep({ cmd = "rg --line-number --column --color=always --trim" })
```

If you wish to set this up permanently with the default options, call `setup` with:
```lua
require("fzf-lua").setup({
  grep = {
    rg_opts = "--column --line-number --no-heading --color=always --smart-case --max-columns=4096 --trim -e",
  },
  -- your other options
})
```

RsrnlWysxr:
This is fantastic, it works perfectly, thank you so much! @ibhagwan

ibhagwan:
> This is fantastic, it works perfectly, thank you so much! @ibhagwan

You’re welcome @RsrnlWysxr!

tssm:
Hey @ibhagwan, is it possible to make this work for live grep? It seems to ignore the `rg_opts`

ibhagwan:
> Hey @ibhagwan, is it possible to make this work for live grep? It seems to ignore the `rg_opts`

Both grep and live grep use the same `rg_opts`, if it’s not working for you the issue is somewhere else.

PR #2337, opened 2025-09-14
refactor: rg_opts add default --trim

tronikelis:
## Changes

Adding `--trim` option by default to the grep picker for `ripgrep`

## Why

This is a personal style change which I think would be a sane default, there was already an issue opened regarding this https://github.com/ibhagwan/fzf-lua/issues/971

Also, I don't see any negatives with having this option, the readability only improves as you can actually see what you matched in the fzf list, otherwise you can opt out with `--no-trim`

<details>
<summary>Before</summary>

image

</details>

<details>
<summary>After</summary>

image

</details>

But as this is a personal "feelings" change feel free to close the PR

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -605,7 +605,7 @@ M.defaults.grep                  = {
       .. "--extended-regexp -e"
       or "--binary-files=without-match --line-number --recursive --color=always "
       .. "--perl-regexp -e",
-  rg_opts        = "--column --line-number --no-heading --color=always --smart-case "
+  rg_opts        = "--trim --column --line-number --no-heading --color=always --smart-case "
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty for the PR @tronikelis.

As you wrote this is a personal preference and while you may not see any downsides I’m sure there will be others thinking differently about losing the indentation, this will be especially apparently with `grep_curbuf` and the treesitter highlighting and I’m sure there will be other reasons.

Given this has worked forever without trim I don’t feel this should be changed at this point.

tronikelis:
Fair enough, thanks for such a quick reply, as a recent migrator from telescope, love the plugin so much

ibhagwan:
> Fair enough, thanks for such a quick reply, as a recent migrator from telescope, love the plugin so much

Ty for your kind words

PR #2339, opened 2025-09-15
Add `~` to the list of very magic special characters

joshbode:
If you search for any pattern containing `~` then you get this error when using `live_grep`:
```
[Fzf-lua] Unable to init vim.regex with "\v~", couldn't parse regex: Vim:E867: (NFA) Unknown operator '\~. . Add 'silent=true' to hide this message.
```

This change adds `~` to the list of characters that need to be escaped when converting the regex back to a very magic vim regex.

Similar PR: https://github.com/ibhagwan/fzf-lua/pull/2270


ibhagwan:
Ty @joshbode!

PR #2348, opened 2025-09-17
Revert "fix: don't fallback to st for function contents"

Parsifa1:
This reverts commit 82f66229d4cc01c5ed2bf8b4f727f45fcae05c2d.


Parsifa1:
some test broken, may not a good idea.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2349

PR #2355, opened 2025-09-23
fix(marks): Use correct condition for marks filter

yilisharcs:
Commit 4cb7d7c614a55d55dcc175c2457b7be0cbbdecec inverted the behavior of `marks.marks` filtering.


ibhagwan:
Ty @yilisharcs

Issue #2353, opened 2025-09-21
Feature: special highlight for current mark in marks picker


yilisharcs:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

If the user is in a file that has a global mark, `:FzfLua marks` will show the text in that file's marked line. But it's all white and hard to immediately pick out on a list. Can't it be a different color instead, like with native `:marks`?

And how could I go about implementing it?

ibhagwan:
Marks uses `FzfLuaCursorLine` and `FzfLuaCursor` to mark the line of the mark and the column of the mark (the only data we have about the mark).

You can also set these just for marks:
```lua
:FzfLua marks profile=default hls.cursorline=IncSearch hls.cursor=PmenuSel
```

yilisharcs:
I wasn't sufficiently clear. I meant in the list itself, in the "file/text" column.

<img width="1002" height="682" alt="Image" src="https://github.com/user-attachments/assets/7ea67ac9-64e8-4aa9-ba40-ffc6a9161862" />

ibhagwan:
Reusing existing highlights:
```lua
:FzfLua marks hls.buf_nr=DiffText hls.path_linenr=IncSearch hls.path_colnr=CurSearch
```

<img width="1210" height="605" alt="Image" src="https://github.com/user-attachments/assets/38156e81-a346-4ee9-be26-e749a36949ac" />

The filename / text colors is fzf's fg color, if you wish to control that via `fzf_colors`:
```lua
:FzfLua marks hls.buf_nr=DiffText hls.path_linenr=IncSearch hls.path_colnr=CurSearch fzf_colors.fg={"fg","Directory"}
```

<img width="1215" height="610" alt="Image" src="https://github.com/user-attachments/assets/1e8fd285-514a-412d-8a8f-cc618672f55f" />

> Can't it be a different color instead, like with native `:marks`?

Technically we can add a highlight to differentiate between a file and text similar to `:marks` but I didn't implement that.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/4cb7d7c614a55d55dcc175c2457b7be0cbbdecec

Refactored marks to use `vim.fn.getmarkslist` api instead of parsing the output of `:marks`, local marks now appear at the top of the list with the same highlight as `:marks` (which uses the `Directory` highlight), to sort the marks as before use `:FzfLua marks sort=true`.

Example output of:
```lua
[...]

PR #2356, opened 2025-09-23
feat: Replace local buf global mark output with highlighted text

yilisharcs:
Commit 4cb7d7c614a55d55dcc175c2457b7be0cbbdecec didn't fully address the github issue

Addresses #2353

It looks like this now:
Screenshot From 2025-09-23 20-10-29


ibhagwan:
Ty @yilisharcs!

fixed the condition:
```lua
if bufnr == utils.CTX().bufnr then
```

PR #2359, opened 2025-09-25
feat(marks): Implement split actions for marks picker

yilisharcs:
`files` has them, `buffers` has them, `helptags` has them... and now `marks` has them too.


ibhagwan:
Ty @yilisharcs !

PR #2365, opened 2025-09-28
ci: keep old gutter shape (for current screenshots)

phanen:
default changed after https://github.com/junegunn/fzf/pull/4521


ibhagwan:
We can also update fzf to latest in the CI and update the screenshots instead maybe more future proof?

but that would depend in the package managers having the latest fzf in the CI

phanen:
Yeah, that also make a huge diff.

phanen:
I forgot fzf_cli_args won't check changlog table.

phanen:
disable all ci for now, since 0.66 is never released we cannot check 0.66 in changelog.

junegunn:
If you wish to hide the gutter, follow the direction here:

https://github.com/junegunn/fzf/blob/master/CHANGELOG.md#hiding-the-gutter-column

ibhagwan:
https://github.com/junegunn/fzf/releases/tag/v0.66.0

phanen:
Not sure why, even changing `--color gutter:xx` won't generate different screenshots. (maybe there's a bug?)

I find a workaround to use `gutter:reverse:bright-black`, it will make 0.65.2 generate the same screenshots as 0.66.

ibhagwan:
@phanen, can I merge it like this or are you still investigating the gutter color issue?

phanen:
ok to merge
i debugprint the args and run in shell and gutter color indeed changed, so maybe some other attr changed

ibhagwan:
> ok to merge i debugprint the args and run in shell and gutter color indeed changed, so maybe some other attr changed

We can also update the screenshots if you think it’s more future proof, lmk

phanen:
i intentionally make test work both in 0.66/0.65 now. since window ci use scoop which may have not been updated
we can keep compat, many user can still use <0.66 for long time.

ibhagwan:
Tysm @phanen, merged.

ibhagwan:
Something is up with the Windows CI, seems unrelated to the gutter change though

PR #2366, opened 2025-10-02
fix: memoize has_ts_parser call

phanen:
fix: duplicated entry


ibhagwan:
Ty @phanen this is great

ibhagwan:
FYI, this creates an issue with some parsers due to the "context" call, for example try to preview `Makefile`:
image

ibhagwan:
For the tiem being https://github.com/ibhagwan/fzf-lua/commit/23fca9ac345b8b7da326f16d0d1db239f293b43c

PR #2369, opened 2025-10-04
fix(commands): cmd_global -> [opts.hls.cmd_global]

phanen:
Fail to call `:FzfLua commands` with buffer-local commands.


ibhagwan:
Whoops, ty @phanen !

PR #2370, opened 2025-10-04
fix(preview): swallow warning for missing ts parsers

kostiskalkounos:
This [commit](https://github.com/ibhagwan/fzf-lua/commit/464bab1fe59c4b92f3a19eda86b1d8536f40d8ad#diff-6610b6cdb84f62055c6687508d2346e099b304fe0095ff33b8fed55474cd77aaR1473) introduced a regression in nvim 0.11+. If a treesitter parser isn't installed, it shows a warning. The functionality before this commit, was to swallow the warning and highlight the buffer with whatever colors it could, if any at all.

This PR aims to restore the original functionality, without affecting older neovim versions.

An example of a warning is shown below:

image


ibhagwan:
Ty @kostiskalkounos !

PR #2371, opened 2025-10-05
Add action that avoids splitting when opening help

tssm:
I've been using this patch for a while, and as suggested [here](https://github.com/ibhagwan/fzf-lua/discussions/1268) created a new action, so the default behaviour is not changed


ibhagwan:
Renamed to `actions.help_curwin` and added help to it, ty @tssm!

PR #2383, opened 2025-10-13
fix(actions): ignore dirty buffers if not last window

gravndal:
Fixes issue where fzf-lua would refuse to switch away from dirty buffers opened in multiple windows when `set nohidden`.


ibhagwan:
Ty @gravndal

PR #2388, opened 2025-10-15
feat(git-worktrees): add & delete

barrettruth:
Support add and delete with the git worktrees picker.

# impl details

- adds *default* bindings to `<c-{a,x}>` for add,delete
- delete asks for confirmation
- worktree paths default to `../<branch>` (standard)
- if a branch exists, switch to it - otherwise, create with `-b` flag

All of the above can definitely be subject to change! Let me know.


ibhagwan:
Tysm @barrett-ruth, added a fallback to git switch -> cd when aworktree is found (otherwise git switch fails)

https://github.com/ibhagwan/fzf-lua/pull/2388/commits/04097f5f6db64815b22abf7faf56a17978650aac

ibhagwan:
Btw, TIL that deleting a worktree isn't that horrible, the branch gets "merged" back into the current cwd :-)

barrettruth:
🐐

PR #2389, opened 2025-10-17
git branch: change default picker list

fin-w:
Hi!

Not sure if you'd be interested to change the default git branch picker list display format? I have changed my personal default and I prefer it. It's not enormously different from the current default but it might make the out-of-box behaviour a bit better?

This MR contains my current config, if you're interested I can tidy it up / change it for merging.

As it stands now, this would add a dependency on `column`, not sure if git can arrange the columns as nicely like `column` does, I think I could at least set unchanging column widths.

The picker list is organised by current branch first, then local branches ordered by most recently changed to least recently changed, then remote branches listed in the same order. And the relative date is also added in the list. I've added the commit hash too because without it fzf-lua's current parsing of the selected picker option doesn't work, so the preview for the git branch doesn't work. If this MR is accepted I can probably change fzf-lua's parsing of the list so the git commit hash isn't required.

**Before:**

**After:**

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -536,7 +536,8 @@ M.defaults.git                   = {
     _treesitter   = function(line) return line:match("(%s+)(%d+)%)(.+)$") end,
   },
   branches = {
-    cmd        = "git branch --all --color",
+    cmd        = [[git branch --all --color -vv ]]
+        .. [[--sort=-'committerdate' --sort='refname:rstrip=-2' --sort=-'HEAD']],
     preview    = "git log --graph --pretty=oneline --abbrev-commit --color {1}",
     remotes    = "local",
     actions    = {
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -184,12 +184,14 @@ M.branches = function(opts)
   if opts.preview then
     local preview = path.git_cwd(opts.preview, opts)
     opts.preview = shell.stringify_cmd(function(items)
-      -- all possible options:
+      -- The beginning of the selected line looks like the below,
+      -- but we only want the string containing the branch name,
+      -- so match the first sequence not including spaces:
       --   branch
       -- * branch
       --   remotes/origin/branch
       --   (HEAD detached at origin/branch)
-      local branch = items[1]:match("[^%s%*]*$"):gsub("%)$", "")
+      local branch = items[1]:match("^[%*+]*[%s]*[(]?([^%s)]+)")
       return (preview:gsub("{.*}", branch))
     end, opts, "{}")
   end


ibhagwan:
This looks great @fin-w.

The only concern is will this work on Windows given the `| column … ` part?

fin-w:
I don't think `column` works on Windows. At least when I checked briefly on W11 it isn't a command. I can see if there's a way to format without using `column`.

1. Do you want the asterisk (HEAD) coloured green or not? Current default is not, this MR currently is coloured.
2. Is it OK that the `remotes/` prefix is dropped from the entries in this MR?
3. Is it a problem that the ` -> origin/main` syntax is removed in this MR? I guess it's to signify that the remote branch has a local counterpart, I'd have to look into how that is created if you want to keep it.
4. Because I only added the commit hash to fix the branch preview not working with this change, I'd rather remove it and change the picker menu parsing logic instead. Is this easy / straightforward? I haven't looked closely at the code but I'm hoping I can just get the first field in the selected menu entry (the branch name) and get the preview from that data?

ibhagwan:
> I don't think column works on Windows. At least when I checked briefly on W11 it isn't a command. I can see if there's a way to format without using column.

This is important, we don’t have to have inconsistencies between windows/linux/mac.

> Do you want the asterisk (HEAD) coloured green or not? Current default is not, this MR currently is coloured.

I think it’s good to have a clear indication of the current branch

> Is it OK that the remotes/ prefix is dropped from the entries in this MR?

As long as it’s still clear enough this is a non local branch (via origin/ or the red color?)

> Is it a problem that the  -> origin/main syntax is removed in this MR? I guess it's to signify that the remote branch has a local counterpart, I'd have to look into how that is created if you want to keep it.

I think it’s fine, if you find a creative way to add it would be a nice “bonus” but IMHO it’s not critical 

> Because I only added the commit hash to fix the branch preview not working with this change, I'd rather remove it and change the picker menu parsing logic instead. Is this easy / straightforward?

Logic is fairly straightforward, you can change it here:
https://github.com/ibhagwan/fzf-lua/blob/c102a4eeb10a04ed7e69ce80918923c4175c19d8/lua/fzf-lua/providers/git.lua#L186-L194

Another important change I recently added was the parsing of work tree branches, since git switch doesn’t work on a work tree branch it falls back to a change directory, we have to make sure this is working.

Probably also have to add a new indication for worktree branches (prefixed with +):
IMG_3754

Said code:
https://github.com/ibhagwan/fzf-lua/blob/c102a4eeb10a04ed7e69ce80918923c4175c19d8/lua/fzf-lua/actions.lua#L844-L849

fin-w:
Latest state doesn't use `column`, preferring Git's built-in column formatting that isn't as good for long column entries.

I've removed the colour on the asterisk, in line with the current default. The asterisk identifies where HEAD is pointed at, what do you think the "+" is / should identify? Do you want to see a "+" marker in front of every work tree entry, or only when the work tree is HEAD?

The name of each branch comes from `%(refname:short)`, which Git tells me will always be unique, so there shouldn't be any issues identifying them.

ibhagwan:
> Do you want to see a "+" marker in front of every work tree entry, or only when the work tree is HEAD?

It seems that all work trees have + preceding the entry, try it out yourself with `:FzfLua git_worktrees` and the ctrl-a action (use both names of branches and random names).

fin-w:
Edit: Sorry, I see what you mean now. The default `git branch --all` shows worktrees with the `+` prefix, `git_branches()` shows worktrees.

That's not how it looks for me. The command fzf-lua is using to generate the picker menu entries is `git worktree list` and within fzf-lua and also in the terminal running the same command, I see an absolute path, a short commit hash, and the worktree name in square brackets.

fin-w:
The `+` prefix means "this BRANCH is checked out in another worktree". I can enter the worktree directory and checkout whatever branch I want, and when I go back to the original worktree directory, the branch I just checked out has the `+` prefix instead.

So I guess a branch is created for each new worktree, which has the same name as the worktree directory.

For `git_branches()` it feels to me like worktree shouldn't be very relevant? We are looking at a list of branches, and while it may be that some branches are checked out in a worktree, switching to the branch from `git_branches()` switches within the current worktree rather than changing worktrees.

I can try to create logic for the `+` prefix but I'm not convinced it's possible without scripting of some sort. I suspect a better approach would be to gather all the Git data separately to the pickers, and manage it all in Lua, displaying it after processing in it Lua, meaning that we can make the view and model separate and not fight with Git's horrible formatting and frustrating behaviour. But that sounds like a lot of work and changes how fzf-lua uses Git in quite a fundamental way.

A very simple, alternative change that still provides a quality-of-life improvement is 
```lua
            cmd = [[git branch --all --color --sort=-'committerdate' --sort='refname:rstrip=-2']],
```
which changes the sort order, but doesn't add the relative date column. This does away with the custom formatting, retains the `*` and `+` prefixes, and keeps the colour highlights the same as they are currently.

Shall I reduce this MR to only change the sort order, or are you willing to lose the worktree marker and colouration?

ibhagwan:
> For git_branches() it feels to me like worktree shouldn't be very relevant? We are looking at a list of branches, and while it may be that some branches are checked out in a worktree, switching to the branch from git_branches() switches within the current worktree rather than changing worktrees.

Your logic makes sense but `git branches` does display it and users which currently use this functionality will have an issue with that.

> Shall I reduce this MR to only change the sort order, or are you willing to lose the worktree marker and colouration?

What I fear is that this becomes overly complex and with reduced functionality the next issue of “why is git branches missing parity with `git branches`” command we’ll have to either revert this commit or start adding hacks - perhaps this PR is better server as you previously used it (i.e. not as default?)?

ibhagwan:
Btw, the worktree branch info is not just “extra data”, if you try to checkout/switch to a branch that’s checked out in another worktree `git switch` will fail with an error `fatal: 'branch' is already used by worktree at '…'`.

Therefore I argue this information is vital for the branches picker.

ibhagwan:
`git branch -vv` is another option:
IMG_3755

fin-w:
I prefer the latter, but it probably only works for the vertical arrangement where the picker menu is a full screen width. I prefer both of these over the current default.

```lua
cmd = [[git branch --all --color --sort=-'committerdate' --sort='refname:rstrip=-2' --sort=-'HEAD']],
```



```lua
cmd = [[git branch --all --color -vv ]]
    .. [[--sort=-'committerdate' --sort='refname:rstrip=-2' --sort=-'HEAD']],
```
(This would require fixing the preview)

ibhagwan:
> I prefer the latter, but it probably only works for the vertical arrangement where the picker menu is a full screen width

I prefer the latter too

> This would require fixing the preview

No big deal, probably requires also fixing the git_switch action.

fin-w:
Edit: figured it out. It's the whole menu entry.

What is `items` in this context? Is it only ever a string as defined by the comments, or is it the whole picker menu entry?
https://github.com/ibhagwan/fzf-lua/blob/c102a4eeb10a04ed7e69ce80918923c4175c19d8/lua/fzf-lua/providers/git.lua#L186-L194

It looks to me like `branches.preview` is a string, and `opts.preview` here is a function that returns the `branches.preview` string with the "{1}" replaced by the git branch name, which is extracted from `items`?..

ibhagwan:
> What is items in this context? Is it only ever a string as defined by the comments, or is it the whole picker menu entry?

Items is an array containing the fzf field index, in this case it’s the selecte(d) items {}, since multi select isn’t used here it’s basically just items[1] which is the selected line text.

fin-w:
OK, I think this is ready for review? Latest changes include more robust parsing of the selected fzf entry so the preview is reliable.

I've learned that git branches can be named almost anything including characters like `@+-_}{` etc but my regex handles that, even when the `+` or `*` marker precedes the horrible branch name, while also stripping the preceding open bracket for e.g. `(HEAD detached...`. Below is a screenshot where you can see the debug printing (bottom RHS) showing that the regex is parsing the branch name correctly.

ibhagwan:
Ty @fin-w! squash-rebase-merged.

PR #2396, opened 2025-10-21
feat: add use_absolute_paths to files picker

obergodmar:
Hi!

I think this improvement could be quite useful for some users. The file picker is extended with a new property, `use_absolute_paths`, which simply disables the `$HOME` replacement with `~` when displaying paths.

I’ve been using this feature for a year already, and it has helped me a lot in scenarios where I have an absolute file path (e.g., generated by a script, linter, or any other external source) and want to open it using the file picker. By default, since all paths start with `~`, the results window remains empty even if the file actually exists.

Here’s the current behavior:

1. I have an absolute file path.
2. I try to find this file using the file picker.
3. The result is empty if I search directly.

image

because all the paths start with `~`:

image

The main issue is that I can’t locate a file using the path I’ve copied from a linter output in another terminal. I either have to convert the path to a relative one or configure the linter (or another tool) to generate relative paths. This can also happen when someone shares an absolute path, and you want to open that file directly.

Here’s how it works when `use_absolute_paths` is passed to the file picker with a value of `true`:

```lua
require('fzf-lua').files({
        use_absolute_paths = true,
        cmd = get_fd_cmd(), -- fd should be called with --absolute-path . to display full paths
})
```

image

Let me know your thoughts. If any other picker could benefit from the same behavior, I’d be happy to update them as well. Personally, I only use the file picker, as this is the only scenario where the issue occurs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,6 +833,7 @@ previewers = {
     hidden             = true,          -- enable hidden files by default
     follow             = false,         -- do not follow symlinks by default
     no_ignore          = false,         -- respect ".gitignore"  by default
+    absolute_path      = false,         -- display absolute paths
     actions = {
       -- inherits from 'actions.files', here we can override
       -- or set bind to 'false' to disable a default action
--- a/lua/fzf-lua/make_entry.lua
+++ b/lua/fzf-lua/make_entry.lua
@@ -580,20 +580,27 @@ M.file = function(x, opts)
   if opts.render_crlf then
     filepath = path.render_crlf(filepath)
   end
-  -- make path relative
-  if opts.cwd and #opts.cwd > 0 then
-    filepath = path.relative_to(filepath, opts.cwd)
+  if opts.absolute_path then
+    -- make path absolute
+    if not path.is_absolute(filepath) then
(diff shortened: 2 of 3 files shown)


ibhagwan:
Ty @obergodmar, ATM this PR just disables the ~ replacement, it doesn’t make relative paths absolute paths.

I’m thinking this may be better servers as a formatter (like path.filename_first), I’ll think about it.

obergodmar:
> ATM this PR just disables the ~ replacement, it doesn’t make relative paths absolute paths.

Yes, that’s correct. As I mentioned in the description, the option, when used in combination with `fd --absolute-path`, displays absolute paths — not just the option itself.

Anyway, the option name can be changed, and its sole purpose is to disable the `$HOME` to `~` replacement and display paths as they are.

ibhagwan:
> Anyway, the option name can be changed, and its sole purpose is to disable the $HOME to ~ replacement and display paths as they are.

Renamed the option to `absolute_path` to match the `fd --absolute-path` flag and made it work regardless of using the flag or fd, the option now does what it says and converts all paths to absolute if need be, this works on all file-like pickers, LSP/quickfix/oldfiles/etc.

PR #2401, opened 2025-10-22
fix(git_branch_del): update regex to reflect new git branch entries

fin-w:
The current regex is wrong now that #2389 is merged. This updates the regex to be the same as that used in the function `branches` in `git.lua`, so that the correct part of the picker menu entry (the branch name) is matched by the regex.

Really these two regexes should be unified so there are not two locations to maintain...


ibhagwan:
> Really these two regexes should be unified so there are not two locations to maintain...

If you really want we can extract it to a function similar  to git hash, not that critical since only 2 locations currently

PR #2408, opened 2025-10-26
fix: snacks inline image broken when use terminal.env() before term detection

phanen:
https://github.com/folke/snacks.nvim/blob/43261baf87fdc5f970c12e9a6c795ba8a4e7595c/lua/snacks/image/terminal.lua#L111-L113
Maybe it cached invalid value here.
Fix by terminal.detect() before terminal.env()


ibhagwan:
Ty @phanen!

Issue #2054, opened 2025-05-19
Always resume to terminal mode when hide


phanen:
```
:term fd
:FzfLua files profile=hide
<esc>
```

Expected: keep in normal mode.
Actual: always resume to terminal mode.

ibhagwan:
Idk if I would consider this a bug, why should we resume in normal mode what purpose does it serve? Since there is no real “normal” mode in fzf-lua (at least not a useful one) the users expect to open fzf-lua and start typing right away.

phanen:
I mean we call fzf-lua from normal mode, but after quit fzf-lua we're in terminal mode.

The behavior is different when use `default` profile.

ibhagwan:
> I mean we call fzf-lua from normal mode, but after quit fzf-lua we're in terminal mode.
> 
> The behavior is different when use `default` profile.

I know but it’s an edge case 99% of the users don’t encounter, idk if we should really bother with it?

phanen:
It seems a neovim feature/bug. Usually mode will be kept when switch buffer. But it's a bit complex when deal with terminal buffer (and there're more backward-compat things to be considered since the behavior seems different on older version):
* terminal buffer -> non-terminal buffer: any mode -> normal mode
* non-terminal buffer -> terminal buffer: insert mode -> terminal mode
* terminal buffer -> terminal buffer: keep mode unchanged?
* terminal buffer killed: resume to normal mode

The only bad thing is that when if switch to insert mode in terminal buffer, your cursor will be put to the end of buffer.

Since we have to resume normal mode before switch terminal buffer (I'm not sure if it's feasible?)... But in that case when `unhide` we have to switch back if needed. So maybe close it. I don't want to dig into it more and find more bug...

I close it. But this comment maybe useful when some other guys met similar problem.

AndreiCalazans:
@ibhagwan thanks for this plugin, I'm loving how snappy it is. 

I ran into this as well, idk why but it worked correctly with Telescope which makes me question if it is a neovim bug. 

Here is a video:

Steps are:
1. open term `:term`
2. note how term open in normal mode
3. trigger any fzf-lua picker, here I trigger buffers
4. exit, goes back to term in insert mode
5. I test opening other things like :LspInfo, term returns in appropriate mode (normal).

https://github.com/user-attachments/assets/aed92d76-0ecb-4ab8-8f1b-6d02f10c8042

ibhagwan:
[...]

PR #2419, opened 2025-11-11
blines on a terminal buffer

phanen:
Cursor won't resume to correct position after `<CR>`/`<ESC>`...
When you run `blines` on a terminal buffer.

I don't think this approach is good... just let u know we have this kind issue.

**EDIT**:
snacks.picker also have this issue... so maybe open an issue ~upstream~ then wait folke to solve/close it.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -133,6 +133,9 @@ M.dummy_abort = function(_, o)
   if o.complete and o.__CTX.mode == "i" then
     vim.cmd [[noautocmd lua vim.api.nvim_feedkeys('i', 'n', true)]]
   end
+  if o.__CTX.mode == "nt" and o.__CTX.cursor then
+    vim.api.nvim_win_set_cursor(0, { o.__CTX.cursor[1], o.__CTX.cursor[2] })
+  end
 end
 
 M.resume = function(_, _)
@@ -261,10 +264,12 @@ M.vimcmd_entry = function(vimcmd, selected, opts, bufedit)
       elseif not opts.no_action_set_cursor and entry.line > 0 or entry.col > 0 then
         -- Make sure we have valid line/column
         -- e.g. qf lists from files (no line/col), dap_breakpoints
-        pcall(vim.api.nvim_win_set_cursor, 0, {
-          math.max(1, entry.line),
-          math.max(1, entry.col) - 1
-        })
+        vim.defer_fn(function()
+          pcall(vim.api.nvim_win_set_cursor, 0, {
+            math.max(1, entry.line),
+            math.max(1, entry.col) - 1
+          })
+        end, 100)
       end
       -- Only "zz" after the last entry is loaded into the origin buffer
       if i == #selected and not opts.no_action_zz and not utils.is_term_buffer(0) then


ibhagwan:
I think the entire mode restoration code should probably be refactored and put somewhere like `action_ post` or `win::post` (gets called on both close and hide).

https://github.com/ibhagwan/fzf-lua/commit/d7497da366b653cd140082bcd8184d7dfa877a23 https://github.com/ibhagwan/fzf-lua/commit/b19dabe6ddc0b1cef4e0b30e864aca88fdabc551

Not 100% sure if we wanna centralize/normalize cursor setting as well.

ibhagwan:
@phanen, apparently this only happens when using `hide` profile, I added  ` print("get cursor", vim.api.nvim_win_get_cursor(0)[1])` at the end of the hide action hijack (`hide.lua:L91`) and it shows the correct line,  AFAIK this is the last line called in the `conver_exect_silent` sequence... consused what is restoring the line to the end of the term buffer lol

ibhagwan:
Interesting, this diff "soles" the buffer lines in term buffer issue, have to AFK, hopefully I have time to solve this later:
```diff
diff --git a/lua/fzf-lua/win.lua b/lua/fzf-lua/win.lua
index 8e49954e..1ce54cd1 100644
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -1424,7 +1424,7 @@ function FzfWin:hide()
   -- do not apply to tmux, validate anyways in case called directly using the API
   if not self or self._o._is_fzf_tmux then return end
   self:detach_fzf_buf()
-  self:close(nil, true)
+  self:close(nil, false)
   self:save_size()
   -- Save self as `:close()` nullifies it
   _self = self
```

ibhagwan:
@phanen, do you have any creative idea what's going on here?

Using `:FzfLua blines profile=default` on a term buffer and pressing `ctrl-c` doesn't move the cursor but pressing `esc` (hide) apparently puts the term buffer in INSERT mode (which is late restored to NORMAL) by `FzfWin:close()`. but **it's the INSERT mode which sends the cursor to the bottom of the buffer**.

The interesting thing is that I don't know what's causing it, it seems that this is somehow related to the existence of the hidden fzf buffer, I even tried killing the fzf process inside `_hidden_fzf_bufnr` but the behavior still persists, the only way to prevent the source terminal from going back to INSERT mode is to kill `_hidden_fzf_bufnr`.

ibhagwan:
So this seems to be a neovim issue but I'm not sure how to solve it (yet), if you have a terminal window in "terminal mode" (insert mode) and you switch to another terminal in normal mode (using `vim.cmd.hide()`, closing current window, etc) the terminal window you're switching into will end up in terminal mode as well (even if it was in normal mode).

One workaround is to first set the exiting terminal window (i.e. in our case the fzf main win) to normal mode before exiting the terminal (via `<ctrl-\>+<ctrl-n>`  after which you can return to the origin terminal window safely.

The issue I'm currently facing is that I can't put the fzf window into terminal mode on time (before switching) as`feedkeys` is most likely queued and I haven't figured out how to programmatically set the window to normal mode on time before the hide (unsure if that's even possible).

ibhagwan:
@phanen, what do you think of https://github.com/ibhagwan/fzf-lua/commit/2ca3dce4d40e9e515102d89ebc5500f78bb94d9a?

It seems that calling `:stopinsert` before closing/hiding solved both #2054 and this issue, instead of restoring normal mode after closure we don’t get into the issue to begin with.

ibhagwan:
Tested on Linux/Mac, works great for both linked issues.

Closed via https://github.com/ibhagwan/fzf-lua/commit/3b53b0eb26972686c5825d7c0e63b3cffd0c4f2b

Issue #2420, opened 2025-11-11
Feature: Nest Subcommands In `FzfLua commands` 


barrett-ruth:
### Have you RTFM'd?

- [x] I have done proper research

### Feature Request

Hi ibhagwan, back with another idea.

Long story short I'm working on my workflow rn and looking for a more versatile and uniform picker environment.

Namely, it would be nice if `FzfLua commands` automatically leveraged completion to allow for further nested picking.

Namely, imagine the following:

1. `:FzfLua commands`
2. click `TSEnable`
3. Rather than `:TSEnable ` populated in the cmd, see the following completion items in *another* picker:

<img width="544" height="383" alt="Image" src="https://github.com/user-attachments/assets/6ce4fb2d-cd92-450f-b01d-fed8d2317530" />

barrettruth:
NGL, I wrote a draft of this - I think this is pretty cool.

The following must be addressed imo in terms of design:

1. keybindings/default behavior for:
- nesting/going one level deeper
- quitting & populating the cmdline in vim
- straight up running the cmd
2. aesthetics
- where in the picker to display the already-typed command

<img width="1017" height="419" alt="Image" src="https://github.com/user-attachments/assets/698976db-4071-494d-aa3e-c732452e81fc" />

barrettruth:
UPD: [pr](https://github.com/ibhagwan/fzf-lua/pull/2421) for you to look at/work off of if you like (i've noticed you prefer to just do things yourself a lot, can't blame you)

nenahp:
it has a opts.flatten not sure if it's similar

ibhagwan:
Unless we need something a bit more sophisticated here isn’t this best case usage for a simple action which sends all subcommand to `vim.ui.select`?

barrettruth:
@ibhagwan 

I've taken a second to think about it and my idea fits most neatly into extending `FzfLua commands` from a naive, basic picker to a command-line-like picker similar to "noice.nvim". This is indeed "a bit more sophisicated" but I think this picker would a) vastly improving latency from thought -> execution b) be easy to maintain by integrating tightly with neovim's native completion and c) in its primitive state, not really be all that complicated to implement. 

Allow me to introduce three examples i find compelling:

1. My plugin [cp.nvim](https://github.com/barrett-ruth/cp.nvim) which support a decently sophisticated command structure under `:CP`:

- `:CP pick`
[...]

PR #2421, opened 2025-11-11
(garbage draft) add nested picker with

barrettruth:
***This pr is utter garbage as it stands but works as described below. I figured you would have better intuition about how to integrate this feature into a picker (obviously) so put this pr up if you'd like to take it as a starting point***

Adds ability to nest command pickers when a command has completions. Press ctrl-s on a command to open a nested picker with its subcommands.

Missing features:
- explicit keybinds to populate the command/run the cmd
- docs


ibhagwan:
Given you closed #2420 do we still need this PR?

PR #2422, opened 2025-11-13
fix: handle paths containing spaces in `arg_exec`

illia-shkroba:
Handle paths containing spaces in the actions using `arg_exec`:

* `fzf.actions.arg_add`
* `fzf.actions.arg_del`

Before:
image
image

After:
image
image

I've handled it as described in the `:help :argedit`:

> Spaces in filenames have to be escaped with "\". [count] is used like with |:argadd|.


ibhagwan:
Ty @illia-shkroba !

Issue #2423, opened 2025-11-14
Bug: vim.ui.select may throw error after lsp_code_actions 


phanen:
Run lsp_code_actions on cursor position without code actions.

Then vim.ui.select is never called in code -> so never deregister
```
Then fzf-lua stop working on new picker:
[Fzf-lua] fn_selected threw an error: /home/phan/b/fzf-lua/lua/fzf-lua/previewer/codeaction.lua:289: assertion failed!
```

phanen:
I'm not sure why I never get into this case before, maybe coroutine.resume hide it before and now we already use coxpcall or maybe code_actions are my only customer to vim.ui.select before

https://github.com/ibhagwan/fzf-lua/blob/634eba3c266f0fa73564d91cdcfd85b321c8abd2/lua/fzf-lua/providers/lsp.lua#L912-L921

There seems code to avoid this case. And it seems true there's no other way to know if there's no code actions.

https://github.com/neovim/neovim/blob/4dd9137215f88bb6eabe386a7e474c59783a6486/runtime/lua/vim/lsp/buf.lua#L1315-L1318

phanen:
Ok upstream also don't provide any callback other than `vim.ui.select`
https://github.com/neovim/neovim/blob/4dd9137215f88bb6eabe386a7e474c59783a6486/runtime/lua/vim/lsp/buf.lua#L1138-L1142
```
--- Can't call/use vim.lsp.handlers['textDocument/codeAction'] because it expects
--- `(err, CodeAction[] | Command[], ctx)`, but we want to aggregate the results
--- from multiple clients to have 1 single UI prompt for the user, yet we still
--- need to be able to link a `CodeAction|Command` to the right client for
```

maybe I can avoid use picker for this, at a cost maybe I lose the diff preview, not a big deal

phanen:
Interestingly if I register fzf-lua vim.ui.select then no error will happen...

~I also have a idae maybe I can wrap all my picker function with a deregister before...!~ But vim.ui.select still can not be used since it's registered

ibhagwan:
For some reason I cannot reproduce this, even when I first call `:FzfLua deregister_ui_select`.

> There seems code to avoid this case. And it seems true there's no other way to know if there's no code actions.

This is why we had  be669fa but it ended up causing more issues so I removed it and conformed to all other pickers requiring ui.select.

phanen:
maybe related to my config. i ll try to repro it next time

[...]

PR #2424, opened 2025-11-14
fix(ui_select): pending deregister when vim.ui.select is not called in codeactions

phanen:
https://github.com/ibhagwan/fzf-lua/issues/2423#issuecomment-3532993681

### DIFF
--- a/lua/fzf-lua/providers/ui_select.lua
+++ b/lua/fzf-lua/providers/ui_select.lua
@@ -156,6 +156,16 @@ M.ui_select = function(items, ui_opts, on_choice)
   -- ui.select is code actions
   -- inherit from defaults if not triggered by lsp_code_actions
   local opts_merge_strategy = "keep"
+
+  -- fix error when vim.lsp.buf.code_action() called but didn't triggers vim.ui.select
+  -- _OPTS_ONCE also means pending deregister
+  -- since we only use it to custom codeaction preview now
+  if _OPTS_ONCE and ui_opts.kind ~= "codeaction" then
+    M.deregister({}, true, true)
+    _OPTS_ONCE = nil
+    return vim.ui.select(items, ui_opts, on_choice)
+  end
+
   if not _OPTS_ONCE and ui_opts.kind == "codeaction" then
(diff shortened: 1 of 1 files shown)


phanen:
another idea is to await on LspRequest autocmd on codeactive until complete

```
---@param handler lsp.MultiHandler (function)
--- Handler called after all requests are completed. Server results are passed as
--- a `client_id:result` map.
```
sad

ibhagwan:
> another idea is to await on LspRequest autocmd on codeactive until complete

Since we only use opts_once with code actions , this solution is perferable IMHO.

If you don;t object I'll nerge this.

ibhagwan:
Actually I think we also need to add `_OPTS_ONCE = nil` inside the condition clause.

phanen:
I can wrap a vim.schedule in LspRequest to make thing work.
https://github.com/neovim/neovim/blob/0472119fe61a50e56382ccdd42fb8324c1f0ab44/runtime/lua/vim/lsp/client.lua#L753
https://github.com/neovim/neovim/blob/0472119fe61a50e56382ccdd42fb8324c1f0ab44/runtime/lua/vim/lsp/client.lua#L742-L756
https://github.com/neovim/neovim/blob/a768d0a95b52bc22cc0b6c3d180980fc354770e3/runtime/lua/vim/lsp/rpc.lua#L283-L286


handler executed after autocmd... from line number
https://github.com/neovim/neovim/blob/a768d0a95b52bc22cc0b6c3d180980fc354770e3/runtime/lua/vim/lsp/rpc.lua#L428-L430
https://github.com/neovim/neovim/blob/a768d0a95b52bc22cc0b6c3d180980fc354770e3/runtime/lua/vim/lsp/rpc.lua#L398-L400
but there's so many wrapper, so the order can be random

ibhagwan:
> but there's so many wrapper, so the order can be random

Why would this approach be prefarable to the code in this PR?

phanen:
yeah, not a better idea, we can try this way

phanen:
> Actually I think we also need to add `_OPTS_ONCE = nil` inside the condition clause.

done

Issue #1712, opened 2025-01-15
Bug: `ModeChaged` event being emitted when scrolling


OXY2DEV:
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [X] I have read the Wiki including the Advanced section
- [X] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux(android)

### Shell

fish

### Neovim version (`nvim --version`)

NVIM v0.10.2 Build type: Release LuaJIT 2.1.1731601260

### Fzf version (`fzf --version`)

0.56 (devel)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--height=40% --layout=reverse --info=inline --border --margin=1 --padding=1

### Is the problem reproducible with `mini.sh`?

- [X] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
  keymap = {
    built-in = {
      ["<PageUp>"] = "preview-page-up",
      ["<PageDown>"] = "preview-page-down"
    }
  }
})
```

### Describe the bug / steps to reproduce

Create an Autocmd like so,

```lua
vim.api.nvim_create_autocmd({ "ModeChanged" }, {
  callback = function ()
    vim.print(vim.api.nvim_get_current_buf());
  end
});
```

Open `:FzfLua files` and scroll through any files.
![Screenshot_2025-01-15-08-50-08-342_com termux-edit](https://github.com/user-attachments/assets/5da9f32c-8bd4-4632-8047-4247b83a6695)

> 3, 7 are the preview buffers for the files.

### Expected behavior 

Do not emit any `ModeChanged` events.

### Relevance

This causes rendering issues with `markview.nvim`(`ModeChanged` isn't denounced due to performance reasons).

OXY2DEV:
Reproduction steps are done inside of `mini.sh`.

ibhagwan:
Ty @OXY2DEV, this happens due to the usage `vim.cmd('norm ^D')`, etc to scroll.

Should not trigger post https://github.com/ibhagwan/fzf-lua/commit/e59fac8ac4188984a2b69d4a7d1375ccb866dbbb.

PR #2428, opened 2025-11-16
fix: ignore ModeChanged event in zz

phanen:
Similar to https://github.com/ibhagwan/fzf-lua/issues/1712

Type in live_grep
-> `set_cursor_hl`
-> Modechanged repeat many times: `t:n`/`n:t`
-> callback repeat

This fix a bug in my input method switching plugin, which
simply use dbus asyncly on Modechanged.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -1133,7 +1133,7 @@ function M.line_count(win, buf)
   end
 end
 
-function M.zz()
+local function _zz()
   -- skip for terminal buffers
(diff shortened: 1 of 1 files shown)


phanen:
I believe it's harmless since some utils.zz has already been called in noau https://github.com/ibhagwan/fzf-lua/blob/b19dabe6ddc0b1cef4e0b30e864aca88fdabc551/lua/fzf-lua/win.lua#L1729

ibhagwan:
Ty @phanen!

Seems like better error handling too with pcall

PR #2436, opened 2025-11-20
fix(init): fname escape `g:fzf_lua_root` & `g:fzf_lua_directory`

AgatZan:
At `fzf-lua/init.lua:12` `:source`, expect fname. If `g:fzf_lua_root` has some special characters, it expands to a relative path and it often throws an error.

Maybe `vim.fn.fnameescape` should be placed at `path.normalize` or so on.

### DIFF
--- a/tests/init_spec.lua
+++ b/tests/init_spec.lua
@@ -2,7 +2,7 @@
 local MiniTest = require("mini.test")
 local helpers = require("fzf-lua.test.helpers")
 local child = helpers.new_child_neovim()
-local expect, eq = helpers.expect, helpers.expect.equality
+local expect, eq, no_err = helpers.expect, helpers.expect.equality, helpers.expect.no_error
 local new_set = MiniTest.new_set
 
 local T = helpers.new_set_with_child(child, nil, { winopts = { col = 0, row = 1 } })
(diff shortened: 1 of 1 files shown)


ibhagwan:
Ty @AgatZan, I’ll have to see how this affects the misc `path.join`’s in the code, can give you an example of a special character causing this to fail so I can test?

Maybe only inside  `source_vimL` need to be escaped?

AgatZan:
> can give you an example of a special character causing this to fail so I can test?


Thanks for your work @ibhagwan. In my case, it is `g:fzf_lua_root = "F:/#plugins/fzf-lua"`. So, for the test case, I think you should use these characters: \t\n\*?[{$%#'"|!<.

I don't have access to any Linux systems, but my Github Action tests give that at `ubuntu-latest`:

```
Run nvim --headless +'source test/test.lua' +qa
Done
Hello, World!

Run nvim --headless +'source #test/test.lua' +qa
Error detected while processing command line:
E194: No alternate file name to substitute for '#'
```

> Maybe only inside source_vimL need to be escaped?


Maybe, but `utils.load_profile_fname` uses `g:fzf_lua_directory` and also throws this error.

ibhagwan:
> Maybe, but utils.load_profile_fname uses g:fzf_lua_directory and also throws this error.

It is still safer to escape both these locations than the global vars, what I'm most concerned is affecting this part:
https://github.com/ibhagwan/fzf-lua/blob/3b4a5d13adeb33f387d744b6ed777bb5401cba94/lua/fzf-lua/shell.lua#L158


Unfortunately I cannot test this as lazy.nvim also fails to load the local plugin when it contains special chars.

ibhagwan:
> Maybe, but utils.load_profile_fname uses g:fzf_lua_directory and also throws this error.

I don't get this error, the function calls lua `dofile` which doesn't have an issue with nvim special chars (the only issue seems to be with `vim.cmd`), I created a profile named `#test` and I can load it successfully:
image


I believe this is the only change needed in this PR:
```diff
diff --git a/lua/fzf-lua/init.lua b/lua/fzf-lua/init.lua
index 9e1a16f8..3852be92 100644
--- a/lua/fzf-lua/init.lua
+++ b/lua/fzf-lua/init.lua
@@ -9,7 +9,7 @@ do
   local function source_vimL(path_parts)
     local vimL_file = path.join(path_parts)
     if uv.fs_stat(vimL_file) then
-      vim.cmd("source " .. vimL_file)
+      vim.cmd("source " .. vim.fn.fnameescape(vimL_file))
       -- print(string.format("loaded '%s'", vimL_file))
     end
   end
```

PR #2437, opened 2025-11-21
feat: new picker for `:foldmarker`

neur1n:
Add a new picker for `:foldmarker` with an option `on_jump`. For example:
```lua
require("fzf-lua").setup({
  foldmarkers = {
    on_jump = function()
      vim.cmd("normal! zz")
    end,
  }
})
```
can be used to center the jumped line on the screen.

https://github.com/user-attachments/assets/c0dc50a5-97f2-421a-b90f-95d25b107408


neur1n:
I was following the existing codes in `nvim.lua` to prepare preview contents. I'm not sure if this is using built-in preview.

ibhagwan:
@neur1n, ty for submitting, unfortunately this isn't using the builtin preview nor is the function is written in a way that is resumeable (if the buffer changed in between resumes), it should probably be written closer to what we currently have with `blines`.

I will rewrite this after #2320 is merged.

ibhagwan:
On 2nd thought I don’t think this picker is needed, what do you think of the below?
```lua
:lua FzfLua.grep_curbuf({ no_esc=true, search="<regex including fold markers>" })
```

neur1n:
> On 2nd thought I don’t think this picker is needed, what do you think of the below?
> 
> ```lua
> :lua FzfLua.grep_curbuf({ no_esc=true, search="<regex including fold markers>" })
> ```

Looks good. How may I search for `{{{` and `}}}` at once? I didn't succeed with the `|` operator.

nenahp:
`lua FzfLua.grep_curbuf({ no_esc=true, search = '({{{)|(}}})', rg_opts = '--pcre2' })`

or `lua FzfLua.blines({query = '{{{ |  }}}'})`

ibhagwan:
Closing since can be replaced by `grep_curbuf search=..|blines query=...`

PR #2450, opened 2025-12-01
feat: cli/fzf-tmux/fzf-native improvements

ibhagwan:
@phanen, still WIP replacing all my custom fzf shell binds with fzf-lua, `cli.lua` is one of the best improvements in a while :-)

FYI, still haven't been able to solve this, turns out that when executing `cli.lua` from within a zsh widget it doesn't work using `FZF_DEFAULT_COMMAND` (even if setting it to a simple command as `echo "foo\nbar"`), reminds me of the behavior experienced with Windows CI which required the `pipe_cmd` option to be added.

If you wanna take a look, this is part of the code I added in my `.zshrc`, doesn't work without `pipe_cmd=true` (interface opens without executing any command):
> NOTE: changed the enter bind in `~/.config/fzf-lua/init.lua` to not exec `nvim`
```zsh
# find fzf-lua
for dir in "$HOME/Sources/nvim/fzf-lua" "$HOME/.local/share/nvim/lazy/fzf-lua"
do; [ -f ${dir}/scripts/cli.lua ] && FZF_LUA_CLI="${dir}/scripts/cli.lua"; done

# ctrl-t for files, ctrl-f for live_grep
if [ ! -z $FZF_LUA_CLI ]; then
    eval_widget() {
        local wn=$1
        local cmd=$2
        eval "${wn}() { local result=\$(nvim -l \"$FZF_LUA_CLI\" ${cmd} pipe_cmd=true </dev/tty | jq -r '.[] | .path' | while read -r item; do; echo -n -E \"\${(q)item} \"; done); zle reset-prompt; LBUFFER+=\$result }"
        eval "zle -N ${wn}"
    }
    typeset -A fzf_widgets
    fzf_widgets=(t files f live_grep)
    for key cmd in "${(@kv)fzf_widgets}"; do
        local wn="_fzf-${cmd}-widget"
        eval_widget ${wn} ${cmd}
        for m in emacs vicmd viins; do
            eval "bindkey -M $m '^${key}' ${wn}"
        done
    done
fi
```

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -123,8 +123,7 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
   end
 
   local co = coroutine.running()
-  local jobstart = _G.fzf_jobstart or opts.is_fzf_tmux and vim.fn.jobstart or
-      utils.termopen
+  local jobstart = _G.fzf_jobstart or opts.is_fzf_tmux and vim.fn.jobstart or utils.termopen
   local shell_cmd = utils.__IS_WINDOWS
       -- MSYS2 comes with "/usr/bin/cmd" that precedes "cmd.exe" (#1396)
       and { "cmd.exe", "/d", "/e:off", "/f:off", "/v:off", "/c" }
--- a/lua/fzf-lua/profiles/cli.lua
+++ b/lua/fzf-lua/profiles/cli.lua
@@ -95,10 +95,18 @@ _G.fzf_pty_spawn = function(cmd, opts)
   return pid
 end
 
+local HAS_TMUX = os.getenv("TMUX")
+
 return {
-  { os.getenv("TMUX") and "fzf-tmux" or "fzf-native", },
+  -- always use fzf-tmux profile for fzf native border labels
+  -- nullify "--tmux" in `fzf_opts` if tmux isn't detected
+  { "fzf-tmux" },
   desc = "run in shell cmdline",
-  fzf_opts = { ["--height"] = "50%" },
+  fzf_opts = {
+    ["--height"] = "50%",
+    ["--border"] = HAS_TMUX and "rounded" or "top",
+    ["--tmux"] = (function() return not HAS_TMUX and false or nil end)(),
+  },
   actions = {
     files = {
       true,
@@ -107,6 +115,10 @@ return {
       ["enter"] = function(s, o)
         local entries = vim.tbl_map(
           function(e) return FzfLua.path.entry_to_file(e, o) end, s)
+        entries = vim.tbl_map(function(e)
+          e.path = FzfLua.path.relative_to(e.path, vim.uv.cwd())
+          return e
+        end, entries)
         if ffi and #entries == 1 then
           posix_exec(fn.exepath("nvim"), entries[1].path,
             entries[1].line and ("+" .. entries[1].line) or nil,
--- a/lua/fzf-lua/profiles/default-prompt.lua
+++ b/lua/fzf-lua/profiles/default-prompt.lua
@@ -26,6 +26,7 @@ return {
   },
   args                 = prompt("Args"),
   oldfiles             = prompt("Oldfiles"),
+  undotree             = prompt("Undotree"),
   quickfix             = prompt("Quickfix"),
   quickfix_stack       = prompt("Quickfix Stack"),
   loclist              = prompt("Location"),
--- a/lua/fzf-lua/profiles/default-title.lua
+++ b/lua/fzf-lua/profiles/default-title.lua
@@ -30,6 +30,7 @@ return {
   },
   args                 = title("Args"),
   oldfiles             = title("Oldfiles"),
+  undotree             = title("Undotree"),
   quickfix             = title("Quickfix List"),
   quickfix_stack       = title("Quickfix List Stack"),
   loclist              = title("Location List"),
--- a/lua/fzf-lua/profiles/fzf-native.lua
+++ b/lua/fzf-lua/profiles/fzf-native.lua
@@ -1,9 +1,22 @@
 return {
   { "default-title" }, -- base profile
   desc = "fzf with `bat` as native previewer",
-  winopts = { preview = { default = "bat" } },
+  winopts = {
+    preview = {
+      default = "bat",
+      border = function(_, m)
+        assert(m.type == "fzf")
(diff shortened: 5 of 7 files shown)


ibhagwan:
Not sure why the CI in `win_spec.lua:L294-L295` is so flaky, it already waits for fzf load event and sleeps for 250ms on Windows, maybe need a bit of sleep on MacOS too?

phanen:
this section don't have retry, maybe just add it
```diff
diff --git a/tests/win_spec.lua b/tests/win_spec.lua
index 1da8bf1..e4a2e16 100644
--- a/tests/win_spec.lua
+++ b/tests/win_spec.lua
@@ -252,7 +252,11 @@ T["win"]["previewer"]["split flex hidden"] = function()
 end
 
 -- E: toggle_behavior=extend, B: builtin previewer, F: "echo {}" preview
-T["win"]["toggle"] = new_set({ parametrize = { { "EB" }, { "EF" }, { "B" }, { "F" } } })
+T["win"]["toggle"] = new_set({
+  parametrize = { { "EB" }, { "EF" }, { "B" }, { "F" } },
+  n_retry = not
+      helpers.IS_LINUX() and 5 or nil
+})
 T["win"]["toggle"][""] = new_set(
   {
     parametrize = {
```

phanen:
> If you wanna take a look, this is part of the code I added in my `.zshrc`, doesn't work without `pipe_cmd=true` (interface opens without executing any command):

No.. your script work in my zsh. no matter pipe_cmd=false/true

ibhagwan:
> > If you wanna take a look, this is part of the code I added in my `.zshrc`, doesn't work without `pipe_cmd=true` (interface opens without executing any command):
> 
> No.. your script work in my zsh. no matter pipe_cmd=false/true

Interesting, I’ll investigate.

ibhagwan:
I found the issue, remove ` </dev/tty` and call the widget from the keymap and you'll see it.
> running `FzfLua git_files` or `_fzf-files-widget` directly works
> only fails when calling via the keymap with `ctrl-t` 
```zsh
    # this fails the ctrl-t keymap, ctrl-f will fail on ctrl-g
    FzfLua() { ${NVIM:-nvim} -l "$FZF_LUA_CLI" $@ | jq -r ".[] | .path" }
    # this works from the keymap
    # FzfLua() { ${NVIM:-nvim} -l "$FZF_LUA_CLI" $@ </dev/tty | jq -r ".[] | .path" }
    eval_widget() {
        eval "${1}() {
            local res=\$(FzfLua ${2} | while read -r line; do; echo -n -E \"\${(q)line} \"; done);
            zle reset-prompt;
            LBUFFER+=\$res
        }"
        eval "zle -N ${1}"
    }
    typeset -A fzf_widgets
    fzf_widgets=(t files f live_grep)
    for key cmd in "${(@kv)fzf_widgets}"; do
        local wn="_fzf-${cmd}-widget"
        eval_widget ${wn} ${cmd}
        for m in emacs vicmd viins; do
            eval "bindkey -M $m '^${key}' ${wn}"
        done
    done
```

The reason is apparently the ZLE (zsh line editor):
https://unix.stackexchange.com/questions/595350/using-fzf-in-a-zle-key-binding
https://unix.stackexchange.com/questions/780624/omz-reload-in-zsh-widget

> Code running in a custom zle widget (i.e. a zsh function that's bound to a key) runs with standard input connected to /dev/null. The reason for this is that zsh is in the middle of reading from the terminal, and commands run from widget should not interfere with this.

ibhagwan:
This is quite cool, using fzf's `--border-label` to add the title+hidden/ignore markers:
image

ibhagwan:
@phanen, need your input on the below, just added it to be able to support preview rotate on fzf-tmux, this got me thinking we can refactor this bind section and get rid of all the SIGWINCH hacks and the toggle preview feedkey in `winopts.split` and instead just initiate the F4/F5/F6 binds directly from fzf with a `transform` instead - what do you think?

I think this will simplify the code greatly with the downside of maybe reduced support for skim or earlier versions of fzf (<0.46, really old)?

https://github.com/ibhagwan/fzf-lua/blob/3855213eb41fbc4e0d97af1b28d139af26c23a12/lua/fzf-lua/win.lua#L214-L228

phanen:
fine but then how to merge user's keymap/actions?

phanen:
iirc multiple resize:+transform is used  before becasuse user won't custom action on resize or it is usually harmless

tcell support f13+ keys, not sure if fzf can now...but this is not choice...

maybe just use --listen... it seems the way fzf recommand. nvim nighly has a vim.net.request although it's just a wrapper of curl..

phanen:
another aggressive idea, can avoid --listen

- create some transform binds as pool 

use pool since async transform need lock var, which maybe block other action if user input rapidly.. otherwise can also run each user input in its own coroutine

- create tmap to dispatch keymap.fzf/actions to transform bind

for any input, it will just register an action, then pick a bind from pool, feedkey/chansend to consume it

to manually call fzf action is nothing different from any input.

ot: fzf cannot wait a transform from another transform.. this don't matter in lua, we can schedule or use coroutine out whenever needed

user can still keymap.set if they want, it won't access the pool

user may want a fast bind don't go through redirect... maybe provide an arg.direct so this key is compile to cli flags directly

phanen:
oop forgot tmux and cli profile...

ibhagwan:
> another aggressive idea, can avoid --listen
> 
> * create some transform binds as pool
> 
> use pool since async transform need lock var, which maybe block other action if user input rapidly.. otherwise can also run each user input in its own coroutine
> 
> * create tmap to dispatch keymap.fzf/actions to transform bind
> 
> for any input, it will just register an action, then pick a bind from pool, feedkey/chansend to consume it
> 
> to manually call fzf action is nothing different from any input.
> 
> ot: fzf cannot wait a transform from another transform.. this don't matter in lua, we can schedule or use coroutine out whenever needed
> 
> user can still keymap.set if they want, it won't access the pool
> 
> user may want a fast bind don't go through redirect... maybe provide an arg.direct so this key is compile to cli flags directly

I like this idea, it’s been long I wanted to get rid of keymap.fzf|builtin/actions and have a single `binds` table for everything,

> oop forgot tmux and cli profile...

That is indeed a bit of a problem for this approach :)

We can do so,thing aggressive the other way around and change all binds to fzf callbacks but I won’t think it’s a good approach, will have issues with skim, limit the bind keys to a subset and be slow on windows.

My old thought process on this was to have a single binds table and we can dynamically decide if this is an Fzf native bind or a neovim bind, it can be seamless for the user.

PR #2455, opened 2025-12-05
fix(lua_ls): make `winopts.preview` optional

MariaSolOs:
`lua_ls` error:
image

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -61,7 +61,7 @@ end
 ---between 0-1 will represent percentage of `vim.o.columns` (0: leftmost, 1: rightmost)
 ---if >= 1 will attempt to place the float in the exact screen column.
 ---@field col? number
----@field preview fzf-lua.config.PreviewOpts Preview window configuration.
+---@field preview? fzf-lua.config.PreviewOpts Preview window configuration.
 ---@field split? string|function|false Neovim split command to use for fzf-lua interface, e.g `belowright new`.
 ---@field backdrop? number|boolean Backdrop opacity, 0-100, or false to disable (requires Neovim >= 0.10).
 ---@field fullscreen? boolean Open fzf-lua in fullscreen.


ibhagwan:
Ty @MariaSolOs, there are probably a few more of these since the latest changes in order to move to emmylua

phanen:
This was used mainly to make emmylua_ls happy before (warnings as error is not enabled in ci currently due to flaky).

I don't find any efficient way to quickly recursively define a parital types for another type to be used in user config. (otherwise it take another thousand of line to make thing correct... maybe we can generate it via meta data from `emmylua_doc_cli`)

btw, I use this in my dotfile like I added in readme...
```
  ---@diagnostics disable: missing-fields
  opts = {}
  ---@diagnostics enable: missing-fields
```

ibhagwan:
> I don't find any efficient way to quickly recursively define a parital types for another type to be used in user config.

@phanen, what’s wrong with the defining the fields optional as in this PR? Does this create an issue with emmylua? Maybe we can open a feature request upstream?

MariaSolOs:
> btw, I use this in my dotfile like I added in readme...

I know how to disable the warnings, but this is a legitimate incorrect type (`preview` can be missing in `winopts`).

phanen:
Yeah, this should be correct. opened issue: https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/871

phanen:
Latest emmylua_ls support `Partial<T>`, which should be define in fzf-lua side, then we need switch from lua_ls to emmylua_ls check.

and that also means user should use emmylua_ls instead of lua_ls at that time.

ibhagwan:
> Latest emmylua_ls support `Partial<T>`, which should be define in fzf-lua side, then we need switch from lua_ls to emmylua_ls check.
> 
> and that also means user should use emmylua_ls instead of lua_ls at that time.

[0.18](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/releases/tag/0.18.0) was just released

Issue #2025, opened 2025-05-03
Bug: Prompt shift when using `hide` profile


MariaSolOs:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

MacOS, Linux

### Shell

fish

### Neovim version (`nvim --version`)

v0.12.0-dev-284+g5a2edc483d

### Fzf version (`fzf --version`)

0.61.1

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

"--color=fg:#f8f8f2,bg:#0e1419,hl:#e11299,fg+:#f8f8f2,bg+:#44475a,hl+:#e11299,info:#f1fa8c,prompt:#50fa7b,pointer:#ff79c6,marker:#ff79c6,spinner:#a4ffff,header:#6272a4 --cycle --pointer=▎ --marker=▎"

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
  { 'border-fused', 'hide' }
})

vim.keymap.set('n', '<leader>fb', function()
  require('fzf-lua').lgrep_curbuf {
    winopts = { height = 0.6, width = 0.5, preview = { vertical = 'up:70%' } }
  }
end, { desc = 'Grep current buffer' })

vim.keymap.set('n', '<leader>f<', '<cmd>FzfLua resume<cr>', { desc = 'Resume last fzf command' })
```

### Describe the bug / steps to reproduce

When using the `hide` profile, the prompt of the picker shifts (and is then restored to the correct size after typing a character) when invoking `FzfLua resume` after using the buffer grep picker.

See the video below for a description of the issue:

https://github.com/user-attachments/assets/c7560263-6e03-47a0-9b2a-71c14eb88817

Interestingly enough, this doesn't seem to happen with other pickers (like "find files"), and it is fixed if I remove the `hide` profile from my configuration.

ibhagwan:
This seems like yet another version of #1929, unfortunately this issue is upstream introduced with neovim 0.11.

I found out this also doesn’t happen when the prompt is at the top  without `--layout=default`).

[...]

PR #2458, opened 2025-12-06
fix(reflow): resume causing lines to shift

MeanderingProgrammer:
## Details

This change disables `FileType` events from triggering on the backdrop buffer. Why exactly this causes issues in the terminal buffer, I'm not sure. My best guess is that having a variety of plugins running autocommands for the backdrop causes some timing issue, maybe in the redraw cycle. I believe this solves my issue and potentially others:

- https://github.com/ibhagwan/fzf-lua/issues/2025
- https://github.com/ibhagwan/fzf-lua/discussions/2457


ibhagwan:
Ty @MeanderingProgrammer !

Issue #2472, opened 2025-12-10
Bug: Windows:  "Invalid value for width: "$COLUMNS" is not an integer"


Phys09:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Windows

### Shell

pwsh v7.5.4 (windows powershell)

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-1779+g647f11e6ae

### Fzf version (`fzf --version`)

0.67.0 (2ab923f3)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
Using lazynvim distribution in particular. The config pasted below is an excerpt from the website's documentation found here: https://www.lazyvim.org/extras/editor/fzf#fzf-lua

This config is also what originally caused the issue in the first place. A simplified config that causes the same issue (with $FZF_PREVIEW_COLUMNS instead) at the bottom.


lsp = {
        symbols = {
          symbol_hl = function(s)
            return "TroubleIcon" .. s
          end,
          symbol_fmt = function(s)
            return s:lower() .. "\t"
          end,
          child_prefix = false,
        },
        code_actions = {
          previewer = vim.fn.executable("delta") == 1 and "codeaction_native" or nil,
        },
      },

^^^ Original fzf-lua config causing issue ^^^

vvv Minimal fzf-lua config that also repeats the issue vvv


return {
  "ibhagwan/fzf-lua",
  opts = function(_, opts)
    opts.lsp.code_actions = {
      previewer = "codeaction_native",

      -- pwsh specifically needs to use $env:* prefix to attempt reading environment variables
      -- other shells can read the environment variable $FZF_PREVIEW_COLUMNS as expected
      preview_pager = "delta --side-by-side --width=$env:FZF_PREVIEW_COLUMNS",

      -- However it used to work fine without the prefix $env:* before updating
      -- preview_pager = "delta --side-by-side --width=FZF_PREVIEW_COLUMNS",

[...]

PR #2473, opened 2025-12-10
fix(symbols): better ignorecase match

phanen:
Previous matching method is a bit confusing.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -21,7 +21,8 @@ function M._default_previewer_fn()
 end
 
 function M._preview_pager_fn()
-  return vim.fn.executable("delta") == 1 and ("delta --width=$COLUMNS --%s"):format(vim.o.bg) or
+  return vim.fn.executable("delta") == 1 and
+      ("delta --width=%s --%s"):format(utils._if_win_normalize_vars("$COLUMNS"), vim.o.bg) or
       nil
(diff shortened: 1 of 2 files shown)


phanen:
what happened..
```
Error: Could not download Neovim release from https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip: Downloading asset failed: Not Found. Please visit https://github.com/neovim/neovim/releases/tag/nightly to check the asset for windows was really uploaded. Note that some assets are sometimes missing on nightly build due to Neovim's CI failure
Error: Could not download Neovim release from https://github.com/neovim/neovim/releases/download/nightly/nvim-win64.zip: Downloading asset failed: Not Found. Please visit https://github.com/neovim/neovim/releases/tag/nightly to check the asset for windows was really uploaded. Note that some assets are sometimes missing on nightly build due to Neovim's CI failure
```

phanen:
ok now, weird

ibhagwan:
Oops I missed this one I see it also fixes #2472 without code duplication, I’ll remove the other calls then.

ibhagwan:
Can't refactor this yet, if we remove all other refs to normalize we lose the var transformation whenthe user set their own pager (and not using the default auto-detect function in defaults), for now I added another call in the undotree_native previewer.

Will still be better with a refactor but we can live with this tiny dedup for now.

PR #2474, opened 2025-12-13
fix(symbols): fuzzy match as fallback

phanen:
When exact match fail use fuzzy as fallback, some lsp can return fuzzy find result.

Not sure if lsp can return the match indices so we don't need implement a heuristics match by our own.


phanen:
There's a matchfuzzypos(), but cannot be used in fast event

ibhagwan:
Ty @phanen!

PR #2476, opened 2025-12-13
feature(git_diff): add ability to diff between two specific commits

fin-w:
An additional field is added to `git_diff()`: `compare_against` (string). If it is not provided, behaviour is also changed to diff against the provided `ref` and its direct parent commit, showing what changes the `ref` commit introduced. When `compare_against` is given, the difference between commits `compare_against` and `ref` is shown. The `git_preview()` function is modified to operate from the nearest Git repository up the `cwd` path, so that the preview for `git_diff()` works.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -597,9 +597,10 @@ M.defaults.git                   = {
   },
   ---@class fzf-lua.config.GitDiff: fzf-lua.config.GitBase
   diff = {
-    cmd               = "git --no-pager diff --name-only {ref}",
+    cmd               = "git --no-pager diff --name-only {compare_against} {ref}",
     ref               = "HEAD",
-    preview           = "git diff {ref} {file}",
+    compare_against   = "",
+    preview           = "git diff {compare_against} {ref} {file}",
     preview_pager     = M._preview_pager_fn,
     multiprocess      = 1, ---@type integer|boolean
     _type             = "file",
--- a/lua/fzf-lua/providers/git.lua
+++ b/lua/fzf-lua/providers/git.lua
@@ -70,6 +70,11 @@ local function git_preview(opts, file)
   if file then
     opts.preview = opts.preview:gsub("[<{]file[}>]", file)
   end
+  -- Normalize the cwd passed to git_preview() so that it is the closest
+  -- directory up the path that is a Git repo. git_diff() passes file
+  -- paths relative to the Git repo root so to ensure the previewer
+  -- can interpret the paths correctly, it must use the repo root as cwd.
+  opts.cwd = path.git_root({ cwd = opts.cwd })
   opts.preview = path.git_cwd(opts.preview, opts)
   if type(opts.preview_pager) == "function" then
     opts.preview_pager = opts.preview_pager()
@@ -92,14 +97,30 @@ M.diff = function(opts)
   ---@type fzf-lua.config.GitDiff
   opts = config.normalize_opts(opts, "git.diff")
   if not opts then return end
-  local cmd = path.git_cwd({ "git", "rev-parse", "--verify", opts.ref }, opts)
-  local _, err = utils.io_systemlist(cmd)
-  if err ~= 0 then
+  -- Ensure that ref is a commit hash in this git repository.
+  local validation_ref = path.git_cwd({ "git", "rev-parse", "--verify", opts.ref }, opts)
+  local _, exit_status_code_ref = utils.io_systemlist(validation_ref)
+  if exit_status_code_ref ~= 0 then
     utils.warn("Invalid git ref %s", opts.ref)
     return
   end
+  -- If compare_against is given, ensure that it is a commit hash in this git repository.
+  if type(opts.compare_against) == "string" and #opts.compare_against > 0 then
+    local validation_comp = path.git_cwd({ "git", "rev-parse", "--verify", opts.compare_against },
+      opts)
+    local _, exit_code_status_comp = utils.io_systemlist(validation_comp)
+    if exit_code_status_comp ~= 0 then
+      utils.warn("Invalid git ref %s", opts.compare_against)
+      return
+    end
+  else
+    -- Default to diffing against ref and its direct parent commit.
+    opts.compare_against = opts.ref .. "^"
+  end
   opts.cmd = opts.cmd:gsub("[<{]ref[}>]", opts.ref)
+  opts.cmd = opts.cmd:gsub("[<{]compare_against[}>]", opts.compare_against)
   opts.preview = opts.preview:gsub("[<{]ref[}>]", opts.ref)
+  opts.preview = opts.preview:gsub("[<{]compare_against[}>]", opts.compare_against)
   opts = set_git_cwd_args(opts)
   if not opts.cwd then return end
   opts.preview = git_preview(opts, "{-1}")


fin-w:
@ibhagwan This is functional now, so I welcome input.

I've noticed an issue where it seems the validation logic is running git commands in the wrong folder. Not sure exactly what's happening but Neovim's PWD might be part of the problem?

In Neovim my PWD is `~/.config/nvim/` , a git repo, and I also have files open in `~/.local/share/nvim/site/pack/core/opt/fzf-lua/lua/fzf-lua/providers/` which is in a separate repo. `git_commits()` open the correct repo with my keymap
```lua
vim.keymap.set('n', '<Leader>gcc', function() fzf.git_commits({ cwd = vim.fn.expand('%:p:h') }) end, {
    desc = 'Fzf-lua: git commits (in submodule or repository)'
})
```
which sets the git repo to the one the current buffer is in. That PWD should be where the `path.git_cwd({ "git", "rev-parse", "--verify", opts.ref }, opts)` runs from too I think.

ibhagwan:
I’m thinking it may be worthwhile to use `utils.input` to ask the user for ref/parent (if not specified via opts), otherwise it’s confusing what this picker does and how is it different than `git_status` - what do you think?

fin-w:
I mean I was surprised when it wasn't comparing to the parent commit by default, hence why I'm trying to make it work how I want to use it. What is the purpose of this function, in your eyes?

ibhagwan:
> I mean I was surprised when it wasn't comparing to the parent commit by default, hence why I'm trying to make it work how I want to use it. What is the purpose of this function, in your eyes?

Ask the user for commit hashes/refs so it’s more clear what the picker is doing no?

fin-w:
I've had a bit of a look at utils.input, but I want to keep this MR focused on providing only two things:
- Enabled diffing against direct commit parent via `parent = true` (the ease of use is worth the extra variable)
- Enable diffing against an arbitrary commit via `compare_against = "abc123xyz"`

ibhagwan:
> I've had a bit of a look at utils.input, but I want to keep this MR focused on providing only two things:
> 
> * Enabled diffing against direct commit parent via `parent = true` (the ease of use is worth the extra variable)
> * Enable diffing against an arbitrary commit via `compare_against = "abc123xyz"`

Sounds good

fin-w:
Once you are happy with this code, I'll update relevant docs provided it's not a very involved process.

fin-w:
Do commits get squashed if this is accepted? I can't figure out how to set the MR to squash commits from my end, like Gitlab offers...

fin-w:
I still don't know how you use git_diff() / expect people to use it. From my perspective, by default it should diff between a commit parent and a commit when given a ref hash and nothing else. Does the function accept anything other than a commit hash? If it accepts only commit hashes, and by default compares against HEAD, it doesn't seem meaningfully different to git_status() apart from not being able to stage / unstage etc. Perhaps I'm not understanding something though.

fin-w:
I think it HAS to accept a `cwd` argument as well to function correctly i.e. to show correct results in submodules, or when viewing git info in another repository to where Neovim PWD currently is :disappointed: so still WIP until that is added too.

ibhagwan:
> Do commits get squashed if this is accepted? I can't figure out how to set the MR to squash commits from my end, like Gitlab offers...

Unless I select a specific option commits don’t get squashed, you can do it on your local repo and force push or i will do it as part of the merge.

> still don't know how you use git_diff() / expect people to use it. From my perspective, by default it should diff between a commit parent and a commit when given a ref hash and nothing else. Does the function accept anything other than a commit hash? If it accepts only commit hashes, and by default compares against HEAD, it doesn't seem meaningfully different to git_status() apart from not being able to stage / unstage etc. Perhaps I'm not understanding something though.

TBH, I haven’t used this picker much, the way I view it is similar, it’s a way to explore a commit in more detail separated by files, kind of like looking at the git_status of the commit before it got committed, in that sense I agree parent commit should be automatically selected.


> I think it HAS to accept a cwd argument as well to function correctly i.e. to show correct results in submodules, or when viewing git info in another repository to where Neovim PWD currently is 😞 so still WIP until that is added too.

All git operations in fzf-lua auto detect the git top dir from the cwd using `rev-parse  --show-toplevel`, not sure I understand the use case of the sub modules completely, not sure if it helps but you can also use `git_worktree=` and `git_dir=` which correspond to `--git-dir`, etc.

ibhagwan:
Squashed into 1 commit, feel free to change the wording

fin-w:
Ok since you agree `parent = true` should be default, I've removed it and behaviour is changed to reflect this instead. Provide `compare_against = "commithash"` to default to something other than diffing parent and child commits.

This is how I use `git_branches()` because I frequently work in git submodules, so when I want to switch branches I want to do so in the git repo the buffer belongs to: everything is relative to the current buffer, not Neovim's PWD.
```lua
vim.keymap.set('n', '<Leader>gb', function() fzf.git_branches({ cwd = vim.fn.expand('%:p:h') }) end, {
    desc = 'Fzf-lua: git branches (in submodule or repository)'
})
```

This means I expect every git operation Fzf-Lua does to be relative to the current buffer. I don't mind passing this info around but adding `cwd` or `git_cwd` to the fields passed to `git_diff()` causes the preview to fail.

```lua
        commits = {
            actions = {
                ['alt-v'] = {
                    fn = function(selected, opts)
                        assert(type(selected) == "table"
                            and type(selected[1]) == "string")
                        -- Get the commit hash from the selected picker entry.
                        local git_commit_to_show = assert(selected[1]:match("^[%*+]*[%s]*[(]?([^%s)]+)"))
                        require('fzf-lua').git_diff({
                            ref = git_commit_to_show,
                            cwd = opts.cwd,
                            git_dir = opts.git_dir,
                        })
                    end
                },
            }
        },
```

ibhagwan:
> This means I expect every git operation Fzf-Lua does to be relative to the current buffer.

This would make the behavior similar to fugitive, the issue I have with this is that all other git pickers in fzf-lua use the cwd as the git anchor, git_files/commits/status all use the cwd regardless of current buffer.

Maybe we can provide an option on how git cwd should work across all pickers in fzf-lua but that would be a different discussion.

fin-w:
I can't figure out how to modify the previewer so that is shows the correct files, I think the filepaths are relative to the wrong dir when passing `cwd` to `git_diff()` so the preview command fails, or else the command itself needs modifying:

Screenshot_20251214_150113

ibhagwan:
> I can't figure out how to modify the previewer so that is shows the correct files, I think the filepaths are relative to the wrong dir when passing `cwd` to `git_diff()` so the preview command fails, or else the command itself needs modifying:
> 
> Screenshot_20251214_150113


I’ll take a look when I get a chance, maybe we need a smarter custom previewer like git_status has.

fin-w:
> similar to fugitive

I believe setting `cwd` should be enough, it already works fine for me when using `git_branches()` and `git_commits()`. If it's possible to fix the previewer, I don't think a different setting / change in behaviour is required, for me at least.

fin-w:
The preview error is coming from the preview command not being able to find the file.

If `cwd` is given to `git_diff()` it's not normalised to the nearest git repo up the path, it's kept as-is. So relative file paths from the nearest git repo root (coming your git diff `cmd` that generates the picker menu entries, `git diff --name-only`) e.g. `lua/fzf-lua/utils.lua` don't work.

The preview DOES work when you pass in `cwd` set to the filepath to the root of the git repo your `git_diff()` call is operating on.

Should `git_diff()` perform some sort of normalisation of the `cwd` value? I don't have the above issue for the `git_branches()` picker for instance, it chooses the closest git repo up the path from the current buffer when passing in the buffer path as `cwd`. I can see `git_branches()` calls `config.normalize_opts()` but it's not clear to me what this function is doing. Should `normalize_opts()` by modified to support `git.preview` and then `git_preview()` call it before operation in the `opts` passed to `git_preview()`?

fin-w:
Hmm, perhaps the other git commands are working because git does its own path walking to find the nearest git repo. Seems like we need to normalise `cwd` to the git repo root in this case, anyhow.

fin-w:
This fixed the preview :grin: , in `git_preview()`:
```lua
opts.cwd = path.git_root({ cwd = opts.cwd })
```

ibhagwan:
> This fixed the preview 😁 , in `git_preview()`:
> 
> ```lua
> opts.cwd = path.git_root({ cwd = opts.cwd })
> ```

Nice!

ibhagwan:
Ty for this fantastic PR @fin-w!

tronikelis:
Hello, I think these changes are confusing for users and I would like to discuss if this could be reverted:

> hence why I'm trying to make it work how I want to use it. What is the purpose of this function, in your eyes?

The picker is called `git_diff` hence it should behave like `git diff` which it (previously) did:

```
git diff [<options>] [--] [<path>...]
           This form is to view the changes you made relative to the index (staging area for the next commit). In other
           words, the differences are what you could tell Git to further add to the index but you still haven’t. You
           can stage these changes by using git-add(1)
```

Previously it behaved like this when called with no arguments, just as everyone would expect, we all use `git diff` so `FzfLua git_diff` I expect should behave the same, I think it's a bad idea making it different than the standard git behavior and expecting all users to figure out what diff they are looking at

> I still don't know how you use git_diff() / expect people to use it. From my perspective, by default it should diff between a commit parent and a commit when given a ref hash and nothing else.

> From my perspective, by default it should diff between a commit parent and a commit 

Hmm, I disagree based on the argument that git which is the de facto tool, does do it differently, and we should do what git does by default.


> hence why I'm trying to make it work how I want to use it.

Seems like a good idea to create your own custom picker then, changing globals like this for everyone is bad UX

ibhagwan:
> Hello, I think these changes are confusing for users and I would like to discuss if this could be reverted:
> 
> > hence why I'm trying to make it work how I want to use it. What is the purpose of this function, in your eyes?
> 
> The picker is called `git_diff` hence it should behave like `git diff` which it (previously) did:
> 
> ```
> git diff [<options>] [--] [<path>...]
>            This form is to view the changes you made relative to the index (staging area for the next commit). In other
>            words, the differences are what you could tell Git to further add to the index but you still haven’t. You
>            can stage these changes by using git-add(1)
> ```
> 
> Previously it behaved like this when called with no arguments, just as everyone would expect, we all use `git diff` so `FzfLua git_diff` I expect should behave the same, I think it's a bad idea making it different than the standard git behavior and expecting all users to figure out what diff they are looking at
> 
> > I still don't know how you use git_diff() / expect people to use it. From my perspective, by default it should diff between a commit parent and a commit when given a ref hash and nothing else.
> 
> > From my perspective, by default it should diff between a commit parent and a commit
> 
> Hmm, I disagree based on the argument that git which is the de facto tool, does do it differently, and we should do what git does by default.
> 
> > hence why I'm trying to make it work how I want to use it.
> 
> Seems like a good idea to create your own custom picker then, changing globals like this for everyone is bad UX

These are valid points @tronikelis, I'll consider an approach to perhaps accomodate both.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/issues/2615

PR #2478, opened 2025-12-14
fix(lsp): jump1 should just hide window rather than kill session

phanen:
git blame tell me hide profile don't exist at that time?
I think it's ok to just hide the window.

So we can still unhide the window after gotodef trigger jump1.

### DIFF
--- a/lua/fzf-lua/providers/lsp.lua
+++ b/lua/fzf-lua/providers/lsp.lua
@@ -60,10 +60,6 @@ local function locations_to_entries(locations, enc)
 end
 
 local jump_to_location = function(opts, result, enc)
-  -- exits the fzf window when use with async
-  -- safe to call even if the interface is closed
-  utils.fzf_exit()
-
   local action = opts.jump1_action
   if action then
     local entries = locations_to_entries({ result }, enc)


phanen:
`opts.async=true` behavior is not change

https://github.com/ibhagwan/fzf-lua/blob/5edc7c55234d88e3fd333a8b892d2b0d63b6935d/lua/fzf-lua/providers/lsp.lua#L615-L616

ibhagwan:
What is the user isn’t using the `hide` profile, this changed the default behavior no?

phanen:
not tested. it's sync op, so i suspect this line can be even removed

ibhagwan:
> not tested. it's sync op, so i suspect this line can be even removed

On sync the interface isn’t being opened and no window is created, can be removed indeed.

phanen:
edit from mobile phone, it seems @copilot cannot help with amend

phanen:
@copilot can u hear me

ibhagwan:
Squashed to 1 cleanup commit, unless more changes are coming I’ll merge, lmk.

phanen:
ok

PR #2487, opened 2025-12-22
fix: b:term_title can be nil

phanen:
providers/buffers.lua:136: attempt to index local 'term_title' (a nil
value)

not sure why/when this happened, I guess this can happened after `:so
Session.vim`


ibhagwan:
Ty @phanen !

PR #2491, opened 2025-12-28
copy extmarks

phanen:
- **chore: trivial linter fixes**
- **feat(previewer): copy extmarks on scroll**

Learn form nvim-bqf's code.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -601,6 +601,33 @@ function Previewer.base:scroll(direction)
   self.win:update_preview_scrollbar()
   self:update_render_markdown()
   self:update_ts_context()
+  self:copy_extmarks()
+end
+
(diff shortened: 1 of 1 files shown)


phanen:
Mainly to copy some lsp highlight, e.g. `@class fzf-lua.xx` is not properly highlighted due to limit of tree-sitter-luadoc parser

phanen:
Mainly to copy some lsp highlight, e.g. `@class fzf-lua.xx` is not properly highlighted due to limit of tree-sitter-luadoc parser





semantic highlighting is also updated on WinScrolled... so to make the highlight exist must scroll to the pos in the source buffer.
https://github.com/neovim/neovim/blob/8a94daf80eadbd9179768fe3d7da2f06c81dc740/runtime/lua/vim/lsp/semantic_tokens.lua#L193-L199

ibhagwan:
Ty @phanen !

PR #2493, opened 2025-12-30
feat(grep): make rg_globs work with search_paths

phanen:
Mainly to make `:FzfLua lgrep_quickfix` work with glob which use `opts.search_paths`.

(closed by ripgrep before https://github.com/BurntSushi/ripgrep/pull/3147)

### DIFF
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -63,6 +63,7 @@ local function normalize_live_grep_opts(opts)
   -- disable treesitter as it collides with cmd regex highlighting
   opts = opts or {}
   opts._treesitter = false
+  if opts.winopts then opts.winopts.treesitter = nil end
 
   ---@type fzf-lua.config.Grep
   opts = config.normalize_opts(opts, "grep")


ibhagwan:
Not sure if this is working as expected, when I add a glob  the search path are being removed, is this what you wanted to happen?

Before adding a glob:
IMG_3877


After adding a glob, search_paths are removed (basically a normal live grep):
IMG_3878

phanen:
> the search path are being removed

~yes? maybe we should add a new option here?~

no, that's not correct.

phanen:
maybe I can add a non-exist file as placeholder.
This is ai answer: `--max-depth=0`..

Another issue it the glob rules are different from rg.

ibhagwan:
> This is ai answer: --max-depth=0..

Nice trick AI :)

Which model are you using?

I recently gave calude opus 4.5 the somewhst complicated task of modifying gen_options.lua to create OPTIONS.md by modifying the emmylua comments, it succeeded, I also tried Gemini and Grok, Gemini croaked and grok solution was incomplete.

I haven’t had a chance to deep dive into the code yet, I’ll upload it and you can see what it did too.

phanen:
Claude Sonnet 4.5. I feed `gO` of ripgrep man to it.

nifoc:
```
if opts.winopts then opts.winopts.treesitter = nil end
```

This breaks if `winopts` is a function. Since `winopts_fn` appears to be deprecated, how can we do (for example) [this](https://github.com/ibhagwan/fzf-lua/wiki#can-i-set-the-window-size-dynamically)?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/940ffbfb54842b1b2bdd7aacc35db998a952b616

@nifoc, should work now

PR #2496, opened 2026-01-02
feat(actions): alt-d to delete command/search history

phanen:
`render_crlf=true` to avoid mess up the index.
A history entry can be multi-line.


ibhagwan:
What do you think of making this `ctrl-x` consistent with other delete operations (buffers, git branches, etc)?

phanen:
done

ibhagwan:
Tysm @phanen !

PR #2498, opened 2026-01-03
feat: allow register_ui_select via setup options

netmute:
This allows configuration in e.g. lazy.nvim to simply be:
```
opts = {
    register_ui_select = true,
},
```
instead of needing to override the setup() call.

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -137,6 +137,14 @@ Example, opening a floating help window at the top of screen with single border:
     })
 ```
 
+#### setup.register_ui_select
+
+Type: `boolean|table|function`, Default: `false`
+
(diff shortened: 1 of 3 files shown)


ibhagwan:
Ty @netmute.

I think we should also accept a table / function as an arguments (not only a boolean) to the setup call to match the existing functionality.

netmute:
Updated to accept table / function 👍

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/a03d68e40eea835a1cdbd9f93049708dab3621e6

Ty @netmute, merged manually as I couldn’t override your fork due to permissions, modified the options to be named simply `ui_select` as the “register” part is somewhat implied:

```lua
opts = {
    ui_select = true,
},
```

Issue #1659, opened 2024-12-31
Bug: unwanted "i" automatically inserted everywhere


dimbleby:
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [X] I have read the Wiki including the Advanced section
- [X] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.11.0-dev-1449+gb5cb69f8a4

### Fzf version (`fzf --version`)

0.57.0 (fd513f8a)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [X] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

### Describe the bug / steps to reproduce

eg `:lua require('fzf-lua').buffers()`, per this asciinema recording https://asciinema.org/a/Qu41Lixu9zDtS9lJfTPsO007w

I see that https://github.com/ibhagwan/fzf-lua/commit/43c3a47c28d4fa93e18450516dbf7fc14e102a42 seems to have been in this ballpark, so perhaps you are aware of this already

ibhagwan:
> I see that https://github.com/ibhagwan/fzf-lua/commit/43c3a47c28d4fa93e18450516dbf7fc14e102a42  seems to have been in this ballpark, so perhaps you are aware of this already

Correct, are you experiencing this with HEAD?
How you tried with `mini.sh`?
Do you have an `autocmd` in your config for TermOpen?

ibhagwan:
My hunch was right, you have an autocmd for TermOpen that already executes `startinsert`:
https://github.com/dimbleby/dotfiles/blob/298242bf6e1102a5ca993894ea4d8eb24c63fa05/.config/nvim/lua/config/autocmds.lua#L2-L7.

Fixed in https://github.com/ibhagwan/fzf-lua/commit/c054a21cd22cef41c62ce96247040abccaab8dbc

dimbleby:
haha, slightly stalker-ish but nice detective work!  Thanks for the fix

ibhagwan:
> haha, slightly stalker-ish but nice detective work! Thanks for the fix

[...]

PR #2501, opened 2026-01-05
custom builtin previewer

phanen:
- **fix: some error when custom picker**
- **fix: detect filetype from content when no filename is given**

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1310,7 +1310,7 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
   pcall(api.nvim_win_call, win, function()
     local cached_pos = self.cached_bufnrs[buf]
     if type(cached_pos) ~= "table" then cached_pos = nil end
-    local lnum, col = entry.line, entry.col or 0
+    local lnum, col = entry.line, math.max(1, entry.col or 1)
     if not lnum or lnum < 1 then
       -- set win option is slow with bigfile
       utils.wo.cursorline = false
@@ -1319,7 +1319,7 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
       return
     end
 
-    self.orig_pos = { lnum, math.max(1, col) - 1 }
+    self.orig_pos = { lnum, col - 1 }
     api.nvim_win_set_cursor(win, cached_pos or self.orig_pos)
     self:maybe_set_cursorline(win, self.orig_pos)
 
@@ -1335,7 +1335,7 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
       local reg = vim.F.npcall(vim.regex, regex)
       if reg then
         if regex ~= regex:lower() then
-          regex_start, regex_end = reg:match_line(buf, lnum - 1, math.max(1, col) - 1)
+          regex_start, regex_end = reg:match_line(buf, lnum - 1, col - 1)
         else
           local line = api.nvim_buf_get_lines(buf, lnum - 1, lnum, false)[1] or ""
           regex_start, regex_end = reg:match_str(line:sub(col):lower())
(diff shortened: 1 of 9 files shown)


phanen:
not sure if my commit related to the failure. will see it later.

phanen:
```
Author: zeertzjq <zeertzjq@outlook.com>
Date:   Sat Jan 3 17:55:07 2026 +0800

    fix(terminal): inconsistent mode change when switching buffer (#37215)
    
    Problem:  When switching to another terminal buffer in Terminal mode,
              usually Nvim stays in Terminal mode, but leaves Terminal mode
              if the old terminal buffer was deleted.
    Solution: Don't always leave Terminal mode when active terminal buffer
              is deleted. Instead let terminal_check_focus() decide that.

 src/nvim/terminal.c                           | 10 ++++++++--
 test/functional/terminal/ex_terminal_spec.lua | 30 +++++++++++++++++++++++++++++-
 2 files changed, 37 insertions(+), 3 deletions(-)
```

phanen:
no better idea to fix, I just remove feedkey now.

ibhagwan:
I see this will also close #2034.

btw, doesn’t seem like `utils.ensure_startinsert` is ever called.

phanen:
> btw, doesn’t seem like `utils.ensure_startinsert` is ever called.

That still fail non-nightly test. Now only remove feedkey on nightly.

ibhagwan:
Now that CI is fixed on nightly is this ready to be merged or you need more time?

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -864,6 +862,7 @@ end
 function Previewer.buffer_or_file:check_bcache(entry)
   if entry.do_not_cache then return end
   local key = self:key_from_entry(entry)
+  if not key then return end
```
ibhagwan, line 864: Can there be a situation where all entry properties are empty (bufnr/uri/path) or is it just for a custom previewer?

What is the use case of a custom previewer that inherits from buffer_or_file but doesn’t have a valid entry (with at least bufnr/uri/path)?

lua/fzf-lua/previewer/builtin.lua
```diff
@@ -1369,14 +1368,10 @@ end
...
-  if filepath then
-    if self.opts.cwd then
-      filepath = path.relative_to(entry.path, self.opts.cwd)
-    end
-    filepath = path.HOME_to_tilde(filepath)
-  end
-  local title = entry.debug and "[DEBUG]" or filepath or entry.uri or entry.bufname
+  local filepath, cwd = entry.path, self.opts.cwd
+  local title = entry.debug and "[DEBUG]"
+      or filepath and (cwd and path.relative_to(filepath, cwd) or path.HOME_to_tilde(filepath))
```
ibhagwan: What if the cwd exists but some paths are outside the cwd (or maybe absolute paths), due to the “or” condition `HOME_to_tilde` will never be called no?

lua/fzf-lua/previewer/builtin.lua
phanen: I add some more field to entry since most of time user may just need to custom `:parse_entry`, so they don't need care about how to handle tmpbuf.

It happened when I'm try to override only `:parse_entry` (but not parse entry via super method since it's not a common path spec) and provide `opts.content`+`opts.filetype` only which is not used as cache key now. User can add a fake `opts.path` to avoid it but sometimes that's a real path on fs.

lua/fzf-lua/previewer/builtin.lua
phanen: right... i'll fix it.

phanen: should be `path.HOME_to_tilde(cwd and path.relative_to(filepath, cwd) or filepath)`

phanen:
fix some lint.

ibhagwan:
> fix some lint.

You feel this is ready to be merged?

I won’t be able to properly test and merge this in a couple of days as traveling, aiming to merge this Thursday

phanen:
Yeah, if it have issue i can fix it quickly.

ibhagwan:
> Yeah, if it have issue i can fix it quickly.

Alright I’ll merge it then, the start insert commit fixes https://github.com/ibhagwan/fzf-lua/issues/1659#issuecomment-3717427546.

lua/fzf-lua/previewer/builtin.lua
ibhagwan: I don’t see `path.HOME_to_tilde` in the changes anymore - did you find out it wasn’t necessary there?

phanen:
The latest 2 commit was to fix some window option issue. (just lazy to open another pr as usual...).

phanen:
preview window still use "global" option, which means it will be inherit by new split window, and switched buffer. since it has issue like https://github.com/ibhagwan/fzf-lua/discussions/2266.

Maybe this can also be fixed, but it's unusual to split a window from previewer, so let it as is.

ibhagwan:
Thank you @phanen for these great changes.

Regarding the commit https://github.com/ibhagwan/fzf-lua/commit/383e605bdd773edb8ff162003b3fbf7e976527ce, say we had a complex regex in live grep and there are some discrepencies between the rg regex and `vim.regex` the `reg:match_line` could fail in which case we lose the exact highlighting of the match, the fallback makes it so that if we know the column (from the rg output directly) we can at least highlight one character where the match should begin.

For example say we search for `fzf-lua` and the result is in `init.lua:1:10` (line 1 col 10) and for the sake of this example assume the regex match fails. we now highlight the first letter `[f]zf-lua` (instead of the entire string) so we can focus where the match begins.

This is also somewhat of a legacy since this is how live grep highlighting worked before the regex matching: cursor line to focus on the correct line, cursor char to focus on the right column.

PR #2510, opened 2026-01-11
feat: history picker alias to oldfiles with buffers

### LINKED ISSUE #2509: [Feature Request] adjust oldfiles picker default options to include history
History makes sense. I just tested `fzf-vim` which has a [`:History command`](https://github.com/junegunn/fzf.vim/#commands) (`v:oldfiles` and open buffers), and it indeed includes the current buffer.

I suppose potential action items could be:
- Change `include_current_session` to include the current buffer (aligns with fzf-vim)
- Add a new `include_current_buffer` config option (confusing to me since it would require `include_current_session` to also be enabled)
- Introduce a new `history` picker (way more work)

_Originally posted by @ezequiel in https://github.com/ibhagwan/fzf-lua/discussions/2508#discussioncomment-15465636_

ibhagwan:
closes #2509

aliases to:
```lua
:FzfLua oldfiles include_current_sesssion=true ignore_current_buffer=false
```

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -808,6 +808,10 @@ Open buffers in tabs
 
 File history (output of `:oldfiles`)
 
+#### history
+
+Combined current buffer + file history (output of `:oldfiles`)
+
 #### quickfix
 
 Quickfix list (output of `:copen`)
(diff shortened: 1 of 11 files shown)


ibhagwan:
whoops found one issue, when just startign nvim the first item will be set to header even if the current buffer doesn't exist in the list, will fix tomorrow.

ibhagwan:
Fixed, only issue remaining is dynamically setting the header line on reload which requires a `change-header-lines` action upstream: https://github.com/junegunn/fzf/issues/4659

PR #2513, opened 2026-01-13
fix(compat): nvim_win_get_config don't return full info when <0.10

phanen:


ibhagwan:
Not sure why this fails on windows with nightly, running the failed job third time now let’s see

ibhagwan:
Guess it was temporary

ibhagwan:
Ty @phanen !

Issue #254, opened 2021-12-13
`cursorline` gets reset when opening files


shyun3:
I have `cursorline` set in my `init.vim`. However, when I open a new file using **fzf-lua**, its uses `nocursorline`.

## Reproduction

1. First, here's the `init.vim` I used for this test:
    ```vim
    let g:vimDir = stdpath('config')

    call plug#begin(g:vimDir . '/bundle')
    Plug 'ibhagwan/fzf-lua'
    call plug#end()

    set cursorline
    ```
2. Create test files:
    ```sh
    echo hello > b
    echo world > a
    ```

3. Open `b`: `nvim b`

    Make sure `cursorline` is set and visible.

4. Run `:FzfLua files`
5. Highlight `a` (this should be the default) and open a new split by hitting **CTRL+v**
6. Note that the window containing the file `a` does not have a cursor line. Confirm by querying `cursorline`:
    ```
    :set cursorline?
    nocursorline
    ```

## Environment
* Neovim v0.6.0
* Ubuntu 20.04.3 LTS
* Windows 10 WSL2

ibhagwan:
Few things to try so we can narrow this down:

1. Can you try also calling the `setup` function from `init.vim` by adding the below?
```lua
lua << EOF
require('fzf-lua').setup{}
EOF
```

2. I suspect this might be related to the builting previewer not restoring options properly (althought weird since it also uses `cursorline`), can you try with the `bat` (or `cat` if you don't have `bat` installed)?
```lua
lua << EOF
require('fzf-lua').setup{
  winopts = {
    preview = { default = 'bat' }
  }
}
EOF
```

3. We can also try with the default previewer with `cursorline` removed from `winopts`:
```lua
lua << EOF
require('fzf-lua').setup{
  winopts = {
    preview = {
      winopts = {
        number            = true,
        relativenumber    = false,
        -- cursorline        = true,
        -- cursorlineopt     = 'both',
        cursorcolumn      = false,
        signcolumn        = 'no',
        list              = false,
        foldenable        = false,
        foldmethod        = 'manual',
      },
    }
  }
}
EOF
```

shyun3:
I tried all 3 and they didn't make a difference ☹

ibhagwan:
> I tried all 3 and they didn't make a difference frowning_face

Don't worry I promise you we'll get to the bottom of this :-)

[...]

PR #2529, opened 2026-01-21
fix: don't set cursorline in terminal buffer

phanen:
I shouldn't ignore previous comment (:


phanen:
maybe related: 
https://github.com/ibhagwan/fzf-lua/issues/254
https://github.com/junegunn/fzf.vim/issues/1400

another workaround is `set nonumber cursorlineopt=number`

PR #2534, opened 2026-01-25
fix: allow TermOpen to detect fzf filetype

phanen:
fzf.vim also defer filetype
https://github.com/junegunn/fzf/blob/821b8e70a80e3ea96772a4dda36311b19b7171b4/plugin/fzf.vim#L976

But maybe no reason we cannot set filetype before jobstart

https://github.com/ibhagwan/fzf-lua/discussions/2532#discussioncomment-15595064

### DIFF
--- a/lua/fzf-lua/fzf.lua
+++ b/lua/fzf-lua/fzf.lua
@@ -62,6 +62,8 @@ function M.raw_fzf(contents, fzf_cli_args, opts)
   table.insert(cmd, libuv.shellescape(outputtmpname))
 
   if not opts.is_fzf_tmux then
+    utils.eventignore(function() vim.bo.filetype = "fzf" end)
+
     -- A pesky bug I fixed upstream and was merged in 0.11/0.10.2:
     -- <C-c> in term buffers was making neovim freeze, as a workaround in older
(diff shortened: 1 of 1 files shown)


ibhagwan:
I don't mind this @phanen but do you feel this could have adverse affects? Some users before have used `FileType` events for term keymaps for example, setting the filetype beforehand would break that.

I have a feeling we shouldn't rush this change but give it some more thought, perhaps there's a reason why fzf.vim also defers the filetype?

phanen:
Emm.. maybe it can have some issue with `au FileType fzf`.
If you assume the buffer is terminal buffer there. but should no issue with keymap.

phanen:
https://github.com/search?q='pattern%20%3D%20"fzf"'&type=code
https://github.com/search?q=%27pattern%20%3D%20%22fzf%22%27&type=code

We can set FileType but with event ignore first, then fire the filetype again.

ibhagwan:
> We can set FileType but with event ignore first, then fire the filetype again.

That's not a bad idea

ibhagwan:
Tested, works great, `FileType` is called only once in the exact same timing as before.

PR #2537, opened 2026-01-26
refactor(actions): simplify vimcmd_entry

phanen:
Split out a few function.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -1364,7 +1364,10 @@ local parse_entry = function(e) return e and e:match("%((.-)%)") or nil end
 M.serverlist_kill = function(sel)
   vim.iter(sel):map(parse_entry):each(function(addr)
     local ok, err = utils.rpcexec(addr, "nvim_exec2", "qa!", {})
-    assert(ok or tostring(err):match("Invalid channel"), err)
+    assert(ok
+      or tostring(err):match("Invalid channel")
+      or tostring(err):match("ch %d+ was closed by the peer"),
+      err)
   end)
 end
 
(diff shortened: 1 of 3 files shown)


ibhagwan:
@phanen, what are we gaining by this refactor? I don't see too much code dedup or reuse it's mostly a few func splits - do you feel this is really necessary?

phanen:
I did this when I borrow some utils here to my config. I modified in my config then copy back here. If you think it cause more confusing feel free to close.

I just understand the code by rewriting.

phanen:
Before I to simply used `vim.cmd.edit`, but I found it can trigger FileType event. Then I use `vim.cmd.buffer` first but I found this is pattern match not exact match.

Finally I decide to use fzf-lua actions in my config so I don't need to consider how to do this correctly again.

neovim really should have something like https://github.com/oantolin/embark. picker/outline/bqf reinvent their own, and have to avoid bugs separatly.

ibhagwan:
> Before I to simply used `vim.cmd.edit`, but I found it can trigger FileType event. Then I use `vim.cmd.buffer` first but I found this is pattern match not exact match.
> 
> Finally I decide to use fzf-lua actions in my config so I don't need to consider how to do this correctly again.
> 
> neovim really should have something like https://github.com/oantolin/embark. picker/outline/bqf reinvent their own, and have to avoid bugs separatly.

I see, I’ll do some testing and merge tomorrow

phanen:
avoid some unnecessary diff

phanen:
Fix my mistaken fix before on serverlist. (cannot vim.schedule quit, because it cause other issue with reload)

ibhagwan:
Changed https://github.com/ibhagwan/fzf-lua/pull/2537/changes/52a1808ca6c62f3b0a19ea74a6b739899acea007 to include both error messages, otherwise it errors on 0.11.5.

ibhagwan:
I love the serverlsit @phanen, mapped it to `ctrl-s` in zsh and `ctrl-z` in neovim and now using it instead of "send to background", it's really nice, like an extra tmux just for neovim.

There are some issues where the preview screenshot doesn't work, still unable to have exact repro conditions, will keep you posted.

Issue #2547, opened 2026-02-01
Bug: Multiple LSP causes 'server does not support incomingCalls' error


andrewbraxton:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

MacOS

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-2162+g1906da52db

### Fzf version (`fzf --version`)

0.67.0 (Homebrew)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua

```

### Describe the bug / steps to reproduce

for Lua files I have 2 LSP servers: stylua and emmylua.

Emmylua supports call hierarchy, stylua does not.

If I run incoming_calls(), I get an error message that 'server does not support incomingCalls'. This is because this call is making the request to all servers

https://github.com/ibhagwan/fzf-lua/blob/b4b1f05957a77f7662de834328827890c1a8c02b/lua/fzf-lua/providers/lsp.lua#L469-L470

and this loop is printing an error because stylua's response is an error

https://github.com/ibhagwan/fzf-lua/blob/b4b1f05957a77f7662de834328827890c1a8c02b/lua/fzf-lua/providers/lsp.lua#L483-L489

but emmylua's response is fine, so the FZF window appears after the error with correct results.

phanen:
`client:supports_method` (called in lsp.buf_request) always return true for incomingCalls:
```
:=(vim.lsp.get_clients({ bufnr = 0, name = 'copilot' })[1]):supports_method("callHierarchy/incomingCalls")
```

Maybe its capability is not documented here, and `client:support_methods` always return true for "off-spec" method.
https://github.com/microsoft/language-server-protocol/blob/f1b5515dd96741041a68fa83da13c4e9fe0c1727/_specifications/lsp/3.18/metaModel/metaModel.json#L421-L454

phanen:
Maybe we can change error to warn: https://github.com/ibhagwan/fzf-lua/pull/2550.

[...]

PR #2550, opened 2026-02-03
fix(lsp): change error to warn on lsp error message (#2547)

phanen:
see https://github.com/ibhagwan/fzf-lua/issues/2547#issuecomment-3840420481

Builtin lsp warn on hierarchy request:
https://github.com/neovim/neovim/blob/0501c5fd0959895b18d8166c6c034cdf1832edf3/runtime/lua/vim/lsp/buf.lua#L1275
Also other "content" request don't throw error message.

So I think we can change error to warn.

Fix https://github.com/ibhagwan/fzf-lua/issues/2547.


ibhagwan:
Ty @phanen, though I still believe users will complain about the warning too

PR #2551, opened 2026-02-03
fix: cancel previous entry if we are previewing different entry

phanen:
The execute order of parse_entry callback can be random when it's slow
(e.g. callback in network request)

self:debounce don't work here, because we run a async parse_entry
callback again in the debounce timer callback


phanen:
Not sure if we have a way to cancel a coroutine. Maybe not possible in lua 5.1.
EDIT: oK it's not possible, can only be done with some framework https://github.com/folke/snacks.nvim/blob/01f6cac48fd7a3ec9bf7e5fc8a5ae22381861baf/lua/snacks/picker/util/async.lua#L66-L68, https://github.com/lewis6991/async.nvim/blob/0f0949cafac197c9b8f2c486fb5eb5865c0abfc3/lua/async.lua#L719-L721

ibhagwan:
Ty @phanen ”

PR #2552, opened 2026-02-05
fix(preview): copy all extmarks

phanen:
see picture in https://github.com/kevinhwang91/nvim-bqf/pull/175#issue-3900401828


ibhagwan:
Ty @phanen!

PR #2557, opened 2026-02-07
fix(cli): don't set lnum/col when it's invalid

phanen:
this break `:h restore-cursor`

### DIFF
--- a/lua/fzf-lua/previewer/fzf.lua
+++ b/lua/fzf-lua/previewer/fzf.lua
@@ -537,25 +537,23 @@ local function make_screenshot(screenshot, addr, lines, columns)
       end)
     ]], { lines, columns, screenshot })
   end
-  local jobstart = _G.fzf_pty_spawn or vim.fn.jobstart
   local nvim_bin = os.getenv("FZF_LUA_NVIM_BIN") or vim.v.progpath
-  local chan = jobstart({ nvim_bin, "--server", addr, "--remote-ui" }, {
+  local chan = vim.fn.jobstart({ nvim_bin, "--server", addr, "--remote-ui" }, {
     pty = true,
     height = lines,
     width = columns,
     env = {
       TERM = "xterm-256color",
-      VIMRUNTIME = vim.env.VIMRUNTIME,
+      VIMRUNTIME = os.getenv("VIMRUNTIME"),
     },
     on_stdout = function(chan0)
       if closing then return end
       closing = true
       vim.defer_fn(function() utils.rpcexec(addr, "nvim__screenshot", screenshot) end, 10)
-      if not _G.fzf_pty_spawn then vim.defer_fn(function() vim.fn.jobstop(chan0) end, 20) end
+      vim.defer_fn(function() vim.fn.jobstop(chan0) end, 20)
     end,
   })
-  if not _G.fzf_pty_spawn then return vim.fn.jobpid(chan) end
-  return chan
+  return vim.fn.jobpid(chan)
 end
 
 function Previewer.nvim_server:cmdline(_)


ibhagwan:
> this break :h restore-cursor

Not sure what you mean as I have a similar autocmd in my config but I trust you on this :)

was just about to merge this and you pushed, do you need more time?

phanen:
> Not sure what you mean as I have a similar autocmd in my config but I trust you on this :)

If open nvim with `nvim README.md +0` restore last jump pos don't seem work for me.



> was just about to merge this and you pushed, do you need more time?

No... I want to test a bit if we use `nvim_open_term` for preview.

ibhagwan:
> If open nvim with nvim README.md +0 restore last jump pos don't seem work for me.

Oh you’re right, if I open directly from the cli it won’t restore cursor pos, amazing fix, I thought this wasn’t possible with the cli, just tested this PR and it works for me too :)

> No... I want to test a bit if we use nvim_open_term for preview.

alright let me know when and I’ll merge

phanen:
image

Just tested, it seem buggy...

ibhagwan:
> Just tested, it seem buggy...

Seems like an issue with ANSI escape sequences for colors.

btw, I love the serverlist now, I modified my ctrl-z bind in neovim to `:detach` and it works amazing.

ibhagwan:
Good to merge?

phanen:
> btw, I love the serverlist now, I modified my ctrl-z bind in neovim to `:detach` and it works amazing.

Yeah. that's a good idea, detach seems always better than suspend.

suspend have a issue since they can also be detected in serverlist (and if you connect a new ui then quit the server, its old tui will still be suspend)

> Good to merge?

Y

ibhagwan:
@phanen, I'm still unsure why but https://github.com/ibhagwan/fzf-lua/pull/2557/changes/d05a83c3c16ff5c952da88d4b907ccebd9c8f160 makes it so `:FzLua serverlist` enter doesn't work anymore, investigating.

ibhagwan:
Reverted https://github.com/ibhagwan/fzf-lua/commit/a3508e89f1662bd8d57ece727343fce4202bdf38 for now, it doesn't work on 0.11.6 (enter does nothing), still unsure why as this commit only affects the preview?

phanen:
Oh, this is issue I forgot... coredumpctl show it sigabort.

phanen:
This is upstream bug. Only happen on cli.
```
(gdb) bt
#0  0x00007fc31936090c in ?? () from /usr/lib/libc.so.6
#1  0x00007fc3193063a0 in raise () from /usr/lib/libc.so.6
#2  0x00007fc3192ed57a in abort () from /usr/lib/libc.so.6
#3  0x00007fc3194f02b1 in ?? () from /usr/lib/libuv.so.1
#4  0x00007fc31950ffbe in ?? () from /usr/lib/libuv.so.1
#5  0x00007fc3194f75a2 in uv_run () from /usr/lib/libuv.so.1
#6  0x0000562438f36005 in loop_uv_run (loop=<optimized out>, ms=1) at /usr/src/debug/neovim-git/neovim/src/nvim/event/loop.c:59
#7  loop_poll_events (loop=<optimized out>, ms=1) at /usr/src/debug/neovim-git/neovim/src/nvim/event/loop.c:80
#8  tui_start (tui_p=0x562439126938 <tui.lto_priv>, width=0x56243912fd64 <tui_width.lto_priv>, height=0x56243912fd60 <tui_height.lto_priv>, term=0x562439113880 <tui_term>, 
    rgb=0x562439130b58 <tui_rgb>) at /usr/src/debug/neovim-git/neovim/src/nvim/tui/tui.c:184
#9  0x0000562438f49f25 in ui_client_run (remote_ui=<optimized out>) at /usr/src/debug/neovim-git/neovim/src/nvim/ui_client.c:164
#10 0x0000562438b317f3 in main (argc=8, argv=<optimized out>) at /usr/src/debug/neovim-git/neovim/src/nvim/main.c:351
```

PR #2561, opened 2026-02-09
fix(actions): ensure requested buffer is listed

danielwe:
Since #2552, specifically 6971762, it's possible for `entry.bufnr` to be populated by an unlisted buffer,[^1] which would remain unlisted even as it was set as the current buffer, because any non-nil `bufnr` means `load_buf` is skipped (see line 249).

This PR moves `vim.bo[bufnr].buflisted = true` from `load_buf` to `set_buf`, such that it applies to any buffer that becomes current through this codepath.

Not sure if/how to write a test for this. Let me know!

[^1]: For me, this frequently happens because the language client opens almost the entire workspace in unlisted buffers as context for the language server's static analysis. When I subsequently use fzf-lua to open one of these files, my buffer line and buffer navigation keymaps break because they only work on listed buffers.

### DIFF
--- a/tests/actions_spec.lua
+++ b/tests/actions_spec.lua
@@ -155,5 +155,19 @@ T["actions"]["vimcmd"] = new_set({
   end
 })
 
+T["actions"]["buflisted #2561"] = function()
+  child.cmd("edit LICENSE")
+  child.bo.buflisted = false
+  child.cmd("enew")
+  helpers.FzfLua.files(child, {
+    __abort_key = "<cr>",
+    __expect_lines = false,
+    __after_open = function()
+      if helpers.IS_WIN() then vim.uv.sleep(250) end
+    end,
+    query = "LICENSE$",
+  })
+  eq(true, child.bo.buflisted)
+end
 
 return T


ibhagwan:
Ty @danielwe!

Let’s ask @phanen, the author of said commit

phanen:
Yes. This should resume to previous behavior.
I didn't notice much different, maybe cause I ~only use tabpage~, don't use `":bnext"` or bufferline.

phanen:
Here's a test, can be put in actions_spec
```lua
T["actions"]["buflisted #2561"] = function()
  child.cmd("edit LICENSE")
  child.bo.buflisted = false
  child.cmd("enew")
  helpers.FzfLua.files(child, {
    __abort_key = "<cr>",
    __expect_lines = false,
    __after_open = function()
      if helpers.IS_WIN() then vim.uv.sleep(250) end
    end,
    query = "LICENSE$",
  })
  eq(true, child.bo.buflisted)
end
```

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2561/commits/f9c40b4c26c222434a7d671d71c0a48bf071407a, good to merge?

phanen:
I'm not sure if there's other case: `entry.bufnr` is `unlisted` but we want keep it unlisted even action "open" it.

phanen:
`Buffers` picker have some options to show unlisted buffer

https://github.com/ibhagwan/fzf-lua/blob/95f6baea73d0440a28c55d370d424ac645b447a0/lua/fzf-lua/providers/buffers.lua#L51-L57

ibhagwan:
> I'm not sure if there's other case: entry.bufnr is unlisted but we want keep it unlisted even action "open" it.

Technically I would expect fzf-lua to never change the buffer options, including visibility, that’s why we have show_unlisted in buffers.

phanen:
> never change the buffer options

Both telescope/snacks change it:
* https://github.com/nvim-telescope/telescope.nvim/blob/bea6665a8f14f31e11a3093a3b4a92be313bf4be/lua/telescope/actions/set.lua#L190-L194
* https://github.com/folke/snacks.nvim/blob/d902c0a415ffbf66321f40ecb07e73fea283d0ab/lua/snacks/picker/actions.lua#L95-L100

... currently I cannot figure out a reason why don't buflist it when "open".

But to strictly resume previous behavior, "parse buf form path" logic should be put here:
https://github.com/ibhagwan/fzf-lua/blob/6b56330beeea6efcb8f832061ec4af12294d8c23/lua/fzf-lua/previewer/builtin.lua#L675-L677

ibhagwan:
I don’t recall why I added buflisted=true in the original code (before https://github.com/ibhagwan/fzf-lua/commit/46d8e5cea38e3373cdefc6993519f366a524f2be), there had to have been a good reason,  and the use case of OP does make sense, when you specifically edit a file you’d expect it to become visible (from the quick fix list, args, etc).

ibhagwan:
> ... currently I cannot figure out a reason why don't buflist it when "open".

I agree.

> But to strictly resume previous behavior, "parse buf form path" logic should be put here:

But do we want to restore it? Isn’t it better (and faster) to reuse the bufnr since we already have it?

phanen:
> But do we want to restore it? Isn’t it better (and faster) to reuse the bufnr since we already have it?

Yes... reuse is fine.

The problem is when should we keep buflisted. And when should not. 

Since we don't know when should ... merge it then let user report (:

ibhagwan:
> The problem is when should we keep buflisted. And when should not.
> Since we don't know when should ... merge it then let user report (:

I think it should be fine, I changed my mind about the “never”,  if the user opened a buffer let it be listed.

The only difference now is that before it was only for buffers that fzf-lua loads and now it’s all the buffers we “edit”.

danielwe:
Thank you!

If you're still unsure about this, perhaps a slightly more conservative option would be to copy the listed-ness of the buffer being replaced instead of unconditionally using `true`:

```diff
diff --git i/lua/fzf-lua/actions.lua w/lua/fzf-lua/actions.lua
index 6139f488..8ae9a356 100644
--- i/lua/fzf-lua/actions.lua
+++ w/lua/fzf-lua/actions.lua
@@ -181,6 +181,7 @@ local set_buf = function(bufnr, will_replace_curbuf)
   then
     vim.bo.bufhidden = "wipe"
   end
+  local listed = vim.bo.buflisted
   -- NOTE: nvim_set_current_buf will load the buffer if needed
   -- calling bufload will mess up `BufReadPost` autocmds
   -- vim.fn.bufload(bufnr)
@@ -190,7 +191,7 @@ local set_buf = function(bufnr, will_replace_curbuf)
   -- user cancelles the save dialog pcall will fail with:
   -- Vim:E37: No write since last change (add ! to override)
   if not ok then return end
-  vim.bo[bufnr].buflisted = true
+  vim.bo[bufnr].buflisted = listed
   return true
 end
 ```

Alternatively, the old behavior can be restored by reverting this PR and only populating `entry.bufnr` if the buffer is listed (but I completely lack context, maybe this would break other parts of #2552):

```diff
diff --git i/lua/fzf-lua/path.lua w/lua/fzf-lua/path.lua
index 10155238..3e5ec76b 100644
--- i/lua/fzf-lua/path.lua
+++ w/lua/fzf-lua/path.lua
@@ -537,7 +537,7 @@ function M.entry_to_file(entry, opts, force_uri)
     end
   elseif file and #file > 0 then -- get bufnr from give path
     local buf = vim.fn.bufnr(file)
-    if buf ~= -1 and vim.api.nvim_buf_get_name(buf) == file then
+    if buf ~= -1 and vim.bo[buf].buflisted and vim.api.nvim_buf_get_name(buf) == file then
       bufnr = buf
       bufname = file
     end
```

Just throwing those out there in case it's helpful. Thanks for a fantastic plugin!

ibhagwan:
@danielwe, I think we should be good, if a user complains I’ll forward them to you :)

Ty for the kind words :heart:

PR #2562, opened 2026-02-09
feat(default): allow `lua$` to match lua buffer only in buffer picker

phanen:
Maybe someone can have different muscle memory like `lua:` rather than `lua$`, I just always find I type something like `lua$` and found it don't work in buffer picker.


phanen:
Also need `["--delimiter"] = ":"`. I forgot we have `opts.line_query`. Maybe leave this unchanged.

PR #2565, opened 2026-02-15
add "none" previewer like "bat" but no external dependencies

phanen:
- **feat(preview): colored previewer**
- **feat(preview): add "none" like "bat" but don't require dependencies**

I don't know why we don't use builtin preview....

But it's interesthing to know how it work.

### DIFF
--- a/.emmyrc.json
+++ b/.emmyrc.json
@@ -12,10 +12,12 @@
     "disable": [
       "unnecessary-if",
       "await-in-sync",
-      "global-in-non-module"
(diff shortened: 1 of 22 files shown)


phanen:
Maybe more hl/ansi related stuffs can be moved to hl.lua.

Also this picker is very slow, better to speed up with `stringify_mt`...

phanen:
Thanks I don't notice you've rebased it for me. I think this can be closed currectly since it's not that useful and revisited when needed

Issue #2567, opened 2026-02-16
Bug: fzf-native profile ⇒ Invalid border style 'border-line', will use 'rounded'


liskin:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.11.6

### Fzf version (`fzf --version`)

0.67.0 (debian)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

n/a

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
'fzf-native'
})
```

### Describe the bug / steps to reproduce

1. `:FzfLua keymaps`
2. get the following error a number of times: `[Fzf-lua] Invalid border style 'border-line', will use 'rounded'.`

Bisected to 4d66710a6e869cd4a72f1013814fb2cbfaf35c7d.
(I mean, I had to skip that one as it was failing an assertion, so technically the first bad commit is 5eed08208ed32b841b97918c698efa21c0e68978, but it's fairly obvious it's 4d66710a6e869cd4a72f1013814fb2cbfaf35c7d that introduced the problem.)

liskin:
This works (and is visually correct):

```lua
require('fzf-lua').setup({
	'fzf-native',
	winopts = {
		preview = {
			border = require'fzf-lua.profiles.border-fused'.winopts.preview.border,
		},
	},
})
```

Perhaps the whole https://github.com/ibhagwan/fzf-lua/blob/main/lua/fzf-lua/profiles/fzf-native.lua#L7-L15 is unnecessary?

nenahp:
`FzfLua keymaps` only have builtin previewer, so this border type must to be converted into nvim border type

```diff
diff --git a/lua/fzf-lua/profiles/fzf-native.lua b/lua/fzf-lua/profiles/fzf-native.lua
index 568aa530..be2cb0f0 100644
--- a/lua/fzf-lua/profiles/fzf-native.lua
+++ b/lua/fzf-lua/profiles/fzf-native.lua
[...]

PR #2568, opened 2026-02-17
fix(preview): convert border-line to border-fused (#2567)

phanen:
Fix https://github.com/ibhagwan/fzf-lua/issues/2567.


ibhagwan:
Ty @phanen!

PR #2570, opened 2026-02-18
fix(path): resolve symlink when compare bufname

phanen:
nvim_buf_get_name/bufname return a resolved version


ibhagwan:
Ty @phanen !

PR #2571, opened 2026-02-18
fix(skim): skim 0.12+ support reload

phanen:


ibhagwan:
Wow how did I miss this commit :)

Tested, works, not as good as fzf (cursor doesn’t stay in place), upstream issue, ty @phanen!

PR #2572, opened 2026-02-18
fix(manpages): warn when man executable or mandb cache is missing

barrettruth:
basically i realized that `man -k .` was not working on nixos (the command for `:FzfLua man_pages` and that i was missing mandb.

So, thought it would be nice to make sure that this doesn't happen to anyone else by warning in the picker :D


barrettruth:
:sob: @ibhagwan thanks

PR #2576, opened 2026-02-21
Ability to filter treesitter nodes

kozorork:
It seems like the treesitter function can only be used to fuzzy-search all symbols. Maybe searching only through specific nodes could be useful (for example to find a function and jump to it)


ibhagwan:
ty @kozorork!

Issue #2587, opened 2026-02-27
Bug: inconsistent background color when toggling ignored files with Alt-I in grep modes


shmerl:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux (Debian testing)

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-2425+gbe8969f4cc

### Fzf version (`fzf --version`)

0.67.0 (debian)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').register_ui_select() -- use fzf-lua for vim.ui.select
```

### Describe the bug / steps to reproduce

Pressing Alt-I to toggle ignored files changes background color of the left fzf-lua pane

Example (using retrobox color theme for more visible contrast):

Opening fzf-lua for the first time:

```
:FzfLua grep_curbuf
```

<img width="1783" height="344" alt="Image" src="https://github.com/user-attachments/assets/b7945cc3-68ad-4024-91eb-25ff142fa793" />

Pressing Alt-I changes background color in the left pane:

<img width="1783" height="344" alt="Image" src="https://github.com/user-attachments/assets/4a17324d-5737-4585-964e-5ea62486ce9d" />

phanen:
b8d1356b7120f5cf111709cb2fa504096248fb26.

phanen:
Should be fixed in https://github.com/ibhagwan/fzf-lua/pull/2588

shmerl:
Thank you!

PR #2588, opened 2026-02-27
fix(win): force update fzf winhl when reuse window (#2587)

phanen:
When reuse window (cmd reload or switch to new picker),
fzf job buffer will change, so have to reapply window-local option again

Fix regression from
https://github.com/ibhagwan/fzf-lua/commit/b8d1356b7120f5cf111709cb2fa504096248fb26


ibhagwan:
tysm @phanen !

PR #2597, opened 2026-03-03
feat: add nonicons.nvim icon provider

barrettruth:
Add icons support for my nonicons plugin [`barrettruth/nonicons.nvim`](https://github.com/barrettruth/nonicons.nvim) as a third icon set

also, add some icon tests


ibhagwan:
Tysm @barrettruth, but ATM I'd rather not add yet another obscure icon provider, if nonicons becomes more prevalent we can revisit this and perhaps refactor and add code that would make adding an icon provider generic via "register" or something similar.

ibhagwan:
Closing as not merging ATM.

Issue #2598, opened 2026-03-03
Bug: Invalid window id error when opening file


mikehaertl:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux Mint 22.3

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.11.6

### Fzf version (`fzf --version`)

0.62.0 (d226d841)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

--color fg:7,bg:0,hl:13,fg+:2,bg+:6,hl+:13                                                                                                                                                                          --color query:13,info:12,prompt:13,spinner:3,marker:9                                                                                                                                                              --cycle --bind=change:first,ctrl-n:page-down,ctrl-p:page-up,ctrl-t:toggle-preview,ctrl-alt-j:preview-down,ctrl-alt-k:preview-up,ctrl-alt-n:preview-page-down,ctrl-alt-p:preview-page-up,tab:accept,ctrl-o:toggle+d own,ctrl-alt-a:toggle-all

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
local actions = require "fzf-lua.actions"
local fzflua = require('fzf-lua')
fzflua.setup({
  'fzf-native',
  winopts = {
    backdrop = false,
    split = "botright new",
    preview = {
      horizontal = 'right:50%',
    },
  },
  actions = {
    files = {
      -- providers that inherit these actions:
      --   files, git_files, git_status, grep, lsp
      --   oldfiles, quickfix, loclist, tags, btags, args
      ["default"]     = actions.file_edit,
      ["ctrl-x"]      = actions.file_split,
      ["ctrl-v"]      = actions.file_vsplit,
      ["alt-q"]       = actions.file_sel_to_qf,
      ["alt-l"]       = actions.file_sel_to_ll,
      ["alt-i"]       = actions.toggle_ignore,
    },
    buffers = {
      -- providers that inherit these actions:
      --   buffers, tabs, lines, blines
      ["default"]     = actions.buf_edit,
      ["ctrl-x"]      = actions.buf_split,
[...]

PR #2599, opened 2026-03-03
fix(win): abort when winobj is closing with sigwinch at the same time

phanen:
Maybe fix https://github.com/ibhagwan/fzf-lua/issues/2598


mikehaertl:
Does not fix the issue for me. But I found what is causing it, see my comment in #2598

phanen:
Now statusline can be overriden
```
augroup fzf
  autocmd!
  autocmd  FileType fzf setlocal statusline=
augroup END
```

PR #2600, opened 2026-03-03
feat(cli): nvim bin/runtime from custom path

ibhagwan:


ibhagwan:
@phanen, did this so I can use neovim nightly compiled from source which is then aliased in the shell to:
```sh
vi='VIMRUNTIME=/home/.../nvim/neovim/runtime /home/bhagwan/.../neovim/build/bin/nvim'
```

Without this I was having issues with `serverlist` and `ctrl-q` due to my installed version being 0.11.6 with different runtimes.

Now it works nicely with:
```sh
    export FZF_LUA_NVIM_BIN=/home/bhagwan/.../neovim/build/bin/nvim
    export FZF_LUA_NVIM_RUNTIME=/home/.../nvim/neovim/runtime
```

Also modified the use of `vim.fn.exepath` to `vim.v.progpath` which is better in this case as it uses the same `nvim` binary that started the process (if you're using `:FzfLua serverlist` it would be an issue due to it finding the distribution nvim and not the custom).

PR #2602, opened 2026-03-06
fix(previewer): always call base:close

ibhagwan:
@phanen, any comments?

Regression from recent refactor `swiper.base:close()` was never called, not clearing the highlights/restoring cursor position.

I believe this is the right way to fix it post refactor but any comments are welcome.

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -780,8 +780,12 @@ function FzfWin:attach_previewer(previewer)
     if not self._previewer.preview_bufnr and self:validate_preview() then
       self._previewer.preview_bufnr = api.nvim_win_get_buf(self.preview_winid)
     end
-    self:close_preview()
+    if self.on_closes.preview then self.on_closes.preview() end
   end
+  -- This makes sure previewer.base:close is always called on :close
+  --   (1) Used by swiper/ivy/custom previewers
+  --   (2) Overwritten (extended) in builtin previewer (in :redraw_preview)
+  self.on_closes.preview = function(hide) self:close_preview(hide) end
   self._previewer = previewer
   self.previewer_is_builtin = previewer and previewer.type == "builtin"
   self.toggle_behavior = previewer and previewer.toggle_behavior or self.toggle_behavior


phanen:
"swiper" seems using original window/buffer as previewer.

Then maybe can move this to `attach_previewer`? If `on_close.preview` is not only about "close the preview window", but to dispose the "preview object"

ibhagwan:
> "swiper" seems using original window/buffer as previewer.
> 
> Then maybe can move this to `attach_previewer`? If `on_close.preview` is not only about "close the preview window", but to dispose the "preview object"

I agree attach is probably better, I was trying to do minimal change to the refactor, I can change it later when not AFK or if you want I think you can push onto this PR?

phanen:
Maybe tomorrow, unluckily I'm not on keyboard now.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2602/changes/432aa4400b666d73b8a698a1c24095ab800fcc42 - what do you think @phanen?

This is cleaner, makese sure this is called for every previewer obj regardless, gets overwritten for the builtin (for the extrmarks copy, etc)

lua/fzf-lua/win.lua
```diff
@@ -782,6 +782,10 @@ function FzfWin:attach_previewer(previewer)
     end
     self:close_preview()
```
phanen: Maybe can call `self.on_closes.preview` here, once redraw_preview, "on_closes.preview" will have copy_extmarks release cb in it.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2602/changes/1511d977ecae0aaf902d76c0c716162299c047a1, @phanen, if you agree I'll merge.

phanen:
Oh maybe nil check on_closes.preview

phanen:
maybe not needed validate_preview ensure we have previewer and its releaser

ibhagwan:
> maybe not needed validate_preview ensure we have previewer and its releaser

I thought about it but since we set it right below I figured it can't be nil, added nil check for good measures anyways.

PR #2613, opened 2026-03-12
fix(win): add EOL padding to compensate for path_shorten concealing (…

### LINKED ISSUE #2607: [BUG] winopts.path_shorten (using extmarks) shifts text
_Originally posted by @liskin in https://github.com/ibhagwan/fzf-lua/issues/2520#issuecomment-4024073295_

2026-03-09-140957_1127x754_scrot

ibhagwan:
…#2607)

When winopts.path_shorten uses ephemeral conceal extmarks to visually shorten directory components, the reduced display width causes fzf right- side UI elements (scrollbar, preview border) to shift left.

Add inline virtual text padding (spaces) at end-of-line to maintain the original visual line width. The padding width is computed using strdisplaywidth to correctly handle multibyte and CJK wide characters.

Refactor _apply_line into a pure _compute_line function (returns extmark specs) and a thin _apply_line wrapper (sets ephemeral extmarks), making the logic testable without ephemeral extmark limitations.

Also fix vim.str_utfindex call to use the 0.11+ single-return API and expose PathShortener via M._PathShortener for testing.

Fixes #2607

@phanen, created by Opus, needs more work

## Summary by CodeRabbit

* **Refactor**
  * Redesigned path-shortening display to use overlayed virtual text instead of per-line concealment, improving alignment and visual consistency across result formats and platforms.
  * Streamlined attach/detach lifecycle to carry per-window state without toggling conceal settings.

* **New Features**
  * Added formatter options for dirname/filename ordering and per-segment highlighting, plus ANSI color stripping for cleaner results.

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -7,11 +7,14 @@ local api = vim.api
 local fn = vim.fn
 
 ---@class fzf-lua.PathShortener
----@field _ns integer?
----@field _wins table<integer, { bufnr: integer, shorten_len: integer }>
+---@field _ns integer? namespace for ephemeral conceal extmarks
+---@field _ns_padding integer? namespace for persistent inline VT padding
+---@field _wins table<integer, fzf-lua.PathShortener.WinData>
+---@field _bufs table<integer, boolean> buffers with active on_lines attachment
 local PathShortener = {}
 
 PathShortener._wins = {}
+PathShortener._bufs = {}
 
 -- Hoist constant byte values for performance
 local DOT_BYTE = path.dot_byte
@@ -19,8 +22,12 @@ local DOT_BYTE = path.dot_byte
 function PathShortener.setup()
   if PathShortener._ns then return end
   PathShortener._ns = api.nvim_create_namespace("fzf-lua.win.path_shorten")
+  PathShortener._ns_padding = api.nvim_create_namespace("fzf-lua.win.path_shorten.padding")
 
-  -- Register decoration provider with ephemeral extmarks
+  -- Register decoration provider with ephemeral conceal extmarks.
+  -- Conceal preserves all terminal highlights (fzf match highlighting,
+  -- cursor line bg, formatter ANSI colors) unlike overlay VT which
+  -- replaces the underlying terminal cell content entirely.
   api.nvim_set_decoration_provider(PathShortener._ns, {
     on_win = function(_, winid, bufnr, _topline, _botline)
       -- Only process registered fzf windows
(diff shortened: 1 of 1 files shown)


ibhagwan:
@phanen, tried all kinds of approaches:
- conceal + inline padding
- conceal + overlay padding
- replaced with an overlay (this code, semi worked, overly complicated/hacky and still broken highlights with formatters)


Not sure if this has a solution, ideas welcome

ibhagwan:
## Detailed Summary: Approaches Tried for `path_shorten` Width Compensation

After extensive investigation, here's a complete breakdown of every approach attempted, why each failed, and the fundamental neovim limitation that prevents a complete solution for native/bat/delta previewers.

### Background

When `winopts.path_shorten` uses ephemeral conceal extmarks to visually shorten directory components (e.g., `lua/fzf-lua/win.lua` → `l/f/win.lua`), the reduced display width causes fzf's right-side UI elements (scrollbar, preview border) to shift left, creating visual misalignment.

The goal is to compensate for that lost display width while preserving all terminal highlights (ANSI colors, fzf match highlighting, cursor line background, formatter colors like `dirname_first`).

---

### Approach 1: Conceal + Ephemeral Inline VT Padding

**Idea:** Use ephemeral conceal extmarks (via decoration provider) to shorten paths, then add ephemeral `virt_text_pos = "inline"` padding to compensate for the concealed width.

**Result: Failed.** Ephemeral `virt_text_pos = "inline"` does **not render in terminal buffers** at all. The `pcall` reports success (`ok=true`), the extmark is accepted, but the virtual text is simply invisible. This appears to be a neovim limitation where the terminal grid doesn't process ephemeral inline virtual text during rendering.

---

### Approach 2: Conceal + Non-ephemeral Inline VT Padding (from decoration provider)

**Idea:** Same as above, but use non-ephemeral inline VT extmarks set from the decoration provider's `on_line` callback.

**Result: Failed.** Setting non-ephemeral extmarks from decoration provider callbacks causes **infinite redraw loops**. The decoration provider fires during rendering → sets extmark → invalidates display → triggers re-render → decoration provider fires again → infinite loop.

---

### Approach 3: Overlay Virtual Text (replacing entire line content)

**Idea:** Instead of conceal, use `virt_text_pos = "overlay"` to replace the visible line content with a shortened version, padding to the full terminal width.

**Result: Partially worked but fundamentally broken.** Overlay VT replaces the underlying terminal cell content entirely. `hl_mode = "combine"` does **not** help because terminal ANSI colors are not neovim highlight groups — they are terminal cell attributes that exist in a completely separate rendering layer. This means:

- ❌ All fzf match highlighting (when you type a search query) is lost
- ❌ Cursor line background highlighting is lost
- ❌ Formatter colors (`dirname_first`, `filename_first` ANSI escapes) are lost
- ❌ All terminal ANSI colors on the line are replaced with whatever highlight group the overlay VT uses

This is the approach in the current PR code — it's overly complex/hacky and still produces broken highlights.

---

### Approach 4: Conceal + Non-ephemeral Inline VT Padding (from `nvim_buf_attach on_lines` + `vim.schedule`) ← Current Working Solution

**Idea:** Use ephemeral conceal extmarks via decoration provider (preserves all terminal highlights on non-concealed characters), combined with non-ephemeral `virt_text_pos = "inline"` padding set from `nvim_buf_attach`'s `on_lines` callback, deferred via `vim.schedule`.

**Result: Works for builtin previewer, but NOT for bat/native/delta previewers.**

Key discoveries along the way:
- Non-ephemeral inline VT **does** render in terminal buffers (unlike ephemeral)
- Setting extmarks directly from `on_lines` during terminal I/O causes **visual corruption** (terminal grid miscalculates line widths, breaking the floating window layout) — solved by wrapping in `vim.schedule` to defer out of the terminal I/O handler
- For **builtin previewer**: the `│` (U+2502) character in lines is fzf's **scrollbar thumb** (last character on the line). Padding is placed right before it, pushing it to the correct display column. This works perfectly — all highlights preserved, border aligned.
- For **bat/native/delta previewer**: the `│` character is fzf's **internal border** between the list pane and the preview pane, with preview content following it. Here inline VT padding **shifts the text** to the correct column, but **terminal cell highlights remain at their original byte-indexed positions**, causing syntax highlight misalignment in the preview pane (e.g., border `│` loses its distinct color, bat syntax highlighting shifts left relative to the text).

---

### Why Bat/Native/Delta Previewers Cannot Work

This is a **fundamental neovim limitation**: in terminal buffers, ANSI colors are stored as terminal cell attributes indexed by grid position, not by the text content. When inline virtual text is inserted, neovim shifts the text rightward to make room for the VT content, but the terminal cell highlight attributes **stay at their original grid positions**. The text moves; the colors don't.

This means for any line where fzf renders both the file list entry AND preview content in the same terminal buffer line (which is what bat/native/delta previewers do — the preview is inside fzf's own terminal, not a separate neovim window), adding inline VT padding will always cause highlight misalignment in the preview portion.

There is no workaround for this within the current neovim terminal rendering architecture. The only options are:

1. **Skip padding for bat/native/delta lines** — paths are shortened (conceal works), but the border shifts left (current implementation)
2. **Disable `path_shorten` entirely** when native previewer is detected
3. **File a neovim issue** requesting that inline VT in terminal buffers also shifts terminal cell attributes

### Current Implementation Status

The working solution (approach 4) is implemented in `win.lua` on the `fix/path_shorten_padding` branch with the following behavior:

| Previewer | Path Shortening | Highlights | Border/Scrollbar Alignment |
|-----------|----------------|------------|---------------------------|
| Builtin (default) | ✅ Works | ✅ All preserved | ✅ Aligned |
| Bat/Native/Delta | ✅ Works (conceal) | ✅ All preserved | ❌ Shifts left |

Detection logic: if `│` is the **last character** on the line → scrollbar (builtin previewer) → add padding. If `│` has **content after it** → internal fzf border (bat/native previewer) → skip padding to preserve highlight integrity.

All 226 test cases pass with `make test`.

ibhagwan:
@phanen, I’m giving up on this idea, the closest I got was the current state but even when the text is aligned the bat syntax highlighting is shifted, this doesn’t seem solvable:
image

phanen:
Related: https://github.com/neovim/neovim/issues/37447

PR #2619, opened 2026-03-15
ci: fix flakey tests

### LINKED ISSUE #2618: CI: fix flakey tests
### Have you RTFM'd?

### Feature Request

@phanen, we still have a lot of flakey tests which I then run manually.

"line 301" also still pops up sometimes, any ideas welcome :-)

https://github.com/ibhagwan/fzf-lua/actions/runs/23112125314/job/67131310077#step:7:92

ibhagwan:
Fix #2618

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer handling when external-process spawning isn't appropriate: non-command results now finalize cleanly and avoid starting processes; post-processing is scheduled more reliably across fast/normal paths. Added a fallback pump-wait to ensure event-loop work completes.

* **Tests**
  * Improved test reliability by waiting for previews and async loads to finish before assertions.

### DIFF
--- a/tests/actions_spec.lua
+++ b/tests/actions_spec.lua
@@ -102,6 +102,9 @@ T["actions"]["vimcmd"] = new_set({
   },
 }, {
   function(action)
+    -- store mini.test float so we can return to it when
+    -- testing in main instance
+    local win = vim.api.nvim_get_current_win()
     local ctx = function()
       return {
         buf = child.api.nvim_get_current_buf(),
@@ -149,6 +152,12 @@ T["actions"]["vimcmd"] = new_set({
       __expect_lines = true,
       __after_open = function()
         child.wait_until(function() return child.lua_get([[_G._fzf_load_called]]) == true end)
+        -- Wait for previewer to load the LICENSE file entry and title to render
+        child.wait_until(function()
+          local entry = child.lua_get([[FzfLua.utils.fzf_winobj()._previewer.last_entry]])
+          return entry and entry:match("LICENSE") ~= nil
+        end)
+        vim.uv.sleep(100)  -- Allow async title update to render
         if helpers.IS_WIN() then vim.uv.sleep(250) end
       end,
       no_esc = true,
@@ -160,6 +169,9 @@ T["actions"]["vimcmd"] = new_set({
       eq({ "LICENSE", 3 }, { _ctx.name, child.fn.line(".") })
     end
     eq("live_grep", exec_lua([[return _G._fzf_info.cmd]]))
+    -- close tab and return to mini.test float
+    vim.api.nvim_buf_delete(0, { force = true })
+    vim.api.nvim_set_current_win(win)
   end
 })
 
--- a/tests/api_spec.lua
+++ b/tests/api_spec.lua
@@ -108,6 +108,9 @@ T["api"]["fzf_live"]["table"] = function(multiprocess)
     {
       __expect_lines = true,
       __postprocess_wait = true,
+      __after_open = function()
+        child.wait_until(function() return child.lua_get([[_G._fzf_load_called]]) == true end)
+      end,
       query = 100,
       keymap = { fzf = { load = "pos(-1)" } },
       multiprocess = multiprocess == true,
--- a/tests/path_spec.lua
+++ b/tests/path_spec.lua
@@ -429,6 +429,7 @@ describe("Testing path module", function()
         stripped = "/tmp/test.txt:42",
         terminal = false
       })
+      vim.api.nvim_buf_delete(buf, { force = true })
     end)
 
     it("man://", function()
@@ -448,6 +449,7 @@ describe("Testing path module", function()
         stripped = "man://ls(1):42:7:",
         terminal = false
       })
+      vim.api.nvim_buf_delete(buf, { force = true })
     end)
 
     it("force_uri", function()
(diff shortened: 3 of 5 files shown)


ibhagwan:
https://github.com/ibhagwan/fzf-lua/pull/2619/changes/8febce40978faa12eadc6c6d710b4b32dff72a94

@phanen, what do you think? this is the reasoning of Claude, I'm not loving the `vim.wait` solution but perhaps the analysis is right and we can have a better solution?

**EDIT:** - analysis seems correct, need to check if this can have possible side effects with table/function commands in other scenarios

image

ibhagwan:
Analysis is accurate, put a `vim.uv.sleep` (better yet add `io.stdout:write`)  inside the `vim.schedule` and you'll see it consistently:
https://github.com/ibhagwan/fzf-lua/blob/bf59dcee2a8d894724b060401d2abf3d71d4bd47/lua/fzf-lua/libuv.lua#L547-L552

I'm trying to think is there every a case where we need to send table/function cmd contents to spawn here (doesn't seem so but mayeb I'm missing something)?
https://github.com/ibhagwan/fzf-lua/blob/4a7a5f86ba4e6a6a42eafe34b98bf550fcc10e6d/lua/fzf-lua/libuv.lua#L614-L620

phanen:
I expected it to be quit here before, so apparently it didn't since schedule. 
```diff
diff --git a/lua/fzf-lua/libuv.lua b/lua/fzf-lua/libuv.lua
index b8f279f..34ca77a 100644
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -604,7 +604,7 @@ M.spawn_stdio = function(opts)
     end
     if type(content) == "function" then content(w, wn) end
     if type(content) == "table" then for _, v in ipairs(content) do w(v) end end
-    if type(content) ~= "string" then on_finish(0) end ---@cast content string
+    if type(content) ~= "string" then return on_finish(0) end ---@cast content string
     cmd = content
     if opts.debug then
       io.stdout:write(("[DEBUG] [mt] %s" .. EOL):format(content))
```

phanen:
> I'm trying to think is there every a case where we need to send table/function cmd contents to spawn here (doesn't seem so but mayeb I'm missing something)?

No, we should already quit.

But then we have to vim.wait for schedule done..

ibhagwan:
> > I'm trying to think is there every a case where we need to send table/function cmd contents to spawn here (doesn't seem so but mayeb I'm missing something)?
> 
> No, we should already quit.
> 
> But then we have to vim.wait for schedule done..

Why better solution than the commit (your diff is somewhat similar albeit less lines)?

phanen:
Will this help?

I think the callback is not in fast context.
```diff
diff --git a/lua/fzf-lua/libuv.lua b/lua/fzf-lua/libuv.lua
index b8f279f..f8643ab 100644
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -544,12 +544,13 @@ M.spawn_stdio = function(opts)
   local on_finish = function(code)
     pipe_close(stdout)
     pipe_close(stderr)
-    if fn_postprocess then
+    if vim.in_fast_event() and fn_postprocess then
       vim.schedule(function()
         fn_postprocess(opts)
         exit(code)
       end)
     else
+      if fn_postprocess then fn_postprocess(opts) end
       exit(code)
     end
   end
```

phanen:
> Why better solution than the commit (your diff is somewhat similar albeit less lines)?

Yes, nothing different just fewer line.

ibhagwan:
> Will this help?
> I think the callback is not in fast context.

I think so too, in the test specific case it’s not so it would exit immediately, but could there be a case where it’s fast event, if so we still need to exit early and `vim.wait` no?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/8698514d8349d3bde5db62947715956bf1c69d7a - implmented both your suggestions because why not, not seeing a way to get rid of the `vim.wait` in spawn.lua (yet), it's somewhat cosmetic, doesn't have real impact.

ibhagwan:
@phanen

Made a few more changes to be able to run tests from the main instance via a keybind->`MiniTest.{run_file|run_at_location}` (not cli `make`).

Any suggestions to make this less flakey?
https://github.com/ibhagwan/fzf-lua/actions/runs/23125980479/job/67168987742?pr=2619#step:7:36

After we fix this, if no objections I will merge?

ibhagwan:
https://github.com/ibhagwan/fzf-lua/blob/2713460db86657eefd4f446c3408b240a7a1f34c/tests/win_spec.lua#L114-L117

The test was failing with "already closed" which is acceptable in this case.

ibhagwan:
@phanen, will you be able to update your puc-lua release,? we're currently testing on a rather old nightly version `NVIM v0.12.0-dev-1187+ge084a99040`.

phanen:
Ok i ll take a look

ibhagwan:
any objections to merging this?

phanen:
No objections.

phanen:
> will you be able to update your puc-lua release,? we're currently testing on a rather old nightly version

No that's stable release https://github.com/phanen/neovim-releases/releases/download/stable/nvim-linux-x86_64.tar.gz

We are using nightly release
https://github.com/ibhagwan/fzf-lua/blob/bf59dcee2a8d894724b060401d2abf3d71d4bd47/.github/workflows/ci.yaml#L56

phanen:
Stable release is done manually upstream so I've delete it now to avoid confusing.

ibhagwan:
> Stable release is done manually upstream so I've delete it now to avoid confusing.

Ty @phanen!

ibhagwan:
@phanen, something happened to the puc lua release, it's no longer there:
https://github.com/phanen/neovim-releases/releases

Issue #1731, opened 2025-01-19
Feature: make `get_last_query` work correctly in  `hide` profile


phanen:
### Have you RTFM'd?

- [x] I have done proper research

## Problem
Continue of https://github.com/ibhagwan/fzf-lua/issues/1686#issuecomment-2600886621

## Solution
* by sending key to fzf terminal https://github.com/ibhagwan/fzf-lua/issues/1686#issuecomment-2600907158
* by `vim._on_key`
* by query parsing (require considering flags)
  * `--layout`
  * `--info-command`
  * `--input-border`
  * `--input-label`
  * `--input-label-pos`

PR #2633, opened 2026-03-27
fix(tsinjector): pass lnum to line_parser

phanen:
It's non trivial to determine where the "prompt" line is.
When we have reverse/default layout, info/header/footer border

This is a workaround for user find the prompt hl annoying:
```lua
opts.command_history._treesitter = function(line, lnum)
  if lnum == 1 then return end
  return line:match("(.-):?(%d+)[: ](.+)$")
end
```

## Summary by CodeRabbit

* **Enhancements**
  * Treesitter callbacks now receive optional line number information, enabling custom parsers to perform more contextual text processing.
  * Existing parsing behavior remains unchanged for current configurations.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -1389,7 +1389,7 @@ M.defaults.tags = {
   -- field_index_expr = "{}", -- For `_fmt.from` to work with `bat_native`
   _actions      = function() return M.globals.actions.files end,
   actions       = { ["ctrl-g"] = { actions.grep_lgrep } },
-  _treesitter   = function(line) return ctag_line_parser(line, true) end,
+  _treesitter   = function(line, _) return ctag_line_parser(line, true) end,
 }
 
 ---Search current buffer ctags.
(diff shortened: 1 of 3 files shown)


phanen:
-> Fix 2633-1 -> https://github.com/ibhagwan/fzf-lua/pull/2632

This is a hack but not sure if there better idea
Related: https://github.com/ibhagwan/fzf-lua/issues/1731#issue-2797711552

shyun3:
Just following up on this since I'm interested in fixing #2632. Are the lint failures the only blockers preventing this from being merged?

phanen:
I just wonder is there any better way to do it. But should be no problem to change it. This is still a internal opts with few users. I'll resolve the lint when available.

phanen:
When query is at the first line: 
```lua
          _treesitter = function(line, lnum)
            if lnum == 1 then return end
            return 'foo.vim', nil, line
          end,
```

cc @ibhagwan for review, not sure why but I could not merge it without review

ibhagwan:
> cc @ibhagwan for review, not sure why but I could not merge it without review

I disabled direct push to master enabled only via PRs - not sure why it wants my review, I’ll take a look at the ruleset.

ibhagwan:
Changed it to required reviewers 0 - can you see if you can merge again?

ibhagwan (approved):

ibhagwan:
I approved this, not sure why it still needed my review after setting required reviewers to 0, maybe in a new PR or resubmitted commit it will work (since I was already required to review).

ibhagwan:
I think I fixed it now, no reviews should be required

Issue #2636, opened 2026-03-30
Bug: repeat text on git_branches


PolygonT:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

zsh

### Neovim version (`nvim --version`)

NVIM v0.12.0-dev-2550+gf847aa6208

### Fzf version (`fzf --version`)

0.44.1 (debian)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
                git = {
                    branches = {
                        preview = [[git log --graph --pretty=format:"%C(yellow)%h %C(green)%ad %C(reset)%s %C(blue)%an%C(reset)" --date=format:"(%Y-%m-%d %H:%M)" --abbrev-commit --color {1}]]
                    },


})
```

### Describe the bug / steps to reproduce

I change git_branches preview ```--pretty=format:"%C(yellow)%h %C(green)%ad %C(reset)%s %C(blue)%an%C(reset)" --date=format:"(%Y-%m-%d %H:%M)"``` to show the date on each commit, i find on each branches first commit, showing double preview text.

<img width="1855" height="887" alt="Image" src="https://github.com/user-attachments/assets/3e4216d1-8c3b-4622-960d-63c5a015e189" />
but when i execute ```git log ...``` in shell, it preform normally.

PR #2637, opened 2026-03-30
fix(libuv): duplicated prev_line_content when stream EOF

phanen:
When the stream ends (signified by `data` being `nil`), the function
checked if there was any `prev_line_content` leftover.

prev_line_content isn't nullified and will be used again in split_lines:
```
prev_line_content .. EOL .. prev_line_content
```

Fix https://github.com/ibhagwan/fzf-lua/issues/2636.


ibhagwan:
Ty @phanen!

Interesting this took so long to become visible

PR #2642, opened 2026-04-02
fix(preview): typing no longer errors when there is no preview buffer

ColinKennedy:
`self.preview_bufnr` has a chance of not being defined when there are no matches to preview. This PR checks in-advance to prevent any spammy errors during typing.

## Before
https://github.com/user-attachments/assets/a7530457-a11a-4895-bfce-b19ac3a25ba9

## After
https://github.com/user-attachments/assets/de9407d6-63d5-44ba-9ac5-ee269269f173

## Bug Fixes
* Fixed a potential error in preview buffer handling that could occur when image attachments are detected.


ibhagwan:
Ty @ColinKennedy !

PR #2644, opened 2026-04-03
feat(win): restore last accessed (CTRL-W p) window after closing fzf …

illia-shkroba:
I often use a tab with 3 windows and I switch between the 2 of them with `CTRL-W p` binding.
The problem is that when I spawn a FZF window (e.g. `FzfLua lines`) and then close it, the `CTRL-W p` binding no longer works.

~Although this PR fixes the issue, I hope there might be a better way of extracting the last window without triggering autocommands `WinEnter` and `WinLeave` on the `last_winid`.~

~If this change is "too aggressive", we might add a new option to the config to enable this last accessed window preserving behaviour.~

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable window focus restoration when closing search/overlay, preventing accidental focus loss.
  * Reduced focus flicker during teardown for a smoother close experience.
  * Closing the overlay now consistently preserves the previously active window.

### DIFF
--- a/lua/fzf-lua/ctx.lua
+++ b/lua/fzf-lua/ctx.lua
@@ -11,6 +11,7 @@ local ctx
 ---@field bufnr integer
 ---@field bname string
 ---@field winid integer
+---@field last_winid integer
 ---@field alt_bufnr integer
 ---@field tabnr integer
 ---@field tabh integer
@@ -51,6 +52,7 @@ M.refresh = function(opts)
       bufnr = vim.api.nvim_get_current_buf(),
       bname = vim.api.nvim_buf_get_name(0),
       winid = vim.api.nvim_get_current_win(),
+      last_winid = vim.fn.win_getid(vim.fn.winnr("#")),
       alt_bufnr = vim.fn.bufnr("#"),
       tabnr = vim.fn.tabpagenr(),
       tabh = vim.api.nvim_win_get_tabpage(0),
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -1140,6 +1140,17 @@ function FzfWin:close_preview(do_not_clear_cache)
   self.preview_winid = nil
 end
 
+---@param winid? integer
+---@param last_winid? integer
+local restore_lastwin = function(winid, last_winid)
+  if winid and last_winid and api.nvim_win_is_valid(last_winid) then
+    utils.eventignore(function()
+      api.nvim_set_current_win(last_winid)
+      api.nvim_set_current_win(winid)
+    end)
+  end
+end
+
 ---@param buf? integer
 local restore_altbuf = function(buf)
   if buf and api.nvim_buf_is_valid(buf) then
@@ -1166,6 +1177,8 @@ function FzfWin:close(fzf_bufnr, hide)
   local ctx = utils.__CTX() or {}
   if ctx.mode == "nt" then vim.cmd "stopinsert" end
   if self.fzf_winid and api.nvim_win_is_valid(self.fzf_winid) then
+    -- restore the original last window
+    restore_lastwin(ctx.winid, ctx.last_winid)
     -- run in a pcall due to potential errors while closing the window
     -- Vim(lua):E5108: Error executing lua
     -- experienced while accessing 'vim.b[]' from my statusline code


illia-shkroba:
I've applied the `noautocmd` change suggested by the coderabbitai.
The autocommands triggering is no longer an issue.

ibhagwan:
Ty for the contribution @illia-shkroba.

I must admit I don’t like the approach of jumping between windows just to get/set the last window for ctrl-w-p, IMHO it’s not clean and error prone, is this really the only possible way to get/set the last window id?

It would be worthwhile to try and look into the internals to get this in a more efficient way that doesn’t use `noautocmd wincmd p` or better yet find a way that opening a new float window won’t affect the stored “last window”?

lua/fzf-lua/ctx.lua
```diff
@@ -51,6 +52,12 @@ M.refresh = function(opts)
       bufnr = vim.api.nvim_get_current_buf(),
       bname = vim.api.nvim_buf_get_name(0),
       winid = vim.api.nvim_get_current_win(),
+      last_winid = (function()
+        vim.cmd("noautocmd wincmd p")
+        local ret = vim.api.nvim_get_current_win()
+        vim.cmd("noautocmd wincmd p")
```
phanen: `:h winnr()`, can be `winnr("#")`?
```
			#	the number of the last accessed window (where
				|CTRL-W_p| goes to).  If there is no previous
				window or it is in another tab page 0 is
				returned.  May refer to the current window in
				some cases (e.g. when evaluating 'statusline'
				expressions).
```

lua/fzf-lua/win.lua
```diff
@@ -1140,6 +1140,20 @@ function FzfWin:close_preview(do_not_clear_cache)
...
+---@param last_winid? integer
+local restore_lastwin = function(winid, last_winid)
+  if
+      winid
+      and api.nvim_win_is_valid(winid)
+      and last_winid
+      and api.nvim_win_is_valid(last_winid)
+  then
+    vim.cmd("noautocmd lua vim.api.nvim_set_current_win(" .. last_winid .. ")")
+    vim.cmd("noautocmd lua vim.api.nvim_set_current_win(" .. winid .. ")")
```
phanen: to avoid concat string, we can use `utils.eventignore()` to wrap the function

lua/fzf-lua/ctx.lua
phanen: OT: vim also have `tabpagenr("#")`, I think fzf-lua can also mess up it when `split=tabnew`

lua/fzf-lua/win.lua
illia-shkroba: Nice! It looks better with the `utils.eventignore` wrapper.

lua/fzf-lua/ctx.lua
illia-shkroba: If you would use `winnr` you'll get a window number from top window starting from 1. It's the same number you use when doing `[count] CTRL-W w`. But this number is incompatible with the `vim.api.nvim_set_current_win`. Also, I couldn't find a function to set the last accessed window.

illia-shkroba:
> Ty for the contribution @illia-shkroba.
> 
> I must admit I don’t like the approach of jumping between windows just to get/set the last window for ctrl-w-p, IMHO it’s not clean and error prone, is this really the only possible way to get/set the last window id?
> 
> It would be worthwhile to try and look into the internals to get this in a more efficient way that doesn’t use `noautocmd wincmd p` or better yet find a way that opening a new float window won’t affect the stored “last window”?

I don't like it neither. But I'm aftraid there is not much we can do about it if there is no function for it exposed in the API.

lua/fzf-lua/win.lua
illia-shkroba: After switching to this function the tests are failing. Could you please take a look and suggest how to fix?

ibhagwan:
> I don't like it neither. But I'm aftraid there is not much we can do about it if there is no function for it exposed in the API.

What about `vim.api.nvim_tabpage_list_wins(0)[vim.fn.winnr("#")]`?

lua/fzf-lua/win.lua
```diff
@@ -1140,6 +1140,22 @@ function FzfWin:close_preview(do_not_clear_cache)
   self.preview_winid = nil
 end
 
+---@param winid? integer
+---@param last_winid? integer
+local restore_lastwin = function(winid, last_winid)
+  if
+      winid
+      and api.nvim_win_is_valid(winid)
```
ibhagwan: Since this is only called after verification of `api.nvim_win_is_valid(self.fzf_winid)` I don’t think we need another call to nvim_win_is_valid unless we plan to call this from somewhere else sometime in the future

lua/fzf-lua/ctx.lua
phanen: `:echo winnr('#')->win_getid()`

lua/fzf-lua/win.lua
phanen: Weird.. I don't think this changes is related.

lua/fzf-lua/win.lua
illia-shkroba: Sure. I did it for consistency with `restore_altbuf`, but I think we can get rid of it.

lua/fzf-lua/ctx.lua
illia-shkroba: Awesome! That should do.

illia-shkroba:
> > I don't like it neither. But I'm aftraid there is not much we can do about it if there is no function for it exposed in the API.
> 
> What about `vim.api.nvim_tabpage_list_wins(0)[vim.fn.winnr("#")]`?

I end up using `vim.fn.win_getid(vim.fn.winnr("#"))`.
As suggested by @phanen .

lua/fzf-lua/win.lua
ibhagwan: You got rid of the nvim_win_is_valid(last_winid) too, that one is important IMHO.

illia-shkroba: No issue. I brought it back.

phanen:
Ok, the ci issue is `vim.NIL` taken as `nil` in test. I will force push the fix on this branch.

ibhagwan:
> Ok, the ci issue is `vim.NIL` taken as `nil` in test. I will force push the fix on this branch.

Ty @phanen!

PR #2659, opened 2026-04-12
refactor: speed up parsing/transform with stringbuf/worker

### LINKED ISSUE #2657: Bug: tags command gets a lot of time indexing
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Mac

### Shell

zsh

0.12.0

0.70.0 (Homebrew)

mostly color customisations

- [ ] My issue is reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua

```

I noticed that the tags command gets a lot of time to index the tags from my repository. Sometimes I search for a tag and is not matched until later because it's still indexing. A long ago I opened an issue as well and I remember rg_glob = false solved it for grep command. But it seems the tags command doesn't have this option, still I'm not even sure if it would be the same issue. On fzf.vim it takes not even 2 seconds to load all the tags, and fzf-lua it takes more than 10 seconds. The number of tags seems to be 1535991, in both of them shows the same quantity.

phanen:
Try to fix https://github.com/ibhagwan/fzf-lua/issues/2657#issuecomment-4231199833.

cannot work correctly now...

~Not sure why it missing some entries.~ should be correct, I change the file so the grep result is changed...

## Summary by CodeRabbit

* **Refactor**
  * Updated internal buffering and data handling mechanisms.
  * Introduced new internal string buffer utility module.
  * Adjusted internal process data handling logic.

* **Chores**
  * Updated internal code documentation and feature flag annotations.


phanen:
Since all vim function (`vim.api/fn`) is unusable in thread, we cannot load icons..

phanen:
This indeed make st significantly faster, I will try if it's possible to pass the serialized options to worker.

phanen:
> Since all vim function (`vim.api/fn`) is unusable in thread, we cannot load icons..

Manage to make nvim-web-devicons work. It's 3~4x faster on large repo. 
```lua
_G.time = vim.uv.hrtime()
require('fzf-lua').grep({
  search = '',
  multiprocess = false,
  file_icons = true,
  path_shorten = 1,
  keymap = { fzf = { load = 'abort' } },
  winopts = { on_close = function() print((vim.uv.hrtime() - _G.time) * 1e-9) end },
})
```
Before: `11.696383104`, After: `2.595442664` (`UV_THREADPOOL_SIZE=8`)

ibhagwan:
> Before: 11.696383104, After: 2.595442664 (UV_THREADPOOL_SIZE=8)

that’s impressive improvement

barrettruth:
is there no stringbuffer in core?(almost certainly) this could be useful there, what do you think :thinking:

phanen:
> is there no stringbuffer in core?(almost certainly) this could be useful there, what do you think 🤔

Yes, currently a shim is needed for PUC lua ("vim._core.stringbuffer" is useful but it missing method like `:ref()`, always `#` operator have to be written as `sb:__len()`).

Related issue here: https://github.com/neovim/neovim/issues/39316

ofseed:
> "vim._core.stringbuffer" is useful but it missing method like `:ref()`

FYI, I previously added support for `ref` to the nvim core shim. https://github.com/neovim/neovim/pull/39389

PR #2663, opened 2026-04-13
fix(help): use action header as desc fallback in F1 keybind help

selmison:
## Problem

The F1 help window shows `table:0x...` for actions that define a `header` field but not `desc`, when the action's `fn` is an anonymous function not registered in the helpstr table.

The `get_desc` function in `win/help.lua` resolves descriptions via this chain:

```
v.desc -> helpstr(v[1]) -> helpstr(v.fn) -> tostring(v)
```

For actions like `git.commits` `ctrl-d` (which has `header = "git diff"` but no `desc`, and uses an anonymous `fn`), every step before `tostring(v)` returns `nil`, so the help window displays the unhelpful `table:0x...` string.

## Fix

Add `v.header` as a fallback between `config.get_action_helpstr(v.fn)` and `tostring(v)`:

```lua
return v.desc or config.get_action_helpstr(v[1]) or config.get_action_helpstr(v.fn) or
    v.header or tostring(v)
```

This is consistent with how `header` is already used as a human-readable label for actions (it populates the `--header` line at the top of pickers).

## Affected actions

6 built-in actions in `defaults.lua` that have `header` but no `desc`:

| Picker | Key | Header (now shown in F1) |
|---|---|---|
| `git.diff` | `ctrl-q` | `"git commits"` |
| `git.commits` | `ctrl-d` | `"git diff"` |
| `spellcheck` | `ctrl-s` | `"spell suggest"` |
| `serverlist` | `ctrl-o` | `"spawn"` |
| `serverlist` | `ctrl-x` | `"kill"` |
| `serverlist` | `ctrl-r` | `"reload"` |

Actions with `header = false` (used to suppress the header line) are unaffected since `false` is falsy and the chain continues to `tostring(v)`.

## Before / After

**Before:** F1 help shows `table: 0x56327a5de490` for these keybinds.

**After:** F1 help shows the `header` string (e.g. `"git diff"`, `"spell suggest"`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Help/description text is now generated consistently so preview keymaps and action descriptions display correctly and reliably.

* **New Features**
  * Help text resolution supports richer action specifications (including an optional header), improving how actions are labeled and presented in UI lists.

### DIFF
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -1063,8 +1063,14 @@ M.set_action_helpstr = function(fn, helpstr)
   M._action_to_helpstr[fn] = helpstr
 end
 
-M.get_action_helpstr = function(fn)
-  return M._action_to_helpstr[fn]
+---@param v fzf-lua.ActionSpec|any
+---@return string
+M.get_action_helpstr = function(v)
+  local t = M._action_to_helpstr
+  if type(v) ~= "table" or t[v] then return t[v] or tostring(v) end
+  local res = v.desc or t[v[1]] or t[v.fn] or v.header
+  if res then return res end
+  return type(v[1]) == "string" and v[1] or tostring(v)
 end
 
 M._action_to_helpstr = {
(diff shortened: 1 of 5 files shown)


phanen:
Good idea, I think we can move all get_desc logic to `get_action_helpstr`, to make it accept table/string/function.

Issue #2670, opened 2026-04-16
Bug: default action cant open file in buffer


vanhtuan0409:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

fish

### Neovim version (`nvim --version`)

0.12.1

### Fzf version (`fzf --version`)

0.71.0 (62899fd7)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
})
```

### Describe the bug / steps to reproduce

when open files picker and select a file, FZF lua failed to open the buffer with error `Unable to add buffer`

<img width="962" height="152" alt="Image" src="https://github.com/user-attachments/assets/73c1b6ea-0d0d-4088-b54b-ece8dd844d3d" />

PR #2671, opened 2026-04-16
fix(actions): pass nil to can_replace_buf instead of 0

vanhtuan0409:
## Problem

Fixing https://github.com/ibhagwan/fzf-lua/issues/2670

When `hidden` is not set (default), selecting a file from fzf-lua pickers (e.g. `files`, `grep`) fails with:

```
Unable to add buffer <path>
```

This happens because `can_replace_buf` is called with `0` as the buffer argument at `actions.lua:242`. In Lua, `0` is truthy, so the fallback `buf = buf or nvim_get_current_buf()` inside `can_replace_buf` never triggers — the function proceeds with `buf = 0`.

This propagates down to `buffer_is_dirty` → `getbufinfo(0)`, which returns an empty table (no buffer with id 0 exists). The `changed` field on that empty table is `nil`, and since `nil ~= 0` evaluates to `true`, the function always reports the buffer as dirty, causing `can_replace_buf` to always return `false`.

## Fix

Pass `nil` instead of `0` so that `can_replace_buf` correctly falls back to `nvim_get_current_buf()` and checks the actual current buffer's dirty state.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal buffer selection handling to more consistently manage buffer references across operations.

### DIFF
--- a/lua/fzf-lua/actions.lua
+++ b/lua/fzf-lua/actions.lua
@@ -192,7 +192,6 @@ end
 
 ---@return boolean
 local can_replace_buf = function(buf)
-  buf = buf or vim.api.nvim_get_current_buf()
   return vim.o.hidden
       or vim.o.confirm
       or vim.o.autowriteall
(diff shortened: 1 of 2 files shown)


phanen:
> This propagates down to `buffer_is_dirty` → `getbufinfo(0)`, which returns an empty table (no buffer with id 0 exists). The `changed` field on that empty table is `nil`, and since `nil ~= 0` evaluates to `true`, the function always reports the buffer as dirty, causing `can_replace_buf` to always return `false`.


Thanks. The confusing thing is it's not alreay empty.

`0` means alternate buffer in vimscript. So it's empty when no altbuffer exsit.
https://github.com/neovim/neovim/blob/19fff0e0bed2a77c8f8d27070cc12084bce24000/src/nvim/buffer.c#L2678-L2683

vanhtuan0409:
no problems, I can only actually trace back to the `can_replace_buf` function. The rest is actually claude troubleshooting

PR #2680, opened 2026-04-17
fix(actions): werid issue when set verbose=1

### LINKED ISSUE #2679: bug: file_tabedit don't work with lsp entry
I cannot believe, why I feel it used to work, but checkout a random commit half year again, it seems never work.

phanen:
Maybe there're other similar issue, to be discovered

Fix https://github.com/ibhagwan/fzf-lua/issues/2679

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable jump-to-location behavior on Neovim 0.11+ to reduce noisy output and improve navigation.
  * Improved stability on Windows by adding short delays to reduce race conditions during rapid file/preview operations.
  * Consistent previewer resize behavior on Windows to prevent flaky layout changes during interactive resizing.

### DIFF
--- a/lua/fzf-lua/utils.lua
+++ b/lua/fzf-lua/utils.lua
@@ -1625,8 +1625,10 @@ end
 ---@return boolean
 function M.jump_to_location(location, offset_encoding, reuse_win)
   if M.__HAS_NVIM_011 then
-    return vim.lsp.util.show_document(location, offset_encoding,
-      { reuse_win = reuse_win, focus = true })
+    return M.with({ o = { verbose = 0 } }, function()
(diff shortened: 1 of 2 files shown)


ibhagwan:
Windows CI is so annoying it has become worse lately, I hate this “solution” but considering skipping more tests on Windows, such a horrible OS

phanen:
Not sure if this is a real bug.
https://github.com/ibhagwan/fzf-lua/blob/77cfef6bf42a9190c9b450982376585cbb8859cb/tests/win_spec.lua#L150
This table is a random order, and each time ci failure will happened on toggle-preview

phanen:
but in any way the child/channel instance should not die

PR #2687, opened 2026-04-20
fix(previewer): rg errmsg is mistakenly taken as ctags pattern

phanen:
Problem:
When rg cmd fail with errmsg e.g. `rg: PCRE2: error`
This will be parsed and treated as an entry(with lnum=0)

`matchadd` will remain even buf changed in window, so
we have to clear the matches on each preview.

To reproduce: `FzfLua live_grep search=\] no_esc=true`
then type `<bs>`, `]`, `<bs>`, `]`... (with rg --pcre2)

Maybe this can also happened in `FzfLua global`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cursor highlighting accuracy in preview mode for entries with missing line numbers by refining condition logic.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1265,7 +1265,7 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
   if type(cached_pos) ~= "table" then cached_pos = nil end
   local lnum, col = entry.line, math.max(1, entry.col or 1)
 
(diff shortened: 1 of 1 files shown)


phanen:
> `matchadd` will remain even buf changed in window, so

> Maybe this can also happened in `FzfLua global`.

If it happened, maybe if anyone mix ctags with lsp in `FzfLua global`...
(I'm not sure if this possible)
We need to move `win_call clearmaches()` out from branch and always run it.

PR #2691, opened 2026-04-22
fix(previewer): make "smartcase" work again

phanen:
No flags for case sensitive, and flags like "\C" may break previous
smartcase hack

Related: https://github.com/ibhagwan/fzf-lua/commit/7b3a7390d01db93d422a4d6d6cf3771038577858

## Summary by CodeRabbit

* **Refactor**
  * Optimized preview highlighting logic by precomputing case sensitivity once instead of repeatedly checking it during regex matching, improving efficiency when displaying search results with case-sensitive and case-insensitive patterns.

### DIFF
--- a/lua/fzf-lua/previewer/builtin.lua
+++ b/lua/fzf-lua/previewer/builtin.lua
@@ -1252,6 +1252,7 @@ function Previewer.buffer_or_file:set_cursor_hl(entry)
     end)()
   end
 
+  local case_sensitive = regex and regex:lower() ~= regex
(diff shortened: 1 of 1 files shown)


ibhagwan:
Where do we need  case insensitive search?

phanen:
When grep result is "case-insensitive"
https://github.com/ibhagwan/fzf-lua/blob/1c983b3892bbc2e1841d65fbeb58f8040f4245d4/lua/fzf-lua/defaults.lua#L968

ibhagwan:
Oh yes Smart Case

PR #2662, opened 2026-04-12
fix(utils): hide profile should not inherit default

leaty:
fix #2661

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed profile loading behavior to prevent automatic injection of the default profile when using the "hide" profile option.

Issue #2661, opened 2026-04-12
Bug: Inability to (easily) override title for all pickers


leaty:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

bash

### Neovim version (`nvim --version`)

NVIM v0.12.1

### Fzf version (`fzf --version`)

0.71.0 (62899fd7)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup {
	defaults = { winopts = { title = false } },
}
```

### Describe the bug / steps to reproduce

See #2660

leaty:
FYI:
https://github.com/ibhagwan/fzf-lua/blob/114da56bb1b14b4aec073123fc1ef9346469967e/lua/fzf-lua/utils.lua#L982-L985
In regards to https://github.com/ibhagwan/fzf-lua/discussions/2660#discussioncomment-16535490; and `default` loads `default-title` via `border-fused` profile.

leaty:
This does in fact override it correctly:
https://github.com/ibhagwan/fzf-lua/blob/114da56bb1b14b4aec073123fc1ef9346469967e/lua/fzf-lua/config.lua#L127-L132

But the problem seems to be that the profile(s) are always merged into  `config.setup_opts` in `init.setup()`. Because when the following runs (*after* the above) it overrides the titles *again*- which wouldn't be a problem if `setup_ops` only contained the user's config, but since it has been merged with the profile it doesn't matter what you do, titles will always be reset by the profile.
https://github.com/ibhagwan/fzf-lua/blob/114da56bb1b14b4aec073123fc1ef9346469967e/lua/fzf-lua/config.lua#L133-L134


It's a bit confusing to me though so I might be wrong. Shifting the order around like this works:
```lua
    -- (3) override with the specific provider options from the users's `setup` option
    ret = vim.tbl_deep_extend("force", ret, utils.map_get(setup_opts(), index) or {})
[...]

PR #2692, opened 2026-04-22
fix(config): user setup should override profile per-picker options

barrettruth:
allows forcing empty titles, which previously was not honored with `winopts.title = false`

example:

image

## Summary by CodeRabbit

* **New Features**
  * `winopts.title` can be set to false to hide the fzf window title and suppress picker label fallbacks (statusline and border-labels are omitted).

* **Documentation**
  * Docs updated to show `winopts.title: string|false` and explain the false behavior.

* **Tests**
  * New tests cover title=false behavior, precedence, and persistence across setups and splits.

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -541,9 +541,9 @@ Neovim split command to use for fzf-lua interface, e.g `belowright new`.
 
 #### globals.winopts.title
 
-Type: `string`, Default: `nil`
+Type: `string|false`, Default: `nil`
 
-Controls title display in the fzf window, set by the calling picker.
+Controls title display in the fzf window, set by the calling picker. Set to `false` to hide the title and suppress picker label fallbacks.
 
 #### globals.winopts.title_flags
 
--- a/doc/fzf-lua-opts.txt
+++ b/doc/fzf-lua-opts.txt
@@ -759,9 +759,10 @@ Neovim split command to use for fzf-lua interface, e.g `belowright new`.
                                                                               
 globals.winopts.title                     *fzf-lua-opts-globals.winopts.title*
 
-Type: `string`, Default: `nil`
+Type: `string|false`, Default: `nil`
 
-Controls title display in the fzf window, set by the calling picker.
+Controls title display in the fzf window, set by the calling picker. Set to
+`false` to hide the title and suppress picker label fallbacks.
 
 
                                                                               
--- a/lua/fzf-lua/config.lua
+++ b/lua/fzf-lua/config.lua
@@ -65,6 +65,7 @@ end
 -- proxy table (with logic) for accessing the global config
 ---@type fzf-lua.Config|{}
 M.setup_opts = {}
+M.setup_call_opts = {}
 
 ---@type fzf-lua.config.Defaults
 M.globals = setmetatable({}, {
@@ -152,7 +153,6 @@ local eval = function(v, ...)
   return v
 end
 
-
 ---expand opts that were specified with a dot
 ---@param opts table
 local normalize_tbl = function(opts)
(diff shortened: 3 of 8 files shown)


ibhagwan:
Is `defaults = { winopts.title = false }` not doing the same?

This seems like an addition of code which could be avoided by changing the default profile too, see #2661, #2662

phanen:
If user still want border-fused, it will unavoidably apply "default-title". How about move default-title out
```diff
diff --git a/lua/fzf-lua/profiles/border-fused.lua b/lua/fzf-lua/profiles/border-fused.lua
index 111a12a..613e845 100644
--- a/lua/fzf-lua/profiles/border-fused.lua
+++ b/lua/fzf-lua/profiles/border-fused.lua
@@ -2,7 +2,6 @@ local _single  = { "┌", "─", "┐", "│", "┘", "─", "└", "│" }
 local _rounded = { "╭", "─", "╮", "│", "╯", "─", "╰", "│" }
 local _border  = true and _rounded or _single
 return {
-  { "default-title" }, -- base profile
   desc = "Single border around the UI",
   -- previewers = { bat = { args = "--color=always --style=default" } },
   winopts = {
diff --git a/lua/fzf-lua/profiles/default.lua b/lua/fzf-lua/profiles/default.lua
index 0cf488b..20a2be0 100644
--- a/lua/fzf-lua/profiles/default.lua
+++ b/lua/fzf-lua/profiles/default.lua
@@ -1,4 +1,4 @@
 return {
   desc = "fzf-lua default options",
-  { "border-fused", "hide" },
+  { "default-title", "border-fused", "hide" },
 }
```

ibhagwan:
> If user still want border-fused, it will unavoidably apply "default-title". How about move default-title out

Yes, we can do that but it’s somewhat breaking as most users expect a title, I generally try to avoid breaking changes,

What look are you going for @barrettruth? I’m pretty sure it can be achieved simply by not using the default profile as in the issue i linked.

barrettruth:
Everything you said makes sense. Profiles are a bundled set of options/settings and I overlooked this (lmao, i've been talking to ai too much i sound like a bot).

I got around this by just not using the default profile with `[1] = false`

However, I’m leaving one last note related to this larger issue: I think global settings ought to override profile defaults. After all, if the user sets a profile and setting X, they want to see the profile and then THEIR opinion of setting X over the profile's opinion. And, I think respecting this is not breaking backwards-compatibility or anything like that in any way.

This is NOT how the current logic works - you HAVE to disable profiles as I did above.

Are you open to adding this logic or nah?

Feel free to close this pr and I'll open a new one if so.

ibhagwan:
I completely agree, it mostly works this way where the setup will override the profile, which is basically a set of options added to setup, perhaps there’s a simpler logic for the one suggested in this PR focused on merge priority (which option is merged first)?

barrettruth:
I pivoted to the merge-priority approach. Namely, user top-level setup now beats profile per-picker values, while call-time opts and explicit per-picker setup still win.
This also fixes handling string titles like title = " Custom " (previously getting dropped)

barrettruth:
is this ready or can it be closed as resolved by `e325fab`

ibhagwan:
> is this ready or can it be closed as resolved by `e325fab`

I’m working on a refactor how config values are inherited, most likely will solve this, up to you if you wanna close it.

barrettruth:
ok... i'll close it - do you mind letting me know if it *doesn't* solve it when merged :sweat:

ibhagwan:
> ok... i'll close it - do you mind letting me know if it _doesn't_ solve it when merged 😓

Absolutely

PR #2705, opened 2026-04-26
fix(previewer): set i keymap only for normal mode

PeterCardenas:
Allows for keymaps like `yip` to work. Previously it would focus the prompt terminal window

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keymap functionality in the preview window to work correctly in normal mode.


ibhagwan:
Ty @PeterCardenas

PR #2727, opened 2026-05-14
fix(health): skip --version for Windows built-ins and fix CRLF

Namacha411:
1. On Japanese Windows, `find --version` invokes Windows' system `find.exe` which outputs an error in CP932 encoding, causing garbled text in `:checkhealth`. Also, `dir` has the same issue.       

2. Skip `--version` for these Windows built-ins and report them as available without a version string. Also switch to `utils.io_system()` to avoid CRLF-corrupted version strings on Windows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tool detection to use more reliable version checking mechanism.
  * Added Windows compatibility for built-in commands to prevent UTF-8 encoding errors.
  * Enhanced version string retrieval for accurate tool identification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ibhagwan/fzf-lua/pull/2727)

PR #2728, opened 2026-05-15
fix: replace os.exit with :cquit for nvim -l

phanen:
https://github.com/ibhagwan/fzf-lua/pull/2727#issuecomment-4457133021

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application exit handling for better compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ibhagwan/fzf-lua/pull/2728)

### DIFF
--- a/lua/fzf-lua/libuv.lua
+++ b/lua/fzf-lua/libuv.lua
@@ -502,7 +502,7 @@ M.spawn_stdio = function(opts)
 
   local function exit(exit_code, msg)
     if msg then stderr_write(msg) end
-    os.exit(exit_code)
+    vim.cmd.cquit(exit_code)
   end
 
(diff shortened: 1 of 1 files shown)


phanen:
not sure if `require('os')` and `os` are the same thing but worth a try.

https://github.com/neovim/neovim/blob/9432e6c1e2e93ffffabeeeaeb6ccc0cbdf9609dd/src/nvim/lua/executor.c#L856-L859

EDIT: ok it don't work.

ibhagwan:
Tysm @phanen!

Reading the code rabbit comment, do you think there is a scenario it needs to be called via vim.schedule?

phanen:
It's never called in fast context directly, so it's ok.

PR #2738, opened 2026-05-31
fix(grep): grep_cword fails for words starting/ending with non-word chars

mikehaertl:
If `iskeyword` contains a character that `rg` does not consider a word character (e.g. `$`) then `grep_cword` will fail if the word starts or ends with such a character.

For example if `$` is in `iskeyword` then for `demo$` the regex will be `\bdemo\$\b` - but `rg` does not consider `$` to be a word character so the `\b` will never match.

The suggested fix here leaves away the `\b` in such cases. It's not as sharp as `\b` matching anymore but I'd say that's more correct than no results at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved word-boundary handling when searching for the word under the cursor: searches now use a safer escaping routine and apply word boundaries only when appropriate, preventing missed or incorrect matches.
  * Ensures grep and tag searches produce more accurate results for edge-case words.

### DIFF
--- a/lua/fzf-lua/providers/grep.lua
+++ b/lua/fzf-lua/providers/grep.lua
@@ -266,7 +266,7 @@ M.grep_cword = function(opts)
   opts = opts or {}
   opts.no_esc = true
   -- match whole words only (#968)
-  opts.search = [[\b]] .. utils.rg_escape(vim.fn.expand("<cword>")) .. [[\b]]
+  opts.search = utils.rg_escape_cword(vim.fn.expand("<cword>"))
   return M.grep(opts)
 end
 
--- a/lua/fzf-lua/providers/tags.lua
+++ b/lua/fzf-lua/providers/tags.lua
@@ -251,7 +251,7 @@ end
 M.grep_cword = function(opts)
   if not opts then opts = {} end
   opts.no_esc = true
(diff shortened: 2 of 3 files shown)


ibhagwan:
> For example if $ is in iskeyword then for demo$ the regex will be \bdemo\$\b - but rg does not consider $ to be a word character so the \b will never match.

Ty @mikehaertl but I’m not sure I understand the problem, word boundaries in nvim do not contain special characters, so grep_cword will never get to search for demo$, for example if my cursor is on d and I run `:= vim.fn.expand("<cword>")` the result is `demo` and not `demo$`.

Can you provide an example where grep_cword fails?

mikehaertl:
>  word boundaries in nvim do not contain special characters

AFAIK `iskeyword` can be different per filetype. So some languages *do* include `$` for example typescript. If I have this `test.ts`:

```ts
class A {
  private demo$;
}
```

```
:set iskeyword
iskeyword=@,48-57,_,192-255,#,$

vim.fn.expand("<cword>")
"demo$"
```

ibhagwan:
> AFAIK iskeyword can be different per filetype. So some languages do include $ for example typescript. If I have this test.ts:

TIL, ty @mikehaertl!

Issue #1462, opened 2024-09-30
Bug: when using `split`, vim resize while fzf is open causes wrong `cmdheight` on close


jfly:
### RTFM Checklist

- [X] I have searched exisiting issues / discussions
- [X] I have read the Wiki including the Advanced section
- [X] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

fish

### Neovim version (`nvim --version`)

v0.11.0-nightly+69553f7

### Fzf version (`fzf --version`)

0.55.0

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [X] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
require('fzf-lua').setup({
    winopts = {
      split = "belowright new",
      height = 0.4,
      preview.hidden = "hidden",
    };
})
```

### Describe the bug / steps to reproduce

Open fzf-lua, resize neovim, and then close fzf lua. You'll be left with an empty space at the bottom that I cannot figure out how to make go away (other than restarting neovim). `:FzfLua redraw` doesn't fix things.

This is easiest to just see in a video, here you go:

https://github.com/user-attachments/assets/18046175-d045-4c40-b669-bd402c0590b0

jfly:
That's it! Before bug: `cmdheight=1`. After bug: `cmdheight=12`. `:set cmdheight=1` fixes it.

ibhagwan:
> That's it! Before bug: `cmdheight=1`. After bug: `cmdheight=12`. `:set cmdheight=1` fixes it.

Thanks for the video which made this clear and for confirming @jfly, I’ll fix it shortly.

ibhagwan:
https://github.com/ibhagwan/fzf-lua/commit/d76cc35a2945beb4d916050a8ad34b7daf1b4054

@jfly, should work as expected now.

jfly:
That works great! Thank for for the quick response and quick fix =)

PR #2739, opened 2026-05-31
fix(win): "E749: Empty buffer" on closing fzf due to continuous "|"

Bekaboo:
Bug: If `cmd` contains empty string "", after concatenation the concatenated command will contains continuous separator "|", e.g. ":1resize 9|vert :1resize 103|||:1resize 9|vert :1resize 103|||", running this command through `vim.cmd()` will cause error: "E749: Empty buffer".

image

Fix: Filter out empty commands in `cmd`.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable restoration of split windows and sizes when closing temporary windows; prevents running empty or invalid restore commands to avoid spurious behavior.
  * Ensures only valid resize fragments are applied during restore, reducing edge-case failures.

* **Refactor**
  * Internal cleanup of window-restoration logic for clearer, safer execution and easier maintenance.

### DIFF
--- a/lua/fzf-lua/win.lua
+++ b/lua/fzf-lua/win.lua
@@ -1066,11 +1066,14 @@ function FzfWin:create()
       -- if we're not doing this the result might be all over the place
       local winnrs = vim.tbl_map(api.nvim_win_get_number, api.nvim_tabpage_list_wins(0))
       local parts = vim.split(winrestcmd, "|")
-      local cmd = vim.tbl_map(function(cmd_part)
-        local winnr = tonumber(cmd_part:match("(.)resize"))
-        return utils.tbl_contains(winnrs, winnr) and cmd_part or ""
-      end, parts)
-      vim.cmd(table.concat(cmd, "|"))
+      local cmd = {}
+      for _, cmd_part in ipairs(parts) do
+        local winnr = tonumber(cmd_part:match("(%d+)resize"))
+        if utils.tbl_contains(winnrs, winnr) then
+          table.insert(cmd, cmd_part)
+        end
+      end
+      if #cmd > 0 then vim.cmd(table.concat(cmd, "|")) end
       -- Also restore cmdheight, will be wrong if vim resized (#1462)
       vim.o.cmdheight = cmdheight
     end


ibhagwan:
@Bekaboo - can you provide a repro/steps where this happens? table.concat won’t prefix a | unless it has a empty string in [1], wouldn’t it be a better fix to remove empty indexes?

phanen:
We may need to maintain 0.9 compatibility
```diff
@@ -1068,9 +1068,9 @@ function FzfWin:create()
       local parts = vim.split(winrestcmd, "|")
       local cmd = vim.tbl_map(function(cmd_part)
         local winnr = tonumber(cmd_part:match("(.)resize"))
-        return utils.tbl_contains(winnrs, winnr) and cmd_part or ""
+        return winnr and utils.tbl_contains(winnrs, winnr) and cmd_part or nil
       end, parts)
-      vim.cmd(table.concat(cmd, "|"))
+      vim.cmd(table.concat(vim.tbl_values(cmd), "|"))
       -- Also restore cmdheight, will be wrong if vim resized (#1462)
       vim.o.cmdheight = cmdheight
     end
```

phanen:
rebase can fix the ci (https://github.com/ibhagwan/fzf-lua/pull/2740).

Bekaboo:
> We may need to maintain 0.9 compatibility
> 
> ```diff
> @@ -1068,9 +1068,9 @@ function FzfWin:create()
>        local parts = vim.split(winrestcmd, "|")
>        local cmd = vim.tbl_map(function(cmd_part)
>          local winnr = tonumber(cmd_part:match("(.)resize"))
> -        return utils.tbl_contains(winnrs, winnr) and cmd_part or ""
> +        return winnr and utils.tbl_contains(winnrs, winnr) and cmd_part or nil
>        end, parts)
> -      vim.cmd(table.concat(cmd, "|"))
> +      vim.cmd(table.concat(vim.tbl_values(cmd), "|"))
>        -- Also restore cmdheight, will be wrong if vim resized (#1462)
>        vim.o.cmdheight = cmdheight
>      end
> ```

Fixed in 6a6e8fb

Bekaboo:
> @Bekaboo - can you provide a repro/steps where this happens? table.concat won’t prefix a | unless it has a empty string in [1], wouldn’t it be a better fix to remove empty indexes?

Seems that closing a buf when fzf-lua creates a new window will cause the error, but I haven't found a reliable minimal way to reproduce it except using my whole config. However, I believe the fix makes sense considering we already have this line: https://github.com/ibhagwan/fzf-lua/blob/fea9eedc6894c44d44cbb772a5cd11c93b82d7a1/lua/fzf-lua/win.lua#L1071
which checks if `winnr` is contained in `winnrs` and maps `cmd_part` to an empty string if not -- if `utils.tbl_contains(winnrs, winnr)` always returns true, then we shouldn't have this check in the first place, else we should consider the possibility of continuous empty strings in `cmd` which result to continuous "|" in the concatenated command and causes the error.

ibhagwan:
@coderabbitai review

ibhagwan:
Ty @Bekaboo

PR #2750, opened 2026-06-17
feat(git follow): Add option to make git_bcommits follow renames

### LINKED ISSUE #1544: Feature: Make git_bcommits follow renamed files
### Have you RTFM'd?

- [X] I have done proper research

### Feature Request

The `git_bcommits` command does not list the full history if a file was renamed in some previous commit. Other tools like [vim-fugitive](https://github.com/tpope/vim-fugitive)'s `:0GClog` can handle this.

I tried to add a `--follow` option here and it now showed also the commits before the file was renamed. But the preview window does not work for these older commits.
https://github.com/ibhagwan/fzf-lua/blob/3bd795ab8536435a1d61b8f394d83c88d1494403/lua/fzf-lua/defaults.lua#L396-L397

mikehaertl:
Core idea:

* Use `git log --follow --name-only` to print the original filename in another line after each log message
* Combine lines with log messages + lines with filenames in a loop into a single entry
* Use `--with-nth` feature of fzf to hide the filename

Fixes #1544 

## Summary by CodeRabbit

* **New Features**
  * Added a `follow` option to the `bcommits` picker to follow a buffer’s history across renames (non-visual mode only).
  * When enabled, entries use commit-aware file resolution and previews.
* **Documentation**
  * Documented the new `follow` option for `bcommits`, explicitly noting its default (`false`).
* **Behavior Changes**
  * Commit-specific file paths are resolved to reflect the filename as it existed at the selected commit, improving renamed-file accuracy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -936,6 +936,8 @@ previewers = {
     },
     bcommits = {
       prompt        = 'BCommits❯ ',
+      -- follow the buffer's history across renames
+      follow        = false,
       -- default preview shows a git diff vs the previous commit
       -- if you prefer to see the entire commit you can use:
       --   git show --color {1} --rotate-to={file}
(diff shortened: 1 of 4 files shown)


ibhagwan:
Tysm @mikehaertl for this wonderful feature!

PR #2752, opened 2026-06-17
fix: add native color support to git diff preview

zachyarbrough:
Added `--color=always` flag to git diff preview and cmd in order to support native git diff color in `git_commit` when navigating to individual commits making the diff easier to read without needing `delta`.

Fixes issue from discussion https://github.com/ibhagwan/fzf-lua/discussions/2751

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved git diff visualization with color output for better readability.


ibhagwan:
Ty @zachyarbrough !

PR #2754, opened 2026-06-19
fix(buffers): assertion failed when buflist is nil

### LINKED ISSUE #2736: fzf-lua/providers/buffers.lua:219: assertion failed!
I cannot stably reproduce, but I occasionally get this:
```
vim.schedule callback: /home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:440: /home/phan/b/fzf-lua/lua/fzf-lua/providers/buffers.lua:219: assertion failed!
stack traceback:
	[C]: in function 'assert'
	/home/phan/b/fzf-lua/lua/fzf-lua/providers/buffers.lua:219: in function 'contents'
	/home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:435: in function </home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:430>
	[C]: in function 'xpcall'
	/home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:430: in function 'fn'
	/home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:151: in function </home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:150>
stack traceback:
	[C]: in function 'error'
	/home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:440: in function 'fn'
	/home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:151: in function </home/phan/b/fzf-lua/lua/fzf-lua/shell.lua:150>
```

phanen:
Problem: `contents` callback calls `utils.CTX()` without `includeBuflist`,
so if the context was recreated between `core.lua:346` (which passes
`includeBuflist=true`) and the async `contents` invocation, `buflist` is
nil and the assertion crashes. The same file already handles this at
`line 282` with `utils.CTX().buflist or {}`.

Solution: pass `{ includeBuflist = true }` explicitly, ensuring the ctx
always has `buflist` populated when accessed inside the callback.

Closes #2736

## Release Notes

**No user-visible changes** in this update. Internal implementation adjustments have been made to improve code reliability.

### DIFF
--- a/lua/fzf-lua/providers/buffers.lua
+++ b/lua/fzf-lua/providers/buffers.lua
@@ -216,7 +216,7 @@ M.buffers = function(opts)
   if not opts then return end
 
   local contents = function(cb)
-    local buflist = assert(utils.CTX().buflist)
+    local buflist = utils.CTX({ includeBuflist = true }).buflist
     local filtered, _, max_bufnr = filter_buffers(opts, buflist)
 
     if next(filtered) then


ibhagwan:
So you found it :)

phanen:
no, it's still a llm slop, I don't understand the real reason. Give llm more time to persuude me...

ibhagwan:
GLM 5.2 is also convinced this is the right fix though I couldn’t reproduce this with its repro steps even when adding uv.sleep to make the refresh slower:
image

phanen:
Fun fact I never reproduce since then..

PR #2760, opened 2026-06-25
refactor(previewer): unify git_status diff previewer commands

### LINKED ISSUE #2747: What `git_diff.cmd_deleted` preview command is needed for? & Maybe add `--` suffix in `git_diff.cmd_modified` preview command?
Another small issue I encountered with `msys2/git` combined with `ucrt64/mingw-w64-ucrt-x86_64-{neovim,fzf}` packages, is that the default `git_diff.cmd_{deleted,modified}` commands cause this error to appear in the preview window of `:FzfLua git_status`:

```
error: Could not access 'HEAD'
```

I managed to get rid of it with:

```diff
diff --git i/lua/fzf-lua/defaults.lua w/lua/fzf-lua/defaults.lua
index f0321dde..e700b1e0 100644
--- i/lua/fzf-lua/defaults.lua
+++ w/lua/fzf-lua/defaults.lua
@@ -310,8 +310,8 @@ M.defaults = {
     ---@class fzf-lua.config.GitDiffPreviewer: fzf-lua.config.Previewer
     git_diff = {
       pager         = M._preview_pager_fn,
-      cmd_deleted   = "git diff --color HEAD --",
-      cmd_modified  = "git diff --color HEAD",
+      cmd_deleted   = "git diff --color --",
+      cmd_modified  = "git diff --color",
       cmd_untracked = "git diff --color --no-index /dev/null",
       -- TODO: modify previewer code to accept table cmd
       -- cmd_deleted   = { "git", "diff", "--color", "HEAD", "--" },
```

Which made me wonder why was `HEAD` so important. I am also unsure as to why I can't reproduce it outside of `fzf-lua` Neovim context.

Interestingly, the issue doesn't exist with `:FzfLua git_diff` which works with/out `HEAD` present there.

ibhagwan:
`git diff HEAD -- <file>` works correctly for both modified and deleted files (staged or unstaged), so we don't need separate commands.

Changes:
- defaults: set cmd_deleted to nil and add -- to cmd_modified for safety against filenames that look like options.
- previewer: when cmd_deleted is nil, fall back to cmd_modified. Users who have customized cmd_deleted in their config are unaffected.

This simplifies the defaults while preserving backward compatibility for any user overrides.

Fix #2747

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Git diff previews so deleted files now display correctly using the standard diff behavior when no custom deleted-file command is set.
  * Added a safer default for diff preview commands, helping previews work more consistently across modified and deleted files.

### DIFF
--- a/lua/fzf-lua/defaults.lua
+++ b/lua/fzf-lua/defaults.lua
@@ -310,12 +310,15 @@ M.defaults = {
     ---@class fzf-lua.config.GitDiffPreviewer: fzf-lua.config.Previewer
     git_diff = {
       pager         = M._preview_pager_fn,
-      cmd_deleted   = "git diff --color HEAD --",
-      cmd_modified  = "git diff --color HEAD",
+      -- NOTE: deleted files are handled by cmd_modified as well since
+      -- `git diff HEAD -- <file>` works for both modified and deleted.
+      -- Users can still override cmd_deleted if they wish.
+      cmd_deleted   = nil,
+      cmd_modified  = "git diff --color HEAD --",
       cmd_untracked = "git diff --color --no-index /dev/null",
       -- TODO: modify previewer code to accept table cmd
       -- cmd_deleted   = { "git", "diff", "--color", "HEAD", "--" },
-      -- cmd_modified  = { "git", "diff", "--color", "HEAD" },
+      -- cmd_modified  = { "git", "diff", "--color", "HEAD", "--" },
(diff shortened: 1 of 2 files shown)


ibhagwan:
@doronbehar, it seems that this will cause issues with new repositories according to @coderabbitai, to solve it we have to remove HEAD but then we’ll have issues with staged changes.

At this point I feel it’s best to leave this as is, the customization via setup should solve your issue rather easily.

PR #2762, opened 2026-06-28
fix(rpc): schedule os.exit

phanen:
https://github.com/neovim/neovim/pull/39937

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of RPC session termination to prevent premature or delayed shutdown behavior.
  * Updated long-running RPC “keep alive” behavior to better maintain responsiveness during extended activity.


phanen:
lint will fail because possible bug in emmylua-check. 
sync-remote fail because codeberg.org is down now. @ibhagwan

PR #2765, opened 2026-06-29
feat(serverlist): cpu usage and every(2)

phanen:
fzf 0.73 add timer support
https://github.com/junegunn/fzf/blob/f2e451596c5d23de0069cdc51ea3fec73b7fe835/CHANGELOG.md#L23-L34

## Summary by CodeRabbit

* **New Features**
  * Added per-server CPU usage to the server list, shown in a new CPU column alongside existing memory and socket details.
  * Enabled periodic server list refreshing (on supported fzf versions) so CPU readings stay up to date.
* **Bug Fixes**
  * Improved RPC request reliability by introducing an explicit timeout/cleanup path, returning a clear “Timeout” result instead of potentially waiting indefinitely.

### DIFF
--- a/lua/fzf-lua/providers/nvim.lua
+++ b/lua/fzf-lua/providers/nvim.lua
@@ -803,42 +803,75 @@ M.serverlist = function(opts)
     return string.format("%.2fG", kb / (1024 * 1024))
   end
 
+  local function cpu_time_us(rusage)
+    if not rusage or not rusage.utime or not rusage.stime then return nil end
+    return (rusage.utime.sec * 1e6 + rusage.utime.usec)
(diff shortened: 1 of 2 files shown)


phanen:
make sense... but this feature may never be useful since when a nvim instance take too much cpu, it usually won't response rpc request...

phanen:
wait.. rpc request TTL indeed take time so the review is even more correct to calculate hrtime  in peer.

phanen:
maybe related https://github.com/neovim/neovim/issues/8658.

ibhagwan:
Just used this for the first time, really cool, @phanen :)

PR #2780, opened 2026-07-18
build(make): parallel test runner

phanen:
`xargs` may be not needed since `make` can handle parallel itself


ibhagwan:
This is very cool @phanen, albeit a lot of work “just” to speed up the CI :-)

You can probably make a library from this or contribute to mini.nvim, would love to hear @echasnovski 2-cents about this as mini.nvim has a lot of tests and would greatly benefit from this.

phanen:
This hasn't not been really serious work, I push the agents to do some PoC. I think I need do more refactor to make more sense to build something more convinient to agent friendly harness, maybe on top of mini.test.

echasnovski:
> You can probably make a library from this or contribute to mini.nvim, would love to hear @echasnovski 2-cents about this as mini.nvim has a lot of tests and would greatly benefit from this.

A basic test parallelization should not be too much effort from the project's side if done per file. The main problem (for me) here is that reporter of parallel execution is so much more complex than for a regular sequential one. So I don't think it will be part of 'mini.test' itself, but I'd probably play around with basic parallelization on CI.

Issue #2808, opened 2026-09-14
Bug: keymaps picker doesn't open location


VladYagl:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [ ] I have read the Wiki including the Advanced section
- [ ] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Arch Linux

### Shell

fish

### Neovim version (`nvim --version`)

NVIM v0.12.5

### Fzf version (`fzf --version`)

0.74.3 (15f64c49)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [x] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [ ] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
local fzf = require('fzf-lua')
fzf.setup { 'default' }
```

### Describe the bug / steps to reproduce

The three "open the keymap definition" actions in the keymaps picker don't seem to work: ctrl-s, ctrl-v, and ctrl-t. The picker just closes without anything happening.

The previewer shows the definition correctly, so the location does seem to be available. It's only the actions that don't use it.

Steps to reproduce:

1. :FzfLua keymaps
2. Select mapping, e.g., grn, which Neovim defines in
   runtime/lua/vim/_defaults.lua. The preview pane shows the right
   file and line.
3. Press ctrl-v (same with ctrl-s / ctrl-t).

Expected: _defaults.lua opens in a vsplit at the mapping's line.
Actual: nothing happens.

Issue #2674, opened 2026-04-16
Bug: Missing file error when looking at keymaps picker


rubin55:
### RTFM Checklist

- [x] I have searched exisiting issues / discussions
- [x] I have read the Wiki including the Advanced section
- [x] I have read `man fzf` / I am well versed in shell fzf

### Operating system

Linux

### Shell

Bash

### Neovim version (`nvim --version`)

NVIM v0.12.1

### Fzf version (`fzf --version`)

0.71.0 (62899fd7)

### Output of `:lua print(os.getenv('FZF_DEFAULT_OPTS'))`

nil

### Is the problem reproducible with `mini.sh`?

- [ ] My issue is reproducible with `mini.sh`
- [ ] My issue **IS NOT** reproducible with `mini.sh`
- [x] I have not tested with `mini.sh` (not relevant, requires LSP, Windows, etc)

### Fzf-lua configuration

```lua
local fzf = require('fzf-lua')
fzf.setup({
  'telescope',
  ui_select = true
})
```

### Describe the bug / steps to reproduce

When I start the `keymaps` picker, I see all kinds of keymaps. The ones I've defined myself give no error, but ones not defined by me, when I select them, they give a missing file error:
<img width="1200" height="792" alt="Image" src="https://github.com/user-attachments/assets/2b29526a-cb1c-4d77-90e6-c2b996e3a33f" />

rubin55:
(Note that I can `:e $VIMRUNTIME/lua/vim/_core/defaults.lua` - the file is not missing, it just seems to not use the `$VIMRUNTIME` prefix - at least, if the error message is accurate).

ibhagwan:
This is a bit tricky to fix, while I can most likely guess that files starting with vim belong to the runtime it’s kind of hacky and I’m not sure if there are other similar issues, the debug info we get for this is incomplete path:
```
{
  currentline = -1,
  func = <function 1>,
  isvararg = false,
  lastlinedefined = 0,
  linedefined = 0,
  namewhat = "",
  nparams = 0,
  nups = 0,
  short_src = "vim/_core/defaults.lua",
  source = "@vim/_core/defaults.lua",
  what = "Lua"
}
```

Maybe @phanen has a non hacky idea how to get full path from `debug.getinfo`?

phanen:
start nvim with `--luamod-dev` can fix this.
[...]

PR #2809, opened 2026-09-15
fix(actions): keymap_edit don't work

phanen:
Fix
* https://github.com/ibhagwan/fzf-lua/issues/2808
* https://github.com/ibhagwan/fzf-lua/issues/2674#issuecomment-4306434012

## Summary by CodeRabbit

* **Enhancements**
  * Callback-generated entries now include source and line-definition details alongside path and line information.
  * This additional context appears when viewing or selecting callback-generated results.

* **Bug Fixes**
  * Failed code-action resolutions no longer trigger an error.
  * Screenshot generation now handles unavailable UI information safely.
  * Improved handling of callback-generated entries with additional source and line metadata.


ibhagwan:
Damn, 73 warnings lol

I totally forgot yo update emmylua after the ping due to the hang issue, might be worthwhile to reset it to `EMMYLUA_VERSION=` so we always use the latest instead of a pin?

phanen:
> I totally forgot yo update emmylua after the ping due to the hang issue, might be worthwhile to reset it to `EMMYLUA_VERSION=` so we always use the latest instead of a pin?

I think we need the same version as https://github.com/neovim/neovim/blob/65ef6fdaeef86f8341293942c076b320acf03fd1/CMakeLists.txt#L354-L356

